PR #23, opened 2018-01-10
Some headerbar simplifications

danrabbit:


src/Widgets/HeaderBar.vala
```diff
@@ -47,38 +44,25 @@ public class Akira.Widgets.HeaderBar : Gtk.Box {
...
-        toolbar = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 0);
-        toolbar.margin_start = 15;
-        toolbar.margin_end = 15;
-        toolbar.margin_top = 5;
-        toolbar.margin_bottom = 15;
+        set_title (APP_NAME);
+        set_show_close_button (true);
 
-        Gtk.Menu menu_items = new Gtk.Menu ();
+        var menu_items = new Gtk.Menu ();
```
danrabbit: This is just a code style thing. It's up to you if you want to use elementary code style or not, of course

src/Widgets/HeaderBar.vala
```diff
@@ -47,38 +44,25 @@ public class Akira.Widgets.HeaderBar : Gtk.Box {
...
         menu_items.add (new Gtk.MenuItem.with_label(_("Open")));
         menu_items.add (new Gtk.MenuItem.with_label(_("Save")));
         menu_items.add (new Gtk.MenuItem.with_label(_("Save As")));
         menu_items.add (new Gtk.SeparatorMenuItem ());
         menu_items.add (new Gtk.MenuItem.with_label(_("Quit")));
         menu_items.show_all ();
 
-        menu = new Akira.Partials.MenuButton ("folder", _("Menu"), _("Open Menu"));
+        menu = new Akira.Partials.MenuButton ("document-open", _("Menu"), _("Open Menu"));
```
danrabbit: You want to be careful to use icon names from actions for tool items. The semantic name of the icon is important and it will keep your app from breaking if this icon ever changes

src/Widgets/HeaderBar.vala
```diff
@@ -92,20 +76,19 @@ public class Akira.Widgets.HeaderBar : Gtk.Box {
...
         ruler = new Akira.Partials.HeaderBarButton ("applications-accessories", _("Ruler"), _("Toggle Ruler (Ctrl+⇧+R)"));
 
-        toolbar.pack_end (settings, false, false, 0);
-        toolbar.pack_end (new Akira.Partials.Spacer (35), false, false, 0);
-        toolbar.pack_end (layout, false, false, 0);
-        toolbar.pack_end (ruler, false, false, 0);
-
-        pack_start (toolbar, true, true, 0);
+        add (menu);
+        add (new Gtk.Separator (Gtk.Orientation.VERTICAL));
```
danrabbit: This spacer box thing is really dirty, there's already Gtk.Separator for this :)

src/Widgets/HeaderBar.vala
```diff
@@ -19,12 +19,9 @@
 * Authored by: Alessandro "Alecaddd" Castellani <castellani.ale@gmail.com>
 */
 
-public class Akira.Widgets.HeaderBar : Gtk.Box {
+public class Akira.Widgets.HeaderBar : Gtk.HeaderBar {
     private const string TOOLS_DIR = "/com/github/alecaddd/akira/tools/";
 
-    private Gtk.HeaderBar headerbar;
```
danrabbit: These items were never actually used outside of the construct method, so there's no need to make them private to the class

src/Widgets/HeaderBar.vala
```diff
@@ -19,12 +19,9 @@
 * Authored by: Alessandro "Alecaddd" Castellani <castellani.ale@gmail.com>
 */
 
-public class Akira.Widgets.HeaderBar : Gtk.Box {
```
danrabbit: In general, the use of Gtk.Box is discouraged. It's an old widget and it adheres to old Gtk2 ways of doing things like specifying padding when packing widgets

src/Widgets/HeaderBar.vala
```diff
@@ -47,38 +44,25 @@ public class Akira.Widgets.HeaderBar : Gtk.Box {
     public signal void new_window ();
 
     public HeaderBar () {
-        Object (orientation: Gtk.Orientation.VERTICAL, toggled: true);
```
danrabbit: I would actually recommend not adding properties here unless they are specific to this constructor. If you ever decide to subclass this widget, these properties won't be inherited. So it's really better to put them in the `construct` block. I learned that one the hard way :grimacing:

src/Widgets/HeaderBar.vala
Alecaddd: Yes, definitely I want to follow elementary code style.
I actually can't find it in the documentation if this is a good or bad practice, or it doesn't matter at all.
Thanks

src/Widgets/HeaderBar.vala
Alecaddd: Make sense, I don't really like the gear as a "Menu Open" image, but it's better to maintain consistency with the naming. I can always create a custom icon if I want to in the future.

src/Widgets/HeaderBar.vala
Alecaddd: Great! I was trying to find a native solution but couldn't. I agree that the spacer is a dirty piece of garbage :D

src/Widgets/HeaderBar.vala
Alecaddd: Oh really?
So, if an item is not used in any method of the class but only in the construct, it doesn't need to be declared before?
Where can I find this explanation in the documentation?

src/Widgets/HeaderBar.vala
Alecaddd: Sounds good.
In case I need a similar solution, is there something I could use instead of the `Gtk.Box` ?

src/Widgets/HeaderBar.vala
Alecaddd: This part is not really clear to me. What's the difference between declaring these attributes inside the Object() and the construct()?

src/Widgets/HeaderBar.vala
Alecaddd: Oh I see, the document-open is the same as the folder, but better for the semantic name. Thank :D

src/Widgets/HeaderBar.vala
danrabbit: The recommended replacement is Gtk.Grid. You can use it like a box, if you want. It has an orientation properly and you can use the `add` method with it instead of `attach`

src/Widgets/HeaderBar.vala
danrabbit: Whatever is done in `construct` is always done any time an instance of this class (even a subclass) is created.

You can have different kinds of constructors like “MyClass ()”, “MyClass.with_some_property ()” and then in your `Object ()` you specify things that are specific to that constructor. 

So for example we have like `MyClass ()`. Let’s say we specified `Object (margin: 6)` in it. In the future if we subclass `MyClass ()` or we create a constructor like `MyClass.with_some_property ()` then that 6px margin we created won’t be applied. It’s specific to the `MyClass ()` constructor. But if we instead add `margin = 6` to our `construct` block, then it will always be applied.

So basically, unless that property is actually specific to that constructor, you probably want to put it in your `construct` block instead

src/Widgets/HeaderBar.vala
danrabbit: Naw you only need to make it private to the class if you need to access it from another method in that class. Otherwise you can just use the `var` keyword. Not sure if that’s specifically documented somewhere or not

src/Widgets/HeaderBar.vala
danrabbit: As far as I know, there is no functional difference. It’s just “implicitly typed”. So basically the Vala compiler infers the type from what’s on the other side of the `=` symbol. It just saves some characters and personally I find the highlighting makes it easier to find when you declare variables :)

Issue #29, opened 2018-01-31
unable to build on arch linux


faebser:
I checked out the master with id ```4dc3532c1e98d7f3aef6623fc64ec40ba1ef1f93``` and tried to build it on Arch linux 4.14.15-1-ARCH.

mason output:

```
The Meson build system
Version: 0.44.0
Source dir: /home/faebser/workspace/Akira
Build dir: /home/faebser/workspace/Akira/build
Build type: native build
Project name: com.github.alecaddd.akira
Native C compiler: cc (gcc 7.2.1)
Native Vala compiler: valac (valac 0.38.6)
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (0.29.2)
Native dependency glib-2.0 found: YES 2.54.3
Configuring config.vala using configuration
Native dependency gtk+-3.0 found: YES 3.22.26
Native dependency granite found: YES 0.5
Native dependency gee-0.8 found: YES 0.20.0
Native dependency libxml-2.0 found: YES 2.9.7
Native dependency gtksourceview-3.0 found: YES 3.24.6

Meson encountered an error in file meson.build, line 20, column 0:
File src/config.vala does not exist.

```

output from meson-log.txt
```
Build started at 2018-01-31T10:35:01.161701
Main binary: /usr/bin/python
Python system: Linux
The Meson build system
Version: 0.44.0
Source dir: /home/faebser/workspace/Akira
Build dir: /home/faebser/workspace/Akira/build
Build type: native build
Project name: com.github.alecaddd.akira
Sanity testing C compiler: cc
Is cross compiler: False.
Sanity check compiler command line: cc /home/faebser/workspace/Akira/build/meson-private/sanitycheckc.c -o /home/faebser/workspace/Akira/build/meson-private/sanitycheckc.exe
Sanity check compile stdout:

-----
Sanity check compile stderr:

-----
Running test binary command: /home/faebser/workspace/Akira/build/meson-private/sanitycheckc.exe
Native C compiler: cc (gcc 7.2.1)
Running compile:
Working directory:  /tmp/tmpwfhgrqzy
Command line:  valac /tmp/tmpwfhgrqzy/testfile.vala -C -C -o /tmp/tmpwfhgrqzy/output.obj 

Code:
 class MesonSanityCheck : Object { }
Compiler stdout:
 Compilation succeeded - 1 warning(s)

Compiler stderr:
[...]

PR #35, opened 2018-03-16
Reconfigure meson to fix config.vala.in

davidmhewitt:
I was getting the following when trying to build:

`meson.build:20:0: ERROR: File src/config.vala does not exist.`

I've tweaked your meson files slightly to properly define a `config_header` variable from the configured version of `config.vala.in` and include that in the build.

I also took the liberty to move your compilation stage into `src/meson.build` for neatness.


davidmhewitt:
Appears to fix #29

PR #37, opened 2018-03-19
Avoid '__hidden' name collision

myfreeweb:
`/usr/include/sys/cdefs.h` on FreeBSD defines `__hidden` as `__attribute__((__visibility__("hidden")))`.
Vala expands `_hidden` into `__hidden`, so the macro gets expanded. Now `__hidden` will expand to `___hidden` and won't collide.

### DIFF
--- a/src/Layouts/Partials/Layer.vala
+++ b/src/Layouts/Partials/Layer.vala
@@ -54,9 +54,9 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
 		get { return _locked; } set { _locked = value; }
 	}
 
-	private bool _hidden { get; set; default = false; }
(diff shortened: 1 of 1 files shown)


Alecaddd:
Mh, interesting.
I'm not 100% convinced about this change. I know it's a super small update, but it breaks the consistency of naming style of my private attributes.

What if I use `_hide` and `_lock` instead of `_hidden` and `_locked`?

It just feels weird to have only one variable with a double underscore, and I'm sure I will forget after months of coding why only that variable has a double underscore, and I'll change it back, or another developer could do that because it's not consistent with the other attributes and looks like a typo.

myfreeweb:
Yeah just rename to whatever.

PR #38, opened 2018-03-23
post_install: honor the prefix

### LINKED ISSUE #36: the post install script does not honor the prefix
The [post install script](https://github.com/Alecaddd/Akira/blob/master/meson/post_install.py) contains hard-coded paths that assume the prefix is `/usr`. This breaks the build if the prefix is not `/usr`.

eyelash:
I'm new to Python so I hope I did everything right.

Fixes #36

Cheers

### DIFF
--- a/meson/post_install.py
+++ b/meson/post_install.py
@@ -3,14 +3,14 @@
 import os
 import subprocess
 
-schemadir = os.path.join(os.environ['MESON_INSTALL_PREFIX'], 'share', 'glib-2.0', 'schemas')
+install_prefix = os.environ['MESON_INSTALL_PREFIX']
+schemadir = os.path.join(install_prefix, 'share', 'glib-2.0', 'schemas')
 
 if not os.environ.get('DESTDIR'):
 	print('Compiling gsettings schemas...')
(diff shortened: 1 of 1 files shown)


eyelash:
If we could depend on Meson 0.46.0 we could use the `rename` argument of [`install_data`](http://mesonbuild.com/Reference-manual.html#install_data) to install the icons to the right place instead of renaming them in the post install script. I guess that would be the proper fix.

Alecaddd:
Thanks for the PR!
I can potentially depend on `meson > 0.46`, what do you think?
I'm new to all of this so I'm not sure what's the best approach. Would adopting a modern version of meson prevent users from install Akira successfully?
Cheers

ice0:
Meson can be easily installed using pip, so it always can be a latest version. As for me that is not a problem. Some time ago i tried to build Akira on Linux Mint 18, and can't because of some dependencies (required Vala version was too high), I build it from the sources, next was graphite and i didn't remember why i stop trying.
Yesterday i installed Elementary for testing and easily compiled Akira. So i think latest Meson is not a biggest problem here. Btw, awesome work, thank you!

PR #46, opened 2018-08-31
Included forum link on readme

AndreaMinato:
As I have seen reading the issues there are times where it is better to post a question in the forum.

So I added it on the readme to try to prevent some repeated issues


Alecaddd:
Thanks

PR #61, opened 2019-01-18
Reduce headerbar and icon complexity

cassidyjames:
This is a pretty opinionated PR, but I'm opening it for discussion and to show how I'd go about it.

image

- **Removes different iconography styles:** this is a design decision that was being passed on to the user, asking them to design the app for you. :wink: It would be less complex to devote the time to shipping one set of very good icons instead of three entirely separate sets.

- **Removes the button labels from the headerbar:** This has never been an elementary design, and even in more complex apps like Code or Spice-Up, they are discoverable without always-visible labels. And similar to the previous point, adding a setting to determine it just adds complexity and passes the design decision off to the user.

- **Adds Granite accel tooltips:** Instead of labels and then near-duplicate tooltips, this branch uses the newer Granite accel tooltips to list the button's name and keyboard shortcut (if any) in a nicely-formatted tooltip. This is what we've been adopting system-wide, i.e. in Code, Files, Mail, etc.

- **Removes unused assets:** including styles and icons that are not being used.

- **Greatly reduces complexity in the headerbar:** Mostly as a consequence of the former points, there's a lot of switching, packing, and methods that are removed because they are no longer needed in this branch.

As a side-effect of this, Akira will be more recognizable as well if it doesn't have several overlapping UI preferences that change the look and feel of it so much. I can understand a dark/light toggle for a content creation app, but a lot of the UI preferences seemed to just add complexity.


Alecaddd:
Thanks for the PR, I'll properly review it later tonight when I'm off work, but I can totally see how much technical debt it removes.
I'll sit on it for a bit and collect my thoughts.
Thank you so much!

benwaffle:
As someone who hasn't seen this UI before, the icons are confusing, especially the 3 purple ones in the top right. I think labels would be useful.

julientaq:
I'm pretty interrested by the MR because it bring a couple of important things. 

- I agree with @cassidyjames that the set of icon should be defined within the app (so user can share screenshot and screencast with the same icons without the need to explain it).

But the idea that you ask the user to do the design for you is not true: a user will design for him (and not for everyone). Therefore, offering modularity in the layout of the app is a pretty neat option: user can make their tool the way they want. Any app should be keeping that in mind and offer modularity where it's possible (depending on the development cost though). 


- Pleaser don't hide the label of an icon. It makes the app harder to work with. Unless you're a power user and then you want to win some space, but hey, we're working on 13-27 inches screens and more, so i dont see where you would absolutely need to do that. (unless @cassidyjames has an idea of when it would be needed). Plus, in terms of accessibility, it's a bad move: for example, with the icons alone, i wouldn't be able to tell which one is what without having the mouse hovering them. Brain work put in understanding the tool instead of working with it. Last thing about icons: they are looked through a cultural biasis so what works for me doesnt work for my neighbour, especially with complex concept: are we talking about the layout of the app or the layout i'm designing right now?

 
-  @cassidyjames is right about the tooltip: they shouldn’t share the same information. Adding the shortcut is already a nice move. It would be interresting to find something else to add in there: Maybe give a bit of context about what the tool do.

Alecaddd:
Thank you so much for this great comments and really insightful conversation.

*Regarding the Icons*
Those are totally work in progress and I'm the first to admin that are not intuitive and clearly understandable. Lots of proper design work needs to be done.

*elementary HIG*
While I agree with @julientaq that this shouldn't be considered an elementary OS only app (and it won't), I consider the elementary HIG a high standard of usability and accessibility which I want to follow as much as possible.
With that said, I've been using proprietary design software for many years and I'm learning also from their decisions. Something I always remind myself is "if Photoshop was perfect no one would have switched to Sketch", and one of the main strengths of Sketch is the labeled header bar tools for first-time users.

Like @cassidyjames said:
> For the few times they need to hover them (likely the first time they use the app), then the tooltip will teach the user and make it so the next time they see that same icon they know what it does.
This is something I want to completely avoid and remove.
Even those few seconds wasted the initial 3 or 4 times you use the app are vital to remove entry barriers and friction points in the process of learning a new application.

*This is a WIP*
No matter what decisions will be used and embraced, this is still a work in progress and we will learn things as we go. The main and pretty much only goal I have for a Akira is to allow designers to open the app the first time and start working. Striving for an almost flat learning curve is what will differentiate Akira.

Imerion:
I have to agree with the original intention of the app here and Alecadds ideas here. A quick learning curve and the ability for power users to adjust settings is the way to go, and Icon Labels and clear, obvious icons is the best way to achieve that.

The importance of text labels or similar should not be underestimated. I have been teaching computer classes for all sorts of people for many years now, and if it's one thing that is very clear to me it's that user interfaces with labels and clear text elements (combined with colorful icons for quick orientation) are so much easier to get into and work with overall. Put a user in front of an app that has labels beneath icons or - even better: drop down menus - and they will immediately start to look for and find the actions they need. Put them in front of an app without that, especially one that relies overtly much on simple icons (in the worst-case scenario simple line-art scribbles or symbolic icons) and I will very soon have to start explaining and showing them where things are - again and again. This is even more important for people with impaired sight, trouble remembering, etc. Many of the people I teach are stroke patients and without labels they often have huge trouble remembering where various parts of the program are. Clear labels or drop-down menus is often the only way for them to use software on a continuous basis. This is especially obvious when comparing with modern mobile phone interfaces - these people have much much harder to use a modern smartphone than a classic piece of PC software. It's about accessability, and a reductionist approach will inevitably make software accessible for fewer people. Being worried about overwhelming user with options can easily be fixed by putting them under an "advanced" tab or similar.

Also, even me, an advanced user, experience a huge increase in working speed when labels and similar are present, even after using an app for a long time. More things for the brain to catch and the eye to zoom in on simply helps efficiency.

I could go on and on about other benefits of this, but will conclude that as long as there is an option it should make everyone happy. Options are important and the current trend of trying to create user interfaces that adhere to "one size fits none" is highly problematic - more options mean more more people will be able to use your software in a useful way. Sure, that has to be balanced against the time it takes to implement, but for this specific case I think the effort is definitely worth it.

Also, I can't be the only one who thinks the labels look better aesthetically. It makes the app look more well-rounded, advanced and complete. The screen above looks so simple and empty I might have skipped right passed it if I saw it in a software list simply because I would think it probably was to simple for my needs.

julientaq:
On that one, i just found this, which is pretty interresting: https://mattwilcox.net/musing/the-ineffectiveness-of-icons

PR #68, opened 2019-01-23
redesign mainwindow

Silur:
TODO:
* load the layout with gtk builder from file
* conditional behaviour

### DIFF
--- a/data/partials/toolbar.css
+++ b/data/partials/toolbar.css
@@ -1,0 +1,12 @@
+.toolbarBtn {
+	border-color: green;
+}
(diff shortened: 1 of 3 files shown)


Alecaddd:
Hi, thank you so much for your work and PR.
Unfortunately, we're not gonna use Glade to handle the UI of the application, and everything will be handled via code.
Also, it is a bit early to accept PRs as we're currently creating precise design references to have guidelines of how the UI will look and behave.
I hope to have future PRs from you once the project properly evolves.
Cheers

PR #109, opened 2019-03-04
New headerbar goocanvas revert

Philip-Scott:
Reverts the changes done on GooCanvas to prevent window breakage. That code is not ready for merging to master


Alecaddd:
@Philip-Scott wait, this is going back too much as it reverts back to goocanvas 1 and discards completely the goocanvas2 upgrade I did.
The issue is only related to the latest rotation Matrix that Alberto did, no need to roll back everything else.

Philip-Scott:
I thought I reverted to what was on master, has the Goocanvas2 update not been pushed though a PR?

Alecaddd:
Yeah, my bad for putting it in there. I wanted to have an updated goocanvas to check the zooming logic. I promise I'll revert back to when it was working without warnings and then I won't do this h**t anymore without proper PRs 😅

PR #110, opened 2019-03-04
initial work to open .akira files

### LINKED ISSUE #105: Open and save .akira files
## Expected Behavior
Akira should be able to open and save .akira files. This means serializing everything into it's respective files inside of a zipped archive.

## Possible Solution
I started making a ZippedArchiveHandler class that allows for a seamless way to unzip zipped files into a hidden directory. It is already working quite well for Spice-Up since i have to take care of multiple files inside of the main file. 

https://github.com/Philip-Scott/Spice-up/pull/264/files#diff-2a318b87418c3942367a6b12e51cbef2

The current implementation in Spice-Up uses the `libarchive` library to take care of zipping and unzipping files

Philip-Scott:
- Changes the Application class to handle opening files
    - only one per window and we do not allow the same file to be opened twice 
- Uses the new ArchiveHandler class to handle zipping and unzipping files
- Creates the AkiraFile class where the file and directory structure will live.

#### Proposed File Structure: 
- `content.json` The json with the akira file data! This will contain every we need to load everything, but not the actual artboards
- `version` File which holds the current version of the file. For future proofing the app
- `/Artboards` Folder which contains all the artboards. Each artboard is it's own SVG file. This allows individual files and diffs to stay smaller (And we could even revert a single artboard)
- `/Objects` Folder with a library of objects the user could create to re-use within the project (?)
- `/Pictures/` For links to images inside the akira file
- `/Thumbnails/` To store the artboard or object thumbnails
- `/.git/` The place for the git repo. Will be invisible from the AkiraFile class though and only used by the git backend

Pictures, Artboards, Thumbnails and others are stored with randomized file names to prevent collisions, and could be re-used in multiple places while only storing one file

## Steps to Test
To open a file, just pass it as a cli arg: 

`com.github.akiraux.akira ~/akira-file-test.akira`

## Screenshots 

image

- [x] Handle saving and updating of JsonObjectArray objects in JsonObject.vala
   - [x] Add/remove nodes as well
- [x] We need a way to mark Json Nodes as to be deleted, so that they wont be serialized when saving. But we also need to hold it's reference for undo/redo 
   - EDIT: The actual object it represents can hold the reference, and the JsonObject can now be removed from arrays, and null properties wont be serialized anymore 
- [x] Saving a property of a ObjectArray needs to update the base json object for serialization later on. 
- [ ] Fix indentation.... because _someone_ decided to use tabs instead of spaces __**cough cough** @Alecaddd__
- [x] Fix debian packaging (?) No idea what's going on here tbh, i did added the package name

- Fixes #105


Philip-Scott:
Had to make use of the JsonObjectArray class because vala does not allow you to `typeof (JsonObject[])` or even something like an array of ints. Only `typeof (string[])` seems to work because vala or glib treats it as it's own separate type. Part of me hopes that we can get rid of it, but it would also allow us to better control arrays as mentioned above. This class could have the logic for deleting nodes but keeping the reference

Philip-Scott:
I'm going to propose it for merging now, I'll fix the indentation once we decide if we want to use tabs or spaces, but the code logic should be ready for reviewing.

I made a standalone project to test the json part over at https://github.com/Philip-Scott/misc/tree/master/json-Vala-properties

src/Window.vala
```diff
@@ -88,6 +90,8 @@ public class Akira.Window : Gtk.ApplicationWindow {
 	public bool before_destroy () {
 		update_status ();
 
+		save_and_close_current_file ();
```
bilelmoussaoui: Can we store the uri of the latest opened file here? If you're planning to do that later, just ignore my comment here.

Philip-Scott: The window actually already stores a local reference to the file it has opened called AkiraFile. :)

`public FileFormat.AkiraFile? akira_file = null;`

Alecaddd:
This looks good to me.
Let's merge it!

PR #119, opened 2019-03-06
Add shapes from menu

albfan:
Allow adding basic shapes by menu and toggle visibility with right side bar

Add shapes to canvas from header bar insert menu
Toggle visibility for added shapes through right side bar

## Steps to Test
You can run through flatpak and just add shapes from menu and toggle visibility from right side bar

## Screenshots 

captura de pantalla de 2019-03-06 23-08-42

Model is needed to make things more maintanable
Z-Order can be implemented from D&D feature from right side bar
Colors and strokes are not retrieved from last used (a properties pane is needed for that)

There's no specific issue, just basic functionality for milestone 1.0


Alecaddd:
The layers panel style it's broken, it doesn't look like it was before the PR, how come?
Also, shouldn't we implement the expected behaviour of generating a shape by click-and-dragging motion?
Maybe this is out of the scope for this PR.

Alecaddd:
Also, an Artboard is another type of CanvasItem, which encapsulates all the shapes and objects like a sort of group.
So, if we're generating shapes on an empty canvas, those layers should be listed not inside an artboard.

PR #121, opened 2019-03-12
Use affine transforms

albfan:
Relay on affine transforms to translate and rotate,
reapplying it on hover, select effect and nob position

Using affine transformations allow us to use the internal state of objects to track its changes and apply to related graphic utilities (like nobs or effects)

## Steps to Test
Reproduce this screencast:

affine-transform

## Screenshots 

See screencast

Stopping translate out of visible canvas is missing for simplicity, it can be reworked now all is solved

Rotate objects, which needs refactoring translate and resize

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -65,12 +65,9 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     private Goo.CanvasRect? hover_effect;
 
     private bool holding;
-    private double event_x_root;
-    private double event_y_root;
-    private double start_x;
-    private double start_y;
-    private double start_w;
-    private double start_h;
+    private bool temp_event_converted;
+    private double temp_event_x;
+    private double temp_event_y;
     private double delta_x;
     private double delta_y;
     private double hover_x;
@@ -87,38 +84,39 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         events |= Gdk.EventMask.BUTTON_PRESS_MASK;
         events |= Gdk.EventMask.BUTTON_RELEASE_MASK;
         events |= Gdk.EventMask.POINTER_MOTION_MASK;
-        get_bounds(out bounds_x, out bounds_y, out bounds_w, out bounds_h);
+        get_bounds (out bounds_x, out bounds_y, out bounds_w, out bounds_h);
     }
 
     public override bool button_press_event (Gdk.EventButton event) {
         remove_hover_effect ();
 
         current_scale = get_scale ();
-        event_x_root = event.x;
-        event_y_root = event.y;
+        temp_event_x = event.x / current_scale;
+        temp_event_y = event.y / current_scale;
+        temp_event_converted = false;
 
-        var clicked_item = get_item_at (event.x / current_scale, event.y / current_scale, true);
+        debug ("canvas temp event x: %f\n", temp_event_x);
+        debug ("canvas temp event y: %f\n", temp_event_y);
+
+        var clicked_item = get_item_at (temp_event_x, temp_event_y, true);
 
         if (clicked_item != null) {
+
             var clicked_id = get_grabbed_id (clicked_item);
             holding = true;
 
             if (clicked_id == Nob.NONE) { // Non-nub was clicked
                 remove_select_effect ();
-                if (clicked_item is Goo.CanvasItemSimple) {
-                    clicked_item.get ("x", out start_x, "y", out start_y, "width", out start_w, "height", out start_h);
-                    print("start event: start_x %f, start_y %f, start_w %f, start_h %f\n", start_x, start_y, start_w, start_h);
-                }
-
                 add_select_effect (clicked_item);
                 grab_focus (clicked_item);
 
                 selected_item = clicked_item;
+
                 holding_id = Nob.NONE;
             } else { // nub was clicked
-                selected_item.get ("x", out start_x, "y", out start_y);
                 holding_id = clicked_id;
             }
+
         } else {
             remove_select_effect ();
             grab_focus (get_root_item ());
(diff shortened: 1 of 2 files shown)


src/Lib/Canvas.vala
```diff
@@ -94,31 +94,33 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
                 selected_item = clicked_item;
                 holding_id = Nob.NONE;
             } else { // nub was clicked
-                selected_item.get ("x", out start_x, "y", out start_y);
                 holding_id = clicked_id;
             }
+
+            event_x_root = canvas_event_x_root;
+            event_y_root = canvas_event_y_root;
+            convert_to_item_space(selected_item, ref event_x_root, ref event_y_root);
```
Philip-Scott: nit, but add a space before the `(`

src/Lib/Canvas.vala
```diff
@@ -153,70 +153,88 @@ public class Akira.Lib.Canvas : Goo.Canvas {
             motion_hover_event (event);
             return false;
         }
+        var canvas_event_x = event.x;
+        var canvas_event_y = event.y;
+        var event_x = canvas_event_x;
+        var event_y = canvas_event_y;
```
Philip-Scott: this block of clode was a bit weird. it may be easier to understand if you all setts come from the same variable like:

```
        var canvas_event_x = event.x;
        var canvas_event_y = event.y;
        var event_x = event.x;
        var event_y = event.y;
```

src/Lib/Canvas.vala
```diff
@@ -153,70 +153,88 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
+
+        convert_to_item_space(selected_item, ref event_x, ref event_y);
+        var temp_event_x = canvas_temp_event_x;
+        var temp_event_y = canvas_temp_event_y;
+        convert_to_item_space(selected_item, ref temp_event_x, ref temp_event_y);
```
Philip-Scott: wait a second, you might be having an extra variable here. canvas_temp_event_x was not changed before, so why make a new variable for these? and isn't `temp_event_x` also equal to `event_x`, so why two variables?

src/Lib/Canvas.vala
```diff
@@ -153,70 +153,88 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
-        var new_width = start_w;
-        var new_height = start_h;
+        var new_height = height;
+        var new_width = width;
 
         switch (holding_id) {
             case Nob.NONE: // Moving
-                new_x = fix_x_position ((delta_x + start_x), start_w);
-                new_y = fix_y_position ((delta_y + start_y), start_h);
+                selected_item.translate(delta_x, delta_y);
```
Philip-Scott: Add spaces before the `(` :)

src/Lib/Canvas.vala
```diff
@@ -339,24 +361,29 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
 
         var item = (target as Goo.CanvasItemSimple);
 
-        var line_width = 2.0 / get_scale ();
-        var stroke = item.line_width;
-        var real_x = x - (line_width * 2);
-        var real_y = y - (line_width * 2);
-        var width = item.bounds.x2 - item.bounds.x1 + stroke - line_width;
-        var height = item.bounds.y2 - item.bounds.y1 + stroke - line_width;
+        var line_width = get_scale() * 2;
```
Philip-Scott: the division might be correct as it was before. When zooming in the hover effect line is now thicker than before

src/Lib/Canvas.vala
albfan: it tries to reflect `event_y` comes from `canvas_event_x` to be in sync with rest of variable naming.

I tried to get rid of canvas_space or item_space variables, but something don't work as expected. Transforms affect object so item_space is transformed, sometimes is useful (translate) and others is not (rotate)

Let's fix the other obvious things and will come back to this later. I'd want to get rid of it too.

src/Lib/Canvas.vala
albfan: It is at the end of this function. canvas_temp_event_x is the last position processed. Happens that rotate transform is not absolute, but cumulative, so the angle from canvas_event_x_root will not work. canvas_temp_event_x (and its item_space pair temp_event_x) allows to get the angle rotated from the very previous notify_motion_event.

albfan:
Resizing elements is almost working. I need to control better what happens with width and height. Some nobs works correctly right now (RIGHT_BOTTOM, BOTTOM_CENTER,...)

About canvas and item position variables, I think it should be good enough right now, until totally understand goocanvas spaces and maybe find a simpler way to calculate transformations, just to go forward.

albfan:
Resize still needs to deal with current_scale, and respect MIN_SIZE on collapse, but at least the "move effect" is gone.

NOTE: there're lot of new vars to track deltas, fixed sizes and fixed deltas. All are supposed to be removed in final implementation

albfan:
This needs to deal better with ignored movements from mouse (noticeable when resize get close to borders) and current_scale, but mostly done

Alecaddd:
@albfan This is looking really good and it almost works perfectly!
I fixed a small thing related to the stroke size of the hover effect when zooming.

There are some small issues:
* A partially glitchy resize from all the left and top nobs
* When resizing from left to right, the action doesn't stop and moves the object
* When multiple identical shapes are on the canvas, the hover and select effect can't distinguish them

I think we could merge this and then open dedicated issues/PRs to tackle these things separately. What do you think?
I'm good with merging if you want.

albfan:
Ok will do a last improvement, and merge it.

albfan:
@Alecaddd Still not perfect but:

- A partially glitchy resize from all the left and top nobs
- When resizing from left to right, the action doesn't stop and moves the object

should be mostly solved. (still need polish, but we can do later)

About:

- When multiple identical shapes are on the canvas, the hover and select effect can't distinguish them

Not really sure what it is, so please can you tweak it and push this branch?

Alecaddd:
Amazing work!
Let's merge this and then start improving it with small and dedicated PRs.
You, sir, just created a pretty outstanding canvas!

albfan:
The best decision. Let's link objects and properties!

Issue #88, opened 2019-02-27
Left Panel: Create Alignment box


Alecaddd:
Create the alignment box in the Left Panel.
This box is always present and doesn't disappear based on the selected item.

Reference UI mockup:
![align-box](https://user-images.githubusercontent.com/2527103/53469454-ccb8db80-3a12-11e9-823d-96e3dc927029.png)

Full left panel as reference:
![left-panel](https://user-images.githubusercontent.com/2527103/53469461-d3475300-3a12-11e9-9f98-502eb9082103.png)

giacomoalbe:
Hi @Alecaddd, 

Just a question: where can one find the vector icons to implement this design? 

I read that you wanted to make custom icons for Akira, so I think that the elementaryOS icons are not going to fit in there. 

I saw that there are some icons in the data folder, but is seems that those icons (and other referenced in other designs as well) are missing. 

Can you provide them? Or it is possibile to get them from the vector mockup of the UI? 

Thanks!

Btw, I'm starting this task right now :muscle:

giacomoalbe:
Hi all,

It's with *huge* pride and *immense* joy that I announce you my first ever contribution to the Akira project, the skeleton of the Alignment Box :)

Here you can see a screenshot I take to the final prototype. 

![image](https://user-images.githubusercontent.com/11556031/53532252-791fbe00-3af6-11e9-932e-70c46fb85564.png)

As you might have noticed, the icons are different, but it's only a matter of uploading the correct ones and everything should be fine!

I'll be making the PR as soon as I add more features, maybe refactor the code a bit (it has been my first time with both Akira and Gtk/Vala, so I'm pretty happy with the results) and add some hooks for the logic behind button press.

Colors and sizes are taken from the image provided, I think they're consistent with the design. 

Overall I'm very satisfied, looking forward to contribute more (maybe with some other shapeObjectPanel task)!

Alecaddd:
@giacomoalbe the designers at elementary OS are helping us a lot in making and upgrading icons specifically to accommodate our needs. So, we decided to go with the native elementary icons.

If you're on elementary, you should be able to pull those without problems. For anyone else, we will soon release a `develop` flatpak installation with the elementary stylesheet and icons baked into it to allow consistency during development.

I'd recommend checking what I did in the `new-headerbar` branch regarding the change between symbolic and regular icons, in order to give users the choice.

Thank you so much for your help!

giacomoalbe:
Hi all, 

[...]

PR #124, opened 2019-03-18
Left Panel: New alignment box

giacomoalbe:

Initial structure for the Alignment Panel has been created. 
There is a `Gtk.Grid` centered in the LeftPanel with all the buttons needed for the alignment of elements on the canvas. 

In addition, I've created a bus (not D-Bus) to ease the communication between components without the need to honor the callback chain from the `window` object to the event triggerer, to help to decouple code and to ease multiple event subscription through the signal-slot mechanism. 

The object is called `EventBus` and is created as a singleton and made available in the same way as the `settings` object. 

Alignment Panel's Buttons already communicate this way and also their images are updated using this technique, instead of manually triggering the `update_icon` method from the Settings dialog page. 
 
## Steps to Test
 
There's no testing needed, maybe it can be proved that the prototype actually respects the mockup sizing. 

## Screenshots 

image

I'm not totally convinced about the size of the icons, I think that `Gtk.IconSize.LARGE_TOOLBAR` is not going to work, I'd rather use the `SMALL_TOOLBAR` in either case (symbolic and not symbolic), I think that the design is more balanced this way. 

- Implements #88

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -21,7 +21,6 @@
 
 namespace Akira {
 	public Akira.Services.Settings settings;
-	public Akira.Services.EventBus event_bus;
 }
 
 public class Akira.Application : Granite.Application {
@@ -36,7 +35,6 @@ public class Akira.Application : Granite.Application {
 		build_version_info = Constants.VERSION_INFO;
 
 		settings = new Akira.Services.Settings ();
-		event_bus = new Akira.Services.EventBus ();
 		windows = new GLib.List <Window> ();
 
 		program_name = "Akira";
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -21,6 +21,7 @@
 
 namespace Akira {
 	public Akira.Services.Settings settings;
+	public Akira.Services.EventBus event_bus;
 }
 
 public class Akira.Application : Granite.Application {
@@ -35,6 +36,7 @@ public class Akira.Application : Granite.Application {
 		build_version_info = Constants.VERSION_INFO;
 
 		settings = new Akira.Services.Settings ();
+		event_bus = new Akira.Services.EventBus ();
 		windows = new GLib.List <Window> ();
 
 		program_name = "Akira";
@@ -58,8 +60,8 @@ public class Akira.Application : Granite.Application {
 	}
 
 	protected override void activate () {
-		Gtk.Settings.get_default().set_property("gtk-icon-theme-name", "elementary");
-		Gtk.Settings.get_default().set_property("gtk-theme-name", "elementary");
+		Gtk.Settings.get_default ().set_property ("gtk-icon-theme-name", "elementary");
+		Gtk.Settings.get_default ().set_property ("gtk-theme-name", "elementary");
 
 		weak Gtk.IconTheme default_theme = Gtk.IconTheme.get_default ();
 		default_theme.add_resource_path ("/com/github/akiraux/akira");
--- a/src/Layouts/Partials/AlignItemsPanel.vala
+++ b/src/Layouts/Partials/AlignItemsPanel.vala
@@ -19,6 +19,8 @@
  * Authored by: Giacomo "giacomoalbe" Alberini <giacomoalbe@gmail.com>
  */
 public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
+	public weak Akira.Window window { get; construct; }
+
     private Gtk.Grid alignment_box;
 
     private struct AlignBoxItem {
@@ -28,7 +30,7 @@ public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
         public string tooltip_text;
     }
 
-    private const AlignBoxItem[] align_items_panel_buttons = {
+    private const AlignBoxItem[] ALIGN_ITEMS_PANEL_BUTTONS = {
         { "btn", "even-h", "distribute-horizontal-center", "Distribute centers evenly horizontally" },
         { "btn", "even-v", "distribute-vertical-center", "Distribute center evenly vertically" },
         { "sep" },
@@ -41,10 +43,11 @@ public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
         { "btn", "alig-v-b", "align-vertical-bottom", "Align bottom sides" }
     };
 
-    public AlignItemsPanel () {
+    public AlignItemsPanel (Akira.Window window) {
         Object (
+            window: window,
             orientation: Gtk.Orientation.VERTICAL
-            );
+        );
     }
 
     construct {
@@ -56,7 +59,7 @@ public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
 
         int current_button_column = 0;
 
-        foreach (var item in align_items_panel_buttons) {
+        foreach (var item in ALIGN_ITEMS_PANEL_BUTTONS) {
             switch (item.type) {
                 case "sep":
                     alignment_box.attach (new Gtk.Separator
@@ -66,7 +69,8 @@ public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
 
                 case "btn":
                     var tmp_align_box_button =
-                        new Akira.Partials.AlignBoxButton (item.action,
+                        new Akira.Partials.AlignBoxButton (window,
+                                                           item.action,
                                                            item.icon_name,
                                                            item.tooltip_text);
 
--- a/src/Layouts/Partials/AlignItemsPanel.vala
+++ b/src/Layouts/Partials/AlignItemsPanel.vala
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2019 Alecaddd (http://alecaddd.com)
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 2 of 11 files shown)


src/Services/EventBus.vala
```diff
@@ -0,0 +1,39 @@
...
+    public EventBus () {
+        Object ();
+    }
+
+    public void emit (string signal_id, string param = "") {
+        switch (signal_id) {
+            case "update_icons_style":
+                update_icons_style ();
+                break;
+            case "align_items":
```
Philip-Scott: I know that this won't seem consistent, but we should change these strings to use `-` instead of `_`, like  `update-icons-style`. This way, whenever we use the EventBus for property changes, these strings will match :) 

(GLib string representation of properties use `-`, why? No idea... but this will allow us to keep it consistent)

src/Application.vala
```diff
@@ -21,6 +21,7 @@
 
 namespace Akira {
 	public Akira.Services.Settings settings;
+	public Akira.Services.EventBus event_bus;
```
Philip-Scott: Sadly, we can't have this event bus as a global variable. This will cause issues when we have more than one instance of the app running at the time. Instead, this bus should be a part of the AkiraWindow, and we need to pass the bus object along the way :)

src/Layouts/Partials/AlignItemsPanel.vala
```diff
@@ -0,0 +1,84 @@
...
+        { "sep" },
+        { "btn", "alig-h-l", "align-horizontal-left", "Align left sides" },
+        { "btn", "alig-h-c", "align-horizontal-center", "Center on horizontal axis" },
+        { "btn", "alig-h-r", "align-horizontal-right", "Align right sides" },
+        { "sep" },
+        { "btn", "alig-v-t", "align-vertical-top", "Align top sides" },
+        { "btn", "alig-v-c", "align-vertical-center", "Center on vertical axis" },
+        { "btn", "alig-v-b", "align-vertical-bottom", "Align bottom sides" }
+    };
+
```
Philip-Scott: nit, but the public constructor should be just above the `construct` block :)

src/Partials/AlignBoxButton.vala
```diff
@@ -0,0 +1,92 @@
...
+    public string action;
+
+    public Gtk.IconSize icon_size;
+
+    private Gtk.Button button;
+    private Gtk.Image image;
+
+    public signal void triggered (Akira.Partials.AlignBoxButton emitter);
+
+    public AlignBoxButton (
```
Philip-Scott: This could be just in one line :) It's short enough i think

src/Partials/AlignBoxButton.vala
```diff
@@ -0,0 +1,92 @@
...
+        string action,
+        string icon_name,
+        string title) {
+        this.action = action;
```
Philip-Scott: We should also use gobject style constructors here :) 

- First, Make the properties we want to set by default like the icon_name, action and title as `public string icon_name { get; construct; }`
- Then, set the values using the `Object()` call, as you did in the AlignItemsPanel
- And finaly, put the rest of the code building the object inside of a `construct {` block :)

src/Partials/AlignBoxButton.vala
```diff
@@ -0,0 +1,92 @@
...
+public class Akira.Partials.AlignBoxButton : Gtk.Grid {
+    public string icon_name;
+    public string action;
+
+    public Gtk.IconSize icon_size;
+
+    private Gtk.Button button;
+    private Gtk.Image image;
+
+    public signal void triggered (Akira.Partials.AlignBoxButton emitter);
```
Philip-Scott: nit, but move the signal at the very top of the declared items :) 

the order should be:

- Consts
- Signals
- public fields
- private fields 
- constructors
- Object functions

src/Partials/AlignBoxButton.vala
```diff
@@ -0,0 +1,92 @@
...
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA
+ *
+ * Authored by: Giacomo "giacomoalbe" Alberini <giacomoalbe@gmail.com>
+ */
+public class Akira.Partials.AlignBoxButton : Gtk.Grid {
+    public string icon_name;
+    public string action;
+
+    public Gtk.IconSize icon_size;
```
Philip-Scott: Do we really need to store a variable for icon_size? :) Can't we just set it directly to the image?

src/Partials/AlignBoxButton.vala
```diff
@@ -0,0 +1,92 @@
...
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA
+ *
+ * Authored by: Giacomo "giacomoalbe" Alberini <giacomoalbe@gmail.com>
+ */
+public class Akira.Partials.AlignBoxButton : Gtk.Grid {
```
Philip-Scott: Also, why did you make it a Grid? :) 

If this is a button, we could just have the class be a Button instead, and we can always add the icon later in the constructor. This will drop the number of grids :)

src/Partials/AlignBoxButton.vala
```diff
@@ -0,0 +1,92 @@
...
+            : Gtk.IconSize.LARGE_TOOLBAR;
+
+        var new_icon_name = settings.use_symbolic == true
+            ? this.icon_name + "-symbolic"
+            : this.icon_name;
+
+        this.button.remove (this.image);
+        this.image = new Gtk.Image.from_icon_name (new_icon_name, new_icon_size);
+        this.button.add (image);
+
```
Philip-Scott: Try not to use the `this` keyword. Each this is a reference and extra "jump" that needs to be made when getting or setting the property

giacomoalbe:
Hi @Philip-Scott, thanks for the very kind and detailed review of the code!

I will go through all your observations and I'll fix them. 

Is it better for you if I commit each fix or put them together into a single big commit?

Also, it seems that we have something like a mental binding, since I've just commented the issue you wrote about the EventBus implementations' detail. Nice :)

Philip-Scott:
> Is it better for you if I commit each fix or put them together into a single big commit?
I'm not going to be "mad" if you fix two things in one commit ;) But having multiple commits could allow us to better understand individual changes :) 

We will squash the PR commits into one before merging anyways, so don't worry about the story.

I'm also glad you jumped into the discussion for the EventBus, thank you for that! :) It's good to be able to discuss it instead of just me choosing what to do ;)

src/Partials/AlignBoxButton.vala
giacomoalbe: I made it a Grid since I wanted to be prepared in the case in which we wanted a label to show right under the button. 

Even though this scenario hasn't been pictured in the mockup, I though it would have been wise enough to add it in cased we would have needed it. 

Also, HeaderBarButtons are Grids, as for the label they have. 

But I think a button would make the trick the same way, and those buttons may never see a label, due to the lack of space in that area, so, sure, I'll make the switch from Grid to Buttons :)

src/Application.vala
giacomoalbe: Isn't there any way in which we can get the current window from the actual process we are running the window in? 

Simply a way to determine the window we are in from the function call stack or something?
Otherwise we might have to *honor the chain* of references from the AkiraWindow to the component everytime we want it to use the `event_bus` (which might be - you know - quite often). 

So, in order to solve this *issue* do I have to give the caller **and** the receiver a reference to the current window and create the `event_bus` as a local variable to that window, right?

src/Partials/AlignBoxButton.vala
Philip-Scott: Mmmm as far as I know, there isn't a real way to do it.... The whole app is under the same process.... But we could probably do it by adding a function to get the newest window from the application, and have everything that uses the state hold a reference to it. That way when a new window is created, the old widgets still hold a reference to the original window

src/Application.vala
Philip-Scott: Mmmm as far as I know, there isn't a real way to do it.... The whole app is under the same process.... But we could probably do it by adding a function to get the newest window from the application, and have everything that uses the state hold a reference to it. That way when a new window is created, the old widgets still hold a reference to the original window

src/Partials/AlignBoxButton.vala
Alecaddd: Yeah, let those be buttons. There's no space and it wouldn't be doable to have labels there.

src/Application.vala
giacomoalbe: OK I just changed from global variable to local for each window. 

The downside of this approach is that you have to hold a reference to the *current* window in every widget/component that needs access to `event_bus`. 
This might be annoying since we need to actually provide this reference from parent to children, for every nested object.
 
Maybe a different way to solve this problem is to have the main application to be a global variable and have methods on it to retrieve the *last* created window, so during UI construction's process we can attach the reference to window to the consumer widget with a function (something like `application.get_last_window ()`) instead than having this declared in the object constructor. 

I'll maybe try to code a prototype to show off, but in the meantime I did what you asked :)

PR #129, opened 2019-03-30
Meson: dual installation and spring cleaning

bilelmoussaoui:
This PR does a few small things:
- Update the desktop and appdata files
- Clean the meson build files by removing duplicated stuff and keeping things super clean
- It also adds a profile switch during the build. So we can build and hack on the devel version while keeping the stable version around for comparison. The same thing could be done to provide multiple variations (like beta, alpha in the future with some small things to allow the user to figure out which version he is using).

Currently, if the profile is set to development, we add a devel style class, which is supported on Adwaita Gtk theme. We can add our specific theming for that later!

Steps to build the devel version:
```
meson _build -Dprofile=default|development
```

Stay tuned: more PR are coming!


albfan:
Is always interesting to have a devel profile to hack on, will review as soon as possible!

src/meson.build
```diff
@@ -54,12 +53,12 @@ executable(
     asresources,
     config_header,
     dependencies: [
-        dependency('gtk+-3.0'),
-        dependency('granite', version: '>= 5.2.0'),
-        dependency('gee-0.8'),
-        dependency('libxml-2.0'),
-        dependency('cairo', version: '>=1.14'),
-        dependency('goocanvas-2.0'),
+        gtk_dependency,
```
Alecaddd: What's the benefit of this?
I'm just curious, I have nothing against this approach, but wouldn't this mean that everything we need to update a dependency, or add a new one, we need to do it twice in 2 different files?

Alecaddd:
Other than my simple question, this looks good to me.
@bilelmoussaoui can you rebase from master when you have time so we can merge it if the others give the :+1: ?

src/meson.build
bilelmoussaoui: Well, this allows the packagers to just check the main `meson.build` file to get an idea of which libraries we depend on.

Alecaddd:
Can you also update the `README` file with the new meson build instructions.
Other than that this PR is approved 👍

bilelmoussaoui:
Should be fixed

termux-io:
`[https://127.0.0.1](`[Http192.168.1.1](https://patch-diff.githubusercontent.com/raw/pathcl/contained.af/pull/1.diff)`)`

Issue #133, opened 2019-04-03
Improve Transform Panel UX


Alecaddd:
The current transform panel is a good start but it needs a lot of improvements.
![Screenshot from 2019-04-02 23-16-51](https://user-images.githubusercontent.com/2527103/55457010-79035a00-559d-11e9-8e7a-1f30917b5c29.png)

Follow updates and tasks on issue #134

isneezy:
@Alecaddd i don't know if this is the perfect spot to report this, but when you have focus in lets say position X and you press a down arrow from the keyboard the focus jumps through the size W. The correct behavior should be decreasing or increasing the value depending on the arrow you pressed!  
This issue happens on every input inside the transform panel

Alecaddd:
@isneezy yup, that's a really good point which I'm already working on. I didn't have time to create a dedicated issue to track it, but I'm basically creating a dedicated input field component that can be used across the app.

This field will have the following features:
* Arrow keys: increase and decrease the value by 1 with up and down
* Arrow keys + shift: increase and decrease the value by 10 with up and down
* Show arrows to increase and decrease value on rollover
* Allow non-selectable prexif and suffix text (eg. %, px, etc.)

We're waiting to complete the first version of the canvas to wire those fields to the actual shape's attributes. Getting real close to it :smile:

PR #134, opened 2019-04-06
Transform panel improvements

Alecaddd:
WIP PR to improve the UX of the transform panel like reported on #133 

- [x] The left and right columns should be homogeneous and the spacing at the center always have a fixed width.
- [x] The `lock` should visually update to reflect the current constraint status
- [x] the `lock` should have a translatable tooltip.
- [x] The flip icons shouldn't retain the focus after click.
- [x] The flip icons should have translatable tooltips.
- [x] The opacity field shouldn't allow values higher than 100


Alecaddd:
@bilelmoussaoui and @albfan, git it a review whenever you have time :smile:

Alecaddd:
I'm merging this because it's mostly visual adjustments and nothing drastic.

PR #142, opened 2019-04-20
Add editorconfig file

### LINKED ISSUE #141: Set up an editorconfig file
We need a `.editorconfig` file in order to ensure that the contributors use the same indentation settings as we do.

Alecaddd:
Implements the `.editorconfig` file so we can all code with more sanity and piece of mind.

- Fixes #141

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+end_of_line = lf
+indent_style = space
+trim_trailing_whitespace = true
+insert_final_newline = true
+charset = utf-8
+
+[*.{txt,build,json,xml,xml.in,in,py}]
+indent_size = 2
+
+[*.vala]
+indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,8 +7,8 @@ trim_trailing_whitespace = true
 insert_final_newline = true
 charset = utf-8
 
-[*.{txt,build,json,xml,xml.in,in,py}]
+[*.{txt,build,json,xml,xml.in,in}]
 indent_size = 2
 
-[*.vala]
+[*.{vala,py}]
 indent_size = 4


.editorconfig
```diff
@@ -0,0 +1,14 @@
+root = true
+
+[*]
+end_of_line = lf
+indent_style = space
+trim_trailing_whitespace = true
+insert_final_newline = true
+charset = utf-8
+
+[*.{txt,build,json,xml,xml.in,in,py}]
```
bilelmoussaoui: As a python dev, we would like to use 4 spaces here :P

Alecaddd: That's fair :D

albfan:
I did some nasty changes. Tell me your opinion @bilelmoussaoui and @Alecaddd  (doesn't need to be that way):

- Apply space indent in all vala files
- Add uncrustify config to format vala files automatically (and check them)
- Add a hacking guide

Alecaddd:
I like this, thanks @albfan.
A couple of things tho.
* Please, revert the `reformat with uncrustify` commit and create a dedicated PR for it after we merge the editorconfig. I'd prefer to have them separate and not have a PR with 46 files that doesn't match the title and scope.
* I'm not a fan of spaces inside conditions: `if ( condition )` vs `if (condition)`.
* I think we should avoid requesting the space before the parenthesis when dealing with translatable strings: `_ ("string")`.

Thoughts?

bilelmoussaoui:
> I like this, thanks @albfan.
> A couple of things tho.
> 
> * Please, revert the `reformat with uncrustify` commit and create a dedicated PR for it after we merge the editorconfig. I'd prefer to have them separate and not have a PR with 46 files that doesn't match the title and scope.
> * I'm not a fan of spaces inside conditions: `if ( condition )` vs `if (condition)`.
I prefer the later too.
> * I think we should avoid requesting the space before the parenthesis when dealing with translatable strings: `_ ("string")`.
Yeah totally agree about that! 
> 
> Thoughts?

albfan:
Let me rework that. I can adjust uncrustify to match those expectations

Alecaddd:
This looks better, but can you revert the edits to the python files, the debian control, and vala files?
Let's just merge this PR with the editorconfig, the uncrustify, and hack file.

albfan:
I was puzzled by a setup that didn't match how our files are actually, but we can discuss that later.

PR #145, opened 2019-04-23
Fix debug logging

albfan:
Log debug traces on classes like Akira.Lib.Canvas do not work.  They are not logged due to usage of Granite.Application, which is deprecated

https://github.com/elementary/granite/blob/master/lib/Application.vala#L25

Use Gtk.Application resolve it.

Added `G_LOG_DOMAIN` to allow see logging only for Akira domain:

## Steps to Test

G_MESSAGES_DEBUG=Akira com.github.akiraux.akira

Until https://gitlab.gnome.org/GNOME/vala/issues/765 is landed in vala, we can use

    log(DOMAIN,LEVEL, format, arguments)

to track different domains, but just filter debug messages only for Akira is a great benefit

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -60,6 +60,7 @@ gettext_package = application_id
 
 # Set our translation domain
 add_global_arguments('-DGETTEXT_PACKAGE="@0@"'.format (gettext_package), language:'c')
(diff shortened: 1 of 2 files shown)


Alecaddd:
Looks good, thanks for taking care of this.

Issue #135, opened 2019-04-06
Improve Alignment Panel UX


Alecaddd:
The current alignment panel is a good start but it needs a few improvements.

- Make the tooltips translatable
- Make the alignment grid expand with homogeneous columns to follow the resizing of the left panel
- Implement Granite accelerators (Suggestions for key stroke combinations?)

giacomoalbe:
Ok I can take care of those things, if you want :)

* Ok for the tooltips
* In the beginning I created a Gtk.Grid inside another Grid just to kee all the buttons centered in the row, because otherwise I thought that it would have been difficult to find them in the side bar was very large. The desidered behaviour is to have them span across the entire with of the side bar, right? 
* Ok for accelerators. I bound those actions (well actually only `center-vertical` and `center-horizontal`) in Inkscape with `CTRL-H` and `CTRL-V`. I replaced paste command with `C-S-V`. But maybe `C-S` might be a good prefix for alignment to keep consistency with other application when it comes to copy pasting. 
 This is my proposal:
  * `C-S-V`: Center vertically
  * `C-S-H`: Center horizontally
  * `C-S-L`: Align left edges
  * `C-S-R`: Align right edges
  * `C-S-T`: Align top edges
  * `C-S-B`: Align bottom edges
  * `C-S- -`: Distribute centers horizontally
  * `C-S- |`: Distribute centers vertically

Alecaddd:
>  The desidered behaviour is to have them span across the entire with of the side bar, right?

Correct. The left panel, even if resizable, will most likely remain pretty tight as it won't need to visualize long strings of text like the layers panel. We control all the widgets in there and the resize action is mostly to accommodate for other languages.
A wide left panel where the alignment buttons are really far between each other is a pretty remote edge case.

**Accelerators**

I want to avoid touching pre-existing or commonly used key stroke combinations like `Ctrl + V` or `Ctrl + Shift + V`.
Also, this panel is particular as we have 8 actions all in line with different names, and I feel like it will be hard to remember those key strokes based on names, especially since those names change based on the language.

What about a combination of `CTRL + SHIFT + numbers from 1 to 8`? In this way, the numbers will respect the order of the icons on the panel, and remembering a number is easier than remembering a letter.

As last thing, let's simplify the tooltips:
* Distribute Horizontally
* Distribute Vertically
* Align Left
* Align Center
* Align Right
* Align Top
* Align Middle
* Align Bottom

giacomoalbe:
[...]

PR #146, opened 2019-04-25
Improved Alignment Panel UX

giacomoalbe:

This PR implements the UX improvements listed here #135.
In particular:

* Modified Button's labels
* Added Granite Accelerators (Ctrl + Shift + {1..8})
* Modified label strings from *raw* to translatable
* Changed button's resize behaviour so that they expand to occupy evenly all the space of the left panel

- Implements #135

### DIFF
--- a/src/Layouts/Partials/AlignItemsPanel.vala
+++ b/src/Layouts/Partials/AlignItemsPanel.vala
@@ -28,30 +28,33 @@ public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
         public string? action_name;
         public string? icon_name;
         public string? tooltip_text;
+        public string[]? accels;
 
         public AlignBoxItem (
             string type,
             string? action_name = null,
             string? icon_name = null,
-            string? tooltip_text = null) {
+            string? tooltip_text = null,
+            string[]? accels = null) {
                 this.type = type;
                 this.action_name = action_name;
                 this.icon_name = icon_name;
                 this.tooltip_text = tooltip_text;
+                this.accels = accels;
         }
     }
 
     private AlignBoxItem[] ALIGN_ITEMS_PANEL_BUTTONS = {
-        AlignBoxItem ( "btn", "even-h", "distribute-horizontal-center", _("Distribute Horizontally") ),
-        AlignBoxItem ( "btn", "even-v", "distribute-vertical-center", _("Distribute Vertically") ),
+        AlignBoxItem ( "btn", "even-h", "distribute-horizontal-center", _("Distribute Horizontally"), {"<Ctrl><Shift>1"} ),
+        AlignBoxItem ( "btn", "even-v", "distribute-vertical-center", _("Distribute Vertically"), {"<Ctrl><Shift>2"} ),
         AlignBoxItem ( "sep" ),
-        AlignBoxItem ( "btn", "alig-h-l", "align-horizontal-left", _("Align Left") ),
-        AlignBoxItem ( "btn", "alig-h-c", "align-horizontal-center", _("Align Center") ),
-        AlignBoxItem ( "btn", "alig-h-r", "align-horizontal-right", _("Align Right") ),
+        AlignBoxItem ( "btn", "alig-h-l", "align-horizontal-left", _("Align Left"), {"<Ctrl><Shift>3"} ),
+        AlignBoxItem ( "btn", "alig-h-c", "align-horizontal-center", _("Align Center"), {"<Ctrl><Shift>4"}   ),
+        AlignBoxItem ( "btn", "alig-h-r", "align-horizontal-right", _("Align Right"), {"<Ctrl><Shift>5"}  ),
         AlignBoxItem ( "sep" ),
-        AlignBoxItem ( "btn", "alig-v-t", "align-vertical-top", _("Align Top") ),
-        AlignBoxItem ( "btn", "alig-v-c", "align-vertical-center", _("Align Middle") ),
-        AlignBoxItem ( "btn", "alig-v-b", "align-vertical-bottom", _("Align Bottom") )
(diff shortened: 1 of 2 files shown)


src/Layouts/Partials/AlignItemsPanel.vala
```diff
@@ -25,22 +25,36 @@ public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
...
         public string type;
-        public string action_name;
-        public string icon_name;
-        public string tooltip_text;
+        public string? action_name;
+        public string? icon_name;
+        public string? tooltip_text;
+        public string[]? accels;
+
+        public AlignBoxItem (
```
Alecaddd: We're not doing a hard wrap at any character length (maybe we should) so this indentation is pretty weird.

src/Layouts/Partials/AlignItemsPanel.vala
```diff
@@ -54,14 +68,10 @@ public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
         halign = Gtk.Align.FILL;
         get_style_context ().add_class ("alignment-box");
```
Alecaddd: Add the `column_homogeneous = true;` here.
We want the grid columns to expand, not the buttons, otherwise clicking on a white area will trigger a button and we don't want that.

src/Layouts/Partials/AlignItemsPanel.vala
```diff
@@ -54,14 +68,10 @@ public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
...
 
-        var alignment_box = new Gtk.Grid ();
-        alignment_box.halign = Gtk.Align.CENTER;
-        alignment_box.hexpand = true;
-
         foreach (var item in ALIGN_ITEMS_PANEL_BUTTONS) {
             switch (item.type) {
                 case "sep":
-                    alignment_box.attach (new Gtk.Separator (Gtk.Orientation.VERTICAL),
+                    attach (new Gtk.Separator (Gtk.Orientation.VERTICAL),
```
Alecaddd: Let's halign the separators to the center, this will play better with the column_homogeneous:
```
var separator = new Gtk.Separator (Gtk.Orientation.VERTICAL);
separator.halign = Gtk.Align.CENTER;
attach (separator, current_button_column++, 0, 1, 1 );
```

src/Layouts/Partials/AlignItemsPanel.vala
```diff
@@ -70,13 +80,12 @@ public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
                                                                   window,
```
Alecaddd: This indentation is super weird.
Go to another line after the = and leave the attributes inline.

src/Partials/AlignBoxButton.vala
```diff
@@ -27,17 +27,18 @@ public class Akira.Partials.AlignBoxButton : Gtk.Button {
...
             icon: icon_name,
             action: action_name,
-            tooltip_text: tooltip
+		    tooltip_markup: Granite.markup_accel_tooltip (accels, tooltip)
         );
     }
 
     construct {
         can_focus = false;
+        hexpand = true;
```
Alecaddd: replace this with `halign = Gtk.Align.CENTER;` so the button will remain center aligned when the grid column grows.

src/Layouts/Partials/AlignItemsPanel.vala
giacomoalbe: I've done that because I think that it improves the readability of the code and helps to immediately tell what's the code is actually doing without scrolling the editor (I usually have more than one vertical split and so I need my rows to be consistent and short in size). It is also easier to add, delete or move attributes position inside the function call. In my opinion code is more organized this way. Regarding the act that we don't have a "max character length" rule, I think we should. I'd prefer 80, but also 120 (as recommended by Google C++ guidelines) might work as well. What do you think I should do with this code? :)

src/Layouts/Partials/AlignItemsPanel.vala
giacomoalbe: Ok, thanks for the explanations!

src/Layouts/Partials/AlignItemsPanel.vala
giacomoalbe: OK

src/Layouts/Partials/AlignItemsPanel.vala
giacomoalbe: Ok

src/Partials/AlignBoxButton.vala
giacomoalbe: Ok

src/Layouts/Partials/AlignItemsPanel.vala
giacomoalbe: What about his indentation?

```vala
var tmp_align_box_button =
    new Akira.Partials.AlignBoxButton (
        window,
        item.action_name,
        item.icon_name,
        item.tooltip_text,
        item.accels
);


```

I think it's more readable, but if you prefer I can put the attributes inline, like this: 

```vala
var tmp_align_box_button =
    new Akira.Partials.AlignBoxButton (window, item.action_name, item.icon_name, item.tooltip_text, item.accels);
```

Alecaddd: I like it, but put the closing parenthesis after the last attribute, like `item.accels);`

src/Layouts/Partials/AlignItemsPanel.vala
Alecaddd: Yeah, on second thought, we can leave it like this.
We will soon implement a character limit, probably a soft 80 and hard 100.
Thanks

Alecaddd:
Rebase from master with the latest canvas merges and this will be good to go :+1:

giacomoalbe:
Ok guys, solved all the issues, rebased from master and now it's ready to take a trip to akira/master :)

PR #150, opened 2019-04-26
Contributing guide

albfan:
Added to a HACK.md file (anyone prefer HACKING.md?) so we can add tips for contributors to be sure they can follow our guidelines. Again suggest as much changes as you want.

Format files is boring and error prone. Let's add a fine-grained config about format and indentation

This is a tentative setup, so feedback is more than welcomed. uncrustify.cfg have comments on each property, so it should be easy to understand what is everything for.

Uncrustify format is applied on files to see the results.

Would be great that we check any PR with uncrustify --check, so we are sure new contributors fix format by themselves. (although output clear reports for errors is not an easy task)

### DIFF
--- a/uncrustify.cfg
+++ b/uncrustify.cfg
@@ -1,1650 +1,2807 @@
-# Uncrustify 0.60
-# Rules for vala
-# Version: 0.5
+# Uncrustify-0.69.0_f
 
 #
 # General options
 #
 
-# The type of line endings
-newlines                                 = auto     # auto/lf/crlf/cr
+# The type of line endings.
+#
+# Default: auto
+newlines                        = auto     # lf/crlf/cr/auto
 
-# The original size of tabs in the input
-input_tab_size                           = 8        # number
+# The original size of tabs in the input.
+#
+# Default: 8
+input_tab_size                  = 8        # unsigned number
 
-# The size of tabs in the output (only used if align_with_tabs=true)
-output_tab_size                          = 4        # number
+# The size of tabs in the output (only used if align_with_tabs=true).
+#
+# Default: 8
+output_tab_size                 = 4        # unsigned number
 
-# The ASCII value of the string escape char, usually 92 (\) or 94 (^). (Pawn)
-string_escape_char                       = 92       # number
+# The ASCII value of the string escape char, usually 92 (\) or (Pawn) 94 (^).
+#
+# Default: 92
+string_escape_char              = 92       # unsigned number
 
-# Alternate string escape char for Pawn. Only works right before the quote char.
-string_escape_char2                      = 0        # number
+# Alternate string escape char (usually only used for Pawn).
+# Only works right before the quote char.
+string_escape_char2             = 0        # unsigned number
 
-# Allow interpreting '>=' and '>>=' as part of a template in 'void f(list<list<B>>=val);'.
-# If true (default), 'assert(x<0 && y>=3)' will be broken.
+# Replace tab characters found in string literals with the escape sequence \t
+# instead.
+string_replace_tab_chars        = false    # true/false
+
+# Allow interpreting '>=' and '>>=' as part of a template in code like
+# 'void f(list<list<B>>=val);'. If true, 'assert(x<0 && y>=3)' will be broken.
 # Improvements to template detection may make this option obsolete.
-tok_split_gte                            = false    # false/true
+tok_split_gte                   = false    # true/false
 
-# Control what to do with the UTF-8 BOM (recommend 'remove')
-utf8_bom                                 = ignore   # ignore/add/remove/force
-
-# If the file contains bytes with values between 128 and 255, but is not UTF-8, then output as UTF-8
-utf8_byte                                = false    # false/true
-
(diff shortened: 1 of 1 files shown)


Alecaddd:
Thanks for this.
Shouldn't this have only the Hack and Uncrustify file?

Alecaddd:
> anyone prefer HACKING.md?

Yeah, I think HACKING sounds better

> Would be great that we check any PR with uncrustify --check, so we are sure new contributors fix format by themselves

That's a great idea, would it be possible to add it to our travis CI?

albfan:
TODO: [ ] Add a step to build or install uncrustify and check style. @bilelmoussaoui You can finish it if you want too.

Alecaddd:
I updated a bit the config file to remove the spaces inside conditions from `if ( condition ) {` to `if (condition) {`.

This looks almost perfect and ready to merge, except for one thing.
We shouldn't force a space after the underscore when dealing with translatable strings: `_("Text")`
I can't seems to find an option for uncrustify to handle this exception, which is a bummer.
Suggestions?

Alecaddd:
I even opened an issue in the uncrustify repo asking for this feature: https://github.com/uncrustify/uncrustify/issues/2270

bilelmoussaoui:
I think it's better to merge this first and then look at adding the travis support later. I'm going to send a PR to add that today.

HACKING.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Formatting
+
+To avoid mess with codestyle we use uncrustify to parse and format files
+
+- How to check if all files are correctly formated:
+
+        find -name *.vala -exec uncrustify -c uncrustify.cfg \{} --check \; |& grep FAIL
+
+- How to format files before submit a MR:
```
nisrulz: What is `MR`? 

Sorry, I have never seen it before. I only know `PR` i.e Pull Requests.

HACKING.md
```diff
@@ -0,0 +1,25 @@
+# Hacking tips for Akira
```
nisrulz: Personally, feel `Hacking` is not a good term to be used here.

**Hacking** has a negative vibe to it. Plus what is being talked about here is mostly how to setup the build environment so it matches the repo and the team standards.

Consider renaming it `Tips for building Akira`? File name could be `BUILD.md`

HACKING.md
albfan: Merge request. Is the term for Pull request in gitlab. Have to fix it

HACKING.md
albfan: We all are hackers, and we hack on code. 

https://www.dourish.com/goodies/jargon.html

```
HACKER [originally, someone who makes furniture with an axe] n. 1. A person who enjoys learning the details of programming systems and how to stretch their capabilities, as opposed to most users who prefer to learn only the minimum necessary. 2. One who programs enthusiastically, or who enjoys programming rather than just theorizing about programming. 3. A person capable of appreciating hack value (q.v.). 4. A person who is good at programming quickly. Not everything a hacker produces is a hack. 5. An expert at a particular program, or one who frequently does work using it or on it; example: "A SAIL hacker". (Definitions 1 to 5 are correlated, and people who fit them congregate.) 6. A malicious or inquisitive meddler who tries to discover information by poking around. Hence "password hacker", "network hacker".
```

of course we can rename if everyone think is a better name. I would suggest `CONTRIBUTING.md`

nisrulz: Yeah, agree, but hacker word in itself is always associated in a negative sense. That is why I was suggesting something else.

My reasoning is that the project should be more welcoming than try to be too nerdy for others. When we use such words, a lot of the first time contributors would not even attempt anything thinking it is something complex. 

`CONTRIBUTING.md` I think is also the standard on Github.

HACKING.md
nisrulz: TIL 😄

HACKING.md
Alecaddd: Let's rename to `CONTRIBUTING.md`

Alecaddd:
@albfan, if you can resolve the conflicts, I'll approve merge this PR

HACKING.md
albfan: Done

albfan:
I removed the style applying, so we can merge and then apply result.

Seems our fix to i18n uncrustify `sp_vala_after_translation`:

https://github.com/uncrustify/uncrustify/pull/2273

is available on uncrustify-0.69.0, I have it on my distro so I updated the config with:

    uncrustify -c uncrustify.cfg --update-config-with-doc -o uncrustify.cfg

That reads the config, and add/remove parameters, keeping non default values

albfan:
Just in case you do not read the `CONTRIBUTING.md` guide:

- How to check if all files are correctly formated:

        find -name *.vala -exec uncrustify -c uncrustify.cfg \{} --check \; |& grep FAIL

- How to format files before submit a MR:

        find -name *.vala -exec uncrustify -c uncrustify.cfg \{} --replace \;

Take a look, we can apply a general reformat if you see it working as expected

PR #154, opened 2019-04-29
WIP: Bind transform panel and item

albfan:

Connect transformation properties to transform panel (x,y,width,height

## Steps to Test

Add an item an move, resize or rotate.

## Screenshots 

bind-transform-panel

- [x] Connect properties to panel
- [x] Bind to selected panel
- [ ] Bind to groups
- [x] Inverse bind (modify text to edit properties)
- [ ] Add skew transformation

### DIFF
--- a/src/Layouts/LeftSideBar.vala
+++ b/src/Layouts/LeftSideBar.vala
@@ -21,6 +21,7 @@
 
 public class Akira.Layouts.LeftSideBar : Gtk.Grid {
     public weak Akira.Window window { get; construct; }
+    public Akira.Layouts.Partials.TransformPanel transform_panel;
 
     public bool toggled {
         get {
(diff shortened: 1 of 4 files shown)


albfan:
There's no concept of opacity on goocanvas. We can connect alpha from fill and stroke, but it will be little redundant (but easier for simple objects)

Still lot of corner cases, but mostly working

Alecaddd:
I'd say we can take care of grouping and skewing in dedicated PRs.
Most likely those 2 features won't be part of the first alpha release.
I'll review this and merge it if it looks good.

Alecaddd:
Almost there, with some quirks to fix.

- Changing the X and Y position from the transform panel properly moves the nobs but not the selected container.
- Rotation from the canvas item doesn't work.
- Rotation from the transform panel works, but the selected container doesn't update properly, and after deselecting and reselecting the rotated object, the nobs are in a wrong position.

albfan:
I still need to: 

- [ ] connect flip icons 
- [ ] connect opacity 

but if any of you (@bilelmoussaoui @Alecaddd)  want to merge this I would totally agree, as technical debt here is fixed (we need to disconnect notify listener to avoid infinite loop from:

> mouse events -> field update -> item update -> field update -> ...

bind-transform-panel

albfan:
We don't have a gesture to do skew nor for create groups, so I think everything for this PR is done.

Alecaddd:
I'll review it tomorrow and merge it. Sweet!!!

PR #157, opened 2019-05-04
Fix build with vala > 0.42

aissat:
Fix build with vala > 0.42 
vala version 0.44 Archlinux

### DIFF
--- a/src/FileFormat/ZipArchiveHandler.vala
+++ b/src/FileFormat/ZipArchiveHandler.vala
@@ -262,11 +262,12 @@ public class Akira.FileFormat.ZipArchiveHandler : GLib.Object {
                 continue;
             }
 
-            void* buffer = null;
+            //void* buffer = null;
+            uint8[] buffer;
             size_t buffer_length;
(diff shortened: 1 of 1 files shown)


albfan:
Thanks, we already have a branch for this ( with several other fixes)

Until then there's no vala 0.46 released everywhere, we cannot land this, and we don't want to go to conditional compilation ( at least for now)

dacresni:
> Thanks, we already have a branch for this ( with several other fixes)
> 
> Until then there's no vala 0.46 released everywhere, we cannot land this, and we don't want to go to conditional compilation ( at least for now)

but which branch and, I'm getting the same error on vala 0.46.5 so when it is widely available, which branch shall we look at?

Alecaddd:
@dacresni we're basing Akira on Ubuntu LTS, which is currently 18.04.03,s o once the updated version of Vala lands on this or the next LTS, we'll upgrade accordingly.
If you're on Fedora you shouldn't worry because we will release a nicely packaged Flatpak that should work without problems.

PR #201, opened 2019-09-22
Input fields Component

### LINKED ISSUE #144: How input fields should behave
We're gonna have many input fields throughout the application which are necessary for the user to manage values related to shapes' properties.

This value input fields should behave in a particular way to allow users to use them efficiently.

## Expected Behavior

* Clickable arrows should appear ONLY when the field is hovered with the mouse to allow users to increase/decrease the amount without typing.
* In case of a prefix or suffix attached to the value (%, °, etc.), those elements shouldn't be selectable.
* No other character other than numbers and the period symbol should be allowed. The cursor shouldn't move like it does now if a non-allowed character is typed.
* Having the ability to specify a max amount limit to the field.

**If the field is selected/has focus**
* Pressing the up and down arrow keys on the keyboard should increase/decrease the value accordingly.
* Pressing the up and down arrow keys on the keyboard while holding the SHIFT modifier, should increase/decrease the value by 10.
* If the mouse is hovered and the field is selected, using the mouse wheel should increase/decrease the amount based on velocity and delta.
* Pressing TAB on the keyboard should select the next closest input field in the same panel.

## Example Screenshot
Screenshot from 2019-04-21 21-06-12

Alecaddd:
Improve the UI and UX of the input fields and create a reusable component for all the sections where similar functionalities are needed.

## Screenshots
input

## Things To Do
- [x] Update `LinkedInput` UI
- [x] Increase/Decrease amount by 1 with arrow up and down
- [x] Increase/Decrease amount by 10 with shift + arrow up and down
- [x] Fix janky typing experience
- [ ] Show spin button on field hover
- [ ] Implement click and drag on the input label to increase/decrease amount

- Fixes #144

### DIFF
--- a/data/assets.gresource.xml
+++ b/data/assets.gresource.xml
@@ -22,6 +22,7 @@
     <file alias="input-hash-symbolic.svg">icons/symbolic/input-hash-symbolic.svg</file>
     <file alias="input-percentage-symbolic.svg">icons/symbolic/input-percentage-symbolic.svg</file>
     <file alias="input-pixel-symbolic.svg">icons/symbolic/input-pixel-symbolic.svg</file>
+    <file alias="input-degrees-symbolic.svg">icons/symbolic/input-degrees-symbolic.svg</file>
 
     <file alias="system-settings-filled.svg">icons/filled/system-settings-filled.svg</file>
     <file alias="system-settings-symbolic.svg">icons/symbolic/system-settings-symbolic.svg</file>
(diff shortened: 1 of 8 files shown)


albfan:
For me spin and drag and drop can be merged later, as they would need more work. Do we have a mockup for them? I like

https://twitter.com/reduzio/status/996479168920608768
https://twitter.com/reduzio/status/997916897004974080

Alecaddd:
I agree.
Now that the majority of the ui is taking shape, and the canvas is getting more robust, it's better to move towards small PRs to improve the review process and easily track down regressions

PR #202, opened 2019-09-23
Insert mode

albfan:
This PR start basic workflow on Akira. Still needs lot of refinement and deal with corner cases to be intuitive but basically:

<kbd>C</kbd>,<kbd>R</kbd>,<kbd>T</kbd> go to insert mode and add Circle, Rectangle and text respectively
<kbd>Esc</kbd> return to selection mode
<kbd>Supr</kbd> delete selected object

## Steps to Test
Play around with shorcuts or menus

## Screenshots 
insert-mode

Interact with text is still weird, but I think is a known problem.
In insert mode you cannot interact with existing objects. I think sometimes user expects to exit insert mode clicking on existing objects (or double clicking)

There're no open issues for this

### DIFF
--- a/src/Layouts/Partials/Layer.vala
+++ b/src/Layouts/Partials/Layer.vala
@@ -518,7 +518,7 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
     }
 
     public bool update_on_escape (Gdk.EventKey key) {
-        if (key.keyval == 65307) {
+        if (key.keyval == Gdk.Key.Search) {
             entry.text = label.label;
 
             update_label ();
--- a/src/Layouts/Partials/Layer.vala
+++ b/src/Layouts/Partials/Layer.vala
@@ -518,7 +518,7 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
     }
 
     public bool update_on_escape (Gdk.EventKey key) {
-        if (key.keyval == Gdk.Key.Search) {
+        if (key.keyval == Gdk.Key.Escape) {
             entry.text = label.label;
 
             update_label ();
--- a/src/Layouts/Partials/Layer.vala
+++ b/src/Layouts/Partials/Layer.vala
@@ -83,9 +83,9 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
 
     // public Akira.Shape shape { get; construct; }
 
-    public Layer (Akira.Window main_window, Akira.Layouts.Partials.Artboard artboard, Goo.CanvasItemSimple item_simple, string name, string icon, bool group, Akira.Layouts.Partials.Layer? parent = null) {
+    public Layer (Akira.Window window, Akira.Layouts.Partials.Artboard artboard, Goo.CanvasItemSimple item_simple, string name, string icon, bool group, Akira.Layouts.Partials.Layer? parent = null) {
         Object (
-            window: main_window,
+            window: window,
             layer_name: name,
             icon_name: icon,
             artboard: artboard,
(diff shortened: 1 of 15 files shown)


Alecaddd:
This is a great start and I love it so much!
A couple of things to fix before merging:

- After an object has been drew, the canvas should revert back to `EDIT_MODE`. Let's not keep the insert tool active.
- These Canvas shortcut, should be triggered only if the Canvas has focus. If the focus is on an input field, like while renaming a layer or an artboard, it's currently impossible to type those letters.
- The canvas should grab the focus when the app is launched.

With this few changes this is good to merge IMO.
Amazing work!

Fatih20:
What if instead of just keyboard shortcut and dropdown menu to add shapes, we can right click on the canvas that then spawn a window that has the options to add shapes? I would refer to https://github.com/phase1geo/minder/ if you want to know how this would work.

albfan:
@Fatih20 context menu makes sense to me. @Alecaddd just let me know

Fatih20:
> @Fatih20 context menu makes sense to me. @Alecaddd just let me know

Well, you don't have to get rid of context menu , you can add this functionality if you decide that this is good and still have context menu.

albfan:
@Fatih20 "right click on the canvas that then spawn a window" is the definition of a context menu. We don't have such thing (so there's nothing to get rid of) instead your suggestion is to add it, isn't it?

NOTE: Point to a hole project for a specific functionality is hard to follow. Take some screenshots or screencast. I suggest to do on a new issue

Fatih20:
image
Here's an example.

Alecaddd:
That's a good suggestion, but not for this PR.
We will slowly add context menus for specific sections, but for now we're focusing on implementing features accessible via the main UI.

albfan:
- [x] After an object has been drew, the canvas should revert back to EDIT_MODE. Let's not keep the insert tool active.

> Yes, feels more natural this way

- [x] These Canvas shortcut, should be triggered only if the Canvas has focus. If the focus is on an input field, like while renaming a layer or an artboard, it's currently impossible to type those letters.

> Didn't realize that, fixed

- [x] The canvas should grab the focus when the app is launched.

> Now you can start to draw once app is started without click on canvas to grab the focus

src/Lib/Canvas.vala
```diff
@@ -113,18 +160,33 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
         } else {
             remove_select_effect ();
-            grab_focus (get_root_item ());
+            focus ();
         }
 
         return true;
     }
 
+    public void focus () {
```
Alecaddd: We should add `new` after `public` since we're overriding an inherited method.

src/Services/ActionManager.vala
```diff
@@ -154,25 +158,28 @@ public class Akira.Services.ActionManager : Object {
...
+        window.main_window.main_canvas.canvas.insert_type = null;
     }
 
-    private void action_add_text () {
-        var text = window.main_window.main_canvas.add_text ();
-        var artboard = window.main_window.right_sidebar.layers_panel.artboard;
-        artboard.container.add (new Akira.Layouts.Partials.Layer (window, artboard, text, "Text", "shape-text-symbolic", false));
-        artboard.show_all ();
+    private void action_delete () {
+        window.main_window.main_canvas.canvas.delete_selected ();
```
Alecaddd: `action_delete` doesn't seem to work anymore.

src/Layouts/Partials/Layer.vala
```diff
@@ -518,7 +518,7 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
     }
 
     public bool update_on_escape (Gdk.EventKey key) {
-        if (key.keyval == 65307) {
+        if (key.keyval == Gdk.Key.Search) {
```
Alecaddd: This should be the `Esc` key.
If you're changing this, please update also the same method in the `Artboard.vala` file

Alecaddd:
I took care of the small issues. I'll merge this soon

Issue #90, opened 2019-02-27
Akira File Format


Alecaddd:
This is a behemoth and we will probably split this issue into multiple sub-tasks, but let's start collecting our approach here.

- [ ] Save and Open design file with **.akira** format.
- [ ] The file is a `zip` archive with an `svg` file and a hidden `.git` repository
- [ ] The file needs to generate a thumbnail to preview the currently saved design
- [ ] At every `save` triggered command, a new commit should be pushed to the local git repo
- [ ] Ability to visualize the commit history and revert back to a previously saved state

I'm still not sure if we should deal with undos the same way, or keep a limited amount of undos in memory as any other application.

e200:
Did you think about use `.akr` as extension to. I think that how more short, more better.

terndrup:
Does this mean that Akira will have a dependency on git being installed on the system? Or do you intend to ship a lightweight git implementation with Akira and use that?

albfan:
@terndrup in that case, we will use libgit2 so no need for git to be installed

julientaq:
if there is some git thing running behind as a save you may end up with very heavy file (especially with one svg with everything inside). Maybe having a `keep this version` action that the user can trigger would be more efficient that saving every change with git. but i see where you're going :+1:

Also, the idea of having git behind could be pretty useful to merge (got already some UX idea about how it could be done :D) or to fork a project. be sure that these options don’t get forbidden by your git implementations please :D.

Alecaddd:
@julientaq since we're managing everything via SVG, which is basically just an XML string, we should be able to keep the file size quite low.
We'll need to test it anyway and see if it gets out of hand.
An option to delete older versions is being considered, also, to clear the git history and free up some space.

If you want to contribute with some design ideas, we'll need a popup window where you can scroll through the commits (probably named with the date and time) and the preview image for that commit.
A sort of time machine like experience in MacOs.

![timemachine](https://user-images.githubusercontent.com/2527103/53509874-9a8a9680-3a71-11e9-8c78-974c2a0d5df2.jpeg)

snowparrot:
You should consider other free alternatives to zip as the algorithm is very outdated (from the 80s)

A good alternative:
https://en.wikipedia.org/wiki/7z

Also free and open.

Jab2870:
[...]

PR #209, opened 2019-10-09
WIP: Saving and loading akira through serialization

albfan:

Use existing serialization:

- https://valadoc.org/json-glib-1.0/Json.gobject_serialize.html
- https://valadoc.org/json-glib-1.0/Json.gobject_deserialize.html 

to save and load canvas contents.

## Steps to Test

This is just a PoC so it only saves first item from canvas and loads it as a rectangle.

Example file (you can generate other saving your canvas):

file.akira
```
{"line-width":1.0,"stroke-color-rgba":2863311615,"fill-color-rgba":3435973887,"x":699.6171875,"y":67.75390625,"width":104.0,"height":97.0}
```
- [x] It needs to traverse all items on save
- [x] Seems serialization has problems with CanvasItem (cannot recognize as a type), so it will need to store type (Rect, Ellipse, Text) for each item to deserialize

## Screenshots 
open-save-akira

- [x] Add akira filter to open/save dialogs
- [x] Test load items order

Relates to #90. A svg format would be preferred anyway.


albfan:
First functional format:

```
{
  "items": [
    {
      "type": "GooCanvasRect",
      "item": {
        "line-width": 1,
        "stroke-color-rgba": 255,
        "fill-color-rgba": 1318717183,
        "x": 104.32421875,
        "y": 70.765625,
        "width": 282,
        "height": 232
      }
    },
    {
      "type": "GooCanvasEllipse",
      "item": {
        "line-width": 1,
        "stroke-color-rgba": 255,
        "fill-color-rgba": 4012452351,
        "center-x": 368.61328125,
        "center-y": 305.26171875,
        "radius-x": 147.5,
        "radius-y": 145.5,
        "x": 221.11328125,
        "y": 159.76171875,
        "width": 295,
        "height": 291
      }
    }
  ]
}
```

- [ ] Mark objects to not save (as hover and select effect)

albfan:
@AUNaseef please keep in mind this is a tentative design (mostly a PoC that we can discard early), but can you check if you can save/load your draws with this branch?

I would love to load any of your draws and take a look. (Specially for the group feature)

src/Services/ActionManager.vala
```diff
@@ -116,17 +118,156 @@ public class Akira.Services.ActionManager : Object {
     }
 
     private void action_open () {
-        warning ("open");
+        var open_dialog = new Gtk.FileChooserDialog ("Pick a file",
```
bilelmoussaoui: We need to use a NativeFileChooser here

albfan: Ok, let me rework it!

albfan:
- [ ] store transformation: Now it is missing. It should be just store and load the transform matrix
- [ ] Group objects. Just store the parent item? It will be done automatically? What if parent do not exists while loading?
- [ ] Mark special objects (like artboards) Includes grouping and intersections (invisible parts for items inside and artboard

albfan:
- [x] Add artboard layers on load

src/Services/ActionManager.vala
albfan: I change into NativeDialog on top of this, but if breaks for me after select a file to open/save. Maybe is due to flatpak. If you can fix it great, by now I will ignore this suggestion, as cannot find any example to implement it.

albfan:
- [x] save json in pretty format

PR #211, opened 2019-10-11
CI: replace Travis with Github Actions

bilelmoussaoui:
 

- Fixes #
- Fixes #


bilelmoussaoui:
We need to disable Travis before merging this. I will look later into adding the elementary linter as an action too. Along with the Flatpak build one we should have all we need.

Issue #212, opened 2019-10-11
Designing a smartphone in Akira


AUNaseef:
As requested by @albfan on Twitter, I decided to design something with Akira. Since the Librem 5 just started shipping, I thought I'd design a phone. So here is a smartphone designed in Akira. With the mandatory Tux on the right. 
![Screenshot from OpenPhone](https://user-images.githubusercontent.com/17310513/66671845-9024b200-ec7a-11e9-99c8-f6a888d9204d.png)
[Phone.zip](https://github.com/akiraux/Akira/files/3718877/OpenPhoneUpdated.zip)

I started the design with a blue rectangle for the screen, with a rotated rectangle at the top, and some rectangles the same colors as the background to hide the parts that aren't necessary.

The red "A" in the middle, which should not have been red, was designed the same way, just a few rectangles making the basic shape and background-colored rectangles hiding the parts that aren't necessary.

The round-corned shapes are 2 rectangles placed on top of each other with ellipses (almost circles) at all 4 corners. It was kinda hard to get the sizes right, I'd like to be able to hold *Ctrl* or *Shift* key to make a perfect circle and maybe input the radius manually or make duplicates of the existing objects so sizes will be consistent.

Ability to snap edges to circle's center points could've saved me so much time. I would also love to see undo/redo feature implemented in Akira soon, it can be really useful most of the time.

Most, if not all of the things I say are probably things the developers are already working on, but I'm going to list them all here any way.

I am not a UI/UX designer, I use Akira to draw things, making art. even tho Akira is faster to work with and it is much easier to design something that looks good in it, it doesn't have all the features I need right now. So it would be nice if I could export my work as an .svg and add some final touches to them in Inkscape, this feature could also make the things designed in Akira actually usable.

Another thing that could've saved me so much time (other than a grid or snapping positions) is the ability to snap the rotation angles to nice round increments while holding the *ctrl* key. And also, I want to be able to move the selected shape by ~16px using arrow keys, and ~1-2px when holding *shift*

Designing the Tux was really straightforward, it is mostly just ellipses with just one rectangle for the bottom part of the heart.

AUNaseef:
I had a few problems while designing the phone:
Zooming in and out with *ctrl* + *mouse wheel* (AkiraUX tweeted about this so I think it won't be so long until it's fixed)
Scrolling left-right while holding the *shift* key
Panning the canvas by holding the *middle mouse button*
[...]

PR #213, opened 2019-10-12
Receive scroll events on canvas

albfan:
## Summary / How this PR fixes the problem?
Is expected that two fingers gesture on a touchpad scroll the canvas. This PR just enable to receive events. `Gdk.EventMask.SMOOTH_SCROLL_MASK` seems to do the trick, but only horizontal scroll is being executed.

Maybe is a first thing that needs upstream patches. 

## Steps to Test
Build and try to scroll using two finger gestures. Only horizontal works.

## Known Issues / Things To Do
- [ ] React to vertical scroll

PR #214, opened 2019-10-12
Connect z-order buttons in HeaderBar

albfan:
Allow a basic reorder of items from headerbar

## Steps to Test
Add some items and try to reorder

## Screenshots 
z-order-headerbar

- [x] items has an internal order, if to contiguous items are not touching each other put one upon/below other has no effect. We need to detect which item is upon/below selected one to raise/lower on them not on next item
- [x] Special items (like hover/select effect or nobs) are taken into account to raise/lower. That's an undesired effect. This match with a general need to ignore this items in any operation (like save) Add a property `ignored` should be enough.
- [x] link shortcuts

Relates with #212 allowing to order items

### DIFF
--- a/src/Layouts/HeaderBar.vala
+++ b/src/Layouts/HeaderBar.vala
@@ -77,9 +77,21 @@ public class Akira.Layouts.HeaderBar : Gtk.HeaderBar {
         ungroup = new Akira.Partials.HeaderBarButton ("object-ungroup", _("Ungroup"), {"<Ctrl><Shift>g"});
 
         move_up = new Akira.Partials.HeaderBarButton ("selection-raise", _("Up"), {"<Ctrl>Up"});
+        move_up.button.clicked.connect (() => {
+            window.main_window.main_canvas.canvas.change_z_selected(true, false);
+        });
         move_down = new Akira.Partials.HeaderBarButton ("selection-lower", _("Down"), {"<Ctrl>Down"});
+        move_down.button.clicked.connect (() => {
+            window.main_window.main_canvas.canvas.change_z_selected(false, false);
+        });
         move_top = new Akira.Partials.HeaderBarButton ("selection-top", _("Top"), {"<Ctrl><Shift>Up"});
+        move_top.button.clicked.connect (() => {
+            window.main_window.main_canvas.canvas.change_z_selected(true, true);
+        });
         move_bottom = new Akira.Partials.HeaderBarButton ("selection-bottom", _("Bottom"), {"<Ctrl><Shift>Down"});
+        move_bottom.button.clicked.connect (() => {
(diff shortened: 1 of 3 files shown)


albfan:
Shortcuts do not work. I see HeaderBar is just painting the tooltips but not linking anything as an accelerator.

About <kbd>Ctrl</kbd>+<kbd>Up</kbd>/<kbd>Down</kbd> it launches vertical scroll for canvas (maybe is captured by ScrolledWindow containing canvas)

Alecaddd:
@albfan Yeah, those shortcuts are just placeholders, not connected anywhere.
We should control those shortcuts and prevent the trigger of the ScrolledWindow.
Moving the canvas horizontally and vertically should be only possible with the mouse or trackpad.

I saw you're already taking care of the trackpad issue on #213, amazing!
For the mouse actions of the canvas, this is the behaviour we should implement:
Pan action with `mouse middle click` and drag or <kbd>Spacebar</kbd> + click and drag.

albfan:
@Alecaddd got it. There's a couple of big fix I want to implement with this:

- [ ] Connect accels to buttons and menus to avoid key_press workarounds
- [ ] Implement a singleton pattern to avoid passing window all around
- [ ] Use `using` to avoid full package names

Will look into avoid the shortcut for ScrolledWindow so it can reach the HeaderButtons `Top` and `Bottom`

Alecaddd:
After testing a bit the Canvas navigation options, I think it's better to change the keyboard shortcuts for the z-index order.
Intercepting and preventing the default `ScrolledWindow` <kbd>CTRL</kbd>+<kbd>Arrow</kbd> is super hacky and not great.

What about using these?
Move Up: <kbd>ALT</kbd>+<kbd>Arrow Up</kbd>
Move Down: <kbd>ALT</kbd>+<kbd>Arrow Down</kbd>
Move to the Top: <kbd>ALT</kbd>+<kbd>SHIFT</kbd>+<kbd>Arrow Up</kbd>
Move to the Bottom: <kbd>ALT</kbd>+<kbd>SHIFT</kbd>+<kbd>Arrow Up</kbd>

albfan:
Good, any common shortcut will work for me

albfan:
- The z-order is relative to existing objects (no matter how much you press up, if item is in top it is just one position above the next one) so there's no need to check touching objects.
- nobs and effects are correctly updated after each action.

Working on accels and refactors, but this is almost ready to go.

albfan:
@Alecaddd <kbd>Ctrl<kbd>+<kbd>Whatever<kbd> is working fine. It was my mistake to do by `key_press`. Relying on accels and hierarchy it just works.

We can do singleton and `using` refactor on next iteration

Issue #217, opened 2019-10-14
Canvas Issues


Alecaddd:
While I'm working on scrolling and dragging actions for the Canvas in order to enable standard movement and zooming like other apps, I noticed some issues with the canvas.

Here's everything I found out so far:
* Creating a shape not at 100% scale is visually broken. The shape is created in a separated location from the mouse click.
* ~~The rotation NOB doesn't keep the distance on zoom out.~~

*(The list might grow or shrink with time)*

Also the bounds of the canvas weren't updating properly and it wasn't possible to move items outside the edges of the initial size of the grid container. I fixed that so ignore it.

albfan:
I see movement is constrained to an I visible limit. Lot to fix in scale tbh.

Should be some fix here and there. I can deal with it.

Alecaddd:
I fixed the constraint issue, don't worry about that.
It's in my PR

Alecaddd:
Another thing we should consider is splitting the Canvas.vala file into multiple smaller files handling unique aspects of the canvas library.

The Canvas.vala file is becoming larger and larger, causing merging issues, and making it hard to debug it and maintain it.

albfan:
Totally agreed. Let's work on split it.

Alecaddd:
Closing as this was fixed by #268

Issue #206, opened 2019-09-29
Alpha Release Wishlist


Alecaddd:
## Setting up expectations for an Alpha release

Here's a list of must-have features for a first release of Akira.
This list is a **Work in Progress** and it can change based on our availability and technical challenges.

### General UI
- [x] All icons can be changed between regular and symbolic from Settings.
- [x] Hide/Show labels underneath Headerbar buttons from Settings.
- [x] Remember side panel sizing in `gsettings`.
- [x] Preview mode with <kbd>CTRL</kbd> + <kbd>.</kbd>.
- [x] Alert dialog before closing if document has been edited and not saved.

### Canvas
- [x] Default Canvas at 100K x 100K pixels.
- [x] Zoom action via Headerbar button and <kbd>CTRL</kbd> + <kbd>+</kbd>/<kbd>CTRL</kbd> + <kbd>-</kbd>
- [x] Zoom action via mouse scroll with <kbd>CTRL</kbd> + <kbd>Scroll</kbd>
- [x] Pan action with mouse middle click and <kbd>Spacebar</kbd> + click and drag.
- [x] Create Artboards.
- ~Shapes should snap to Artboard's edges.~
- ~Click and drag on an empty canvas area to select multiple shapes.~

### Shapes
- [x] Manage default fill and border color in Settings.
- [x] Add Rectangle.
- [x] Add Ellipse.
- [x] Import/Add Image.
- [x] Control shapes order via Headerbar buttons.
- [x] Control Position and sizing via Transform Panel.
- [x] Lock resize ratio via Transform Panel.
- [x] Lock resize ratio when holding <kbd>CTRL</kbd> on shape creation.

### File format
- [x] Save simple file with version control in `.akira` format (No UI for alpha release).
- [x] Export Canvas and Artboards as PNG.
- ~Export Canvas and Artboards as SVG.~

### Options Panel
- [x] Control fill color.
- [x] Control border size and color.
- [x] Control border radius for Rectangle shapes.
- ~Control alignment relative to the Canvas or Artboard.~

### Layers Panel
- [x] Lock/Hide canvas items from layer panel.
- [x] Edit names of layers and artboards.
- [x] Delete layers and artboards.
- ~Search by name.~
- [x] Control shapes order via Layers panel.

### Extras
- [x] Shortcuts helper panel with <kbd>F1</kbd> key.

mirekmazel:
[...]

PR #218, opened 2019-10-14
Canvas actions

Alecaddd:
This PR will implement the following Mouse and Trackpad actions to interact with the Canvas:

## Mouse actions:
- [x] Move Up: <kbd>Scroll Up</kbd>
- [x] Move Down: <kbd>Scroll Down</kbd>
- [x] Move Right: <kbd>SHIFT</kbd> + <kbd>Scroll Up</kbd>
- [x] Move Left: <kbd>SHIFT</kbd> + <kbd>Scroll Down</kbd>
---
- [x] Zoom in: <kbd>CTRL</kbd> + <kbd>Scroll Up</kbd>
- [x] Zoom out: <kbd>CTRL</kbd> + <kbd>Scroll down</kbd>
---
- [ ] Move Up/Down/Right/Left: Click and hold <kbd>Sroll Wheel</kbd> while dragging
---
- [ ] If an item is selected, zoom is anchored to it.
- [ ] Zoom 100% on selected object: <kbd>CTRL</kbd> + <kbd>1</kbd>
- [ ] Zoom fit on selected object: <kbd>CTRL</kbd> + <kbd>2</kbd>
- [ ] Zoom fit everything: <kbd>CTRL</kbd> + <kbd>3</kbd>

## Trackpad actions:
- [ ] Recognize and follow the standard two finger swipes.

## Fixes
Fixes canvas bounds on creation.

PR #219, opened 2019-10-14
WIP: Canvas Scroll, Pan & Other Item handling improvements

giacomoalbe:

With this PR I'd like to implement some Canvas actions and movements and to improve the usability of the Canvas tools (move, resize, rotate, etc) as stated in #212, #217 and #206.

Here's the list of desidered/already done implementations: 

- [x] Reordered methods in Canvas to be more consistent  (item add methods, event handling methods, ecc)
- [x] Pan Canvas with Space + mouse move
- [x] Drag item everywhere on the canvas
- [x] Add rotation step increments while holding CTRL

I know @Alecaddd is working on this too, but those particular actions were missing from his PR, so I  thought about implementing them.

If you think with PR is too broad, I can simply split in two or even more, or if you think there's should be more urgent work to do on other topics, please let me know and I'll be working on that instead.

### DIFF
--- a/src/Layouts/MainCanvas.vala
+++ b/src/Layouts/MainCanvas.vala
@@ -42,9 +42,6 @@ public class Akira.Layouts.MainCanvas : Gtk.Grid {
         canvas.set_bounds (0, 0, 10000, 10000);
         canvas.set_scale (1.0);
 
-        // Necessary in order to update bounds in fix_position
-        canvas.update_bounds ();
-
         canvas.canvas_moved.connect ((event_x, event_y) => {
             // Move scroll window according to normalized mouse delta
             // relative to the scroll window, so with Canvas' pixel
--- a/src/Layouts/MainCanvas.vala
+++ b/src/Layouts/MainCanvas.vala
@@ -42,6 +42,9 @@ public class Akira.Layouts.MainCanvas : Gtk.Grid {
         canvas.set_bounds (0, 0, 10000, 10000);
         canvas.set_scale (1.0);
 
+        // Necessary in order to update bounds in fix_position
+        canvas.update_bounds ();
+
         canvas.canvas_moved.connect ((event_x, event_y) => {
             // Move scroll window according to normalized mouse delta
             // relative to the scroll window, so with Canvas' pixel
--- a/src/Layouts/MainCanvas.vala
+++ b/src/Layouts/MainCanvas.vala
@@ -24,6 +24,9 @@ public class Akira.Layouts.MainCanvas : Gtk.Grid {
     public Gtk.Allocation main_window_size;
     public weak Akira.Window window { get; construct; }
 
+    private double scroll_origin_x = 0;
+    private double scroll_origin_y = 0;
+
     public MainCanvas (Akira.Window window) {
         Object (window: window, orientation: Gtk.Orientation.VERTICAL);
     }
@@ -34,9 +37,27 @@ public class Akira.Layouts.MainCanvas : Gtk.Grid {
         main_scroll.expand = true;
 
         canvas = new Akira.Lib.Canvas (window);
+
         canvas.set_size_request (main_window_size.width, main_window_size.height);
         canvas.set_bounds (0, 0, 10000, 10000);
         canvas.set_scale (1.0);
+
+        canvas.canvas_moved.connect((event_x, event_y) => {
+            // Move scroll window according to normalized mouse delta
+            // relative to the scroll window, so with Canvas' pixel
+            // coordinates translated into ScrolledWindow's one.
+            var delta_x = event_x - scroll_origin_x;
+            var delta_y = event_y - scroll_origin_y;
+
+            main_scroll.hadjustment.value -= delta_x;
+            main_scroll.vadjustment.value -= delta_y;
+        });
+
+        canvas.canvas_scroll_set_origin.connect((origin_x, origin_y) => {
+            // Update scroll origin on Canvas' button_press_event
+            scroll_origin_x = origin_x;
+            scroll_origin_y = origin_y;
+        });
 
         main_scroll.add (canvas);
 
--- a/src/Layouts/MainCanvas.vala
+++ b/src/Layouts/MainCanvas.vala
@@ -42,7 +42,7 @@ public class Akira.Layouts.MainCanvas : Gtk.Grid {
         canvas.set_bounds (0, 0, 10000, 10000);
         canvas.set_scale (1.0);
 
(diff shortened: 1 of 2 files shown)


albfan:
I discourage to mix whitespace corrections and method reorder when adding functionality. There's lot of movement this days, and that just make merges hard.

You can always fix those things but please do in separate commits, so we can see what you're really adding or fixing

giacomoalbe:
Hi @albfan, thanks for the response! I've seeen that there's a lot of movement these days, I'm very happy about it. 

I can get your point, but I think that my little reorder is simple enough not to break things. I already merged it in my repo from current master and it has not been that hard to fix the conflicts. 

Nevertheless, I'll be marking each edit accordingly and I'll put those changes in different commits.

albfan:
Yes is just a general advice. surely we can deal with this changes. Thanks for working on this!

Alecaddd:
Thanks for working on this, I'll review it later and see how it behaves.
As you noticed, I'm working on something similar in canvas actions, more focused on the canvas rather than the objects, so I think some merging issues are likely to happen.

giacomoalbe:
Hi @Alecaddd, thanks for your answer, as well. 

I implemented the space pan behaviour and the constrained rotation with CTRL, but now I think it's wise enough to step there and maybe open another PR for the other things (like centered resize and aspect ratio contrained resize), in order to avoid unneeded conflicts during merge. 

I expect the zoom with ctrl + wheel and pan with middle mouse to be implemented in your PR, so I just don't want to make any "double" in here. Am I right? 

I would like to work on something else for the 1.0 Milestone, maybe the Artwork/multiple select topic might be more suitable in order to minimize conflicts?

src/Layouts/MainCanvas.vala
```diff
@@ -34,10 +37,31 @@ public class Akira.Layouts.MainCanvas : Gtk.Grid {
         main_scroll.expand = true;
 
         canvas = new Akira.Lib.Canvas (window);
+
         canvas.set_size_request (main_window_size.width, main_window_size.height);
         canvas.set_bounds (0, 0, 10000, 10000);
         canvas.set_scale (1.0);
 
+        // Necessary in order to update bounds in fix_position
+        canvas.update_bounds ();
```
Alecaddd: Please remove this as I've implemented it in PR #218

src/Lib/Canvas.vala
```diff
@@ -124,9 +132,98 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         events |= Gdk.EventMask.BUTTON_PRESS_MASK;
         events |= Gdk.EventMask.BUTTON_RELEASE_MASK;
         events |= Gdk.EventMask.POINTER_MOTION_MASK;
+
+        update_bounds ();
```
Alecaddd: Remove this as well.
Also, there's no need to call this method on `construct` if you're calling it in the MainCanvas file as you're basically tapping the same method twice.

src/Lib/Canvas.vala
```diff
@@ -124,9 +132,98 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
+        text.set_transform (Cairo.Matrix.identity ());
+        var artboard = window.main_window.right_sidebar.layers_panel.artboard;
+        var layer = new Akira.Layouts.Partials.Layer (window, artboard, text, "Text", "shape-text-symbolic", false);
+        text.set_data<Akira.Layouts.Partials.Layer?> ("layer", layer);
+        artboard.container.add (layer);
+        artboard.show_all ();
+        return text;
+    }
+
+    public void update_bounds () {
```
Alecaddd: remove this method

src/Lib/Canvas.vala
```diff
@@ -98,6 +102,10 @@ public class Akira.Lib.Canvas : Goo.Canvas {
 
     private Goo.CanvasRect? hover_effect;
 
+    private bool ctrl_is_pressed = false;
+    private bool pan_mode_enabled = false;
+    private bool zoom_mode_enabled = false;
+    private bool pan_zoom_is_holding = false;
```
Alecaddd: Are these two variables and related methods used anywhere?
It's seems this is a leftover for the zoom handling, which I've already implemented in my WIP PR.
Please, remove variables and methods that are not used and are not part of the scope of this PR.

src/Lib/Canvas.vala
```diff
@@ -50,12 +53,13 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
-      if (_edit_mode == EditMode.MODE_SELECTION) {
-        set_cursor (Gdk.CursorType.ARROW);
-      } else {
-        set_cursor (Gdk.CursorType.CROSSHAIR);
-      }
+        if (_edit_mode == EditMode.MODE_SELECTION) {
+            set_cursor (Gdk.CursorType.ARROW);
+        } else {
+            set_cursor (Gdk.CursorType.CROSSHAIR);
+        }
```
Alecaddd: We should visually show when the user enables the drag action to pan.
Let's change the cursor to the `grab` when holding spacebar, and the `grabbing` when clicking:
https://developer.gnome.org/gdk3/stable/gdk3-Cursors.html

src/Lib/Canvas.vala
giacomoalbe: Totally agree with you. I removed those variables and I took care of removing everything not related to the pan and click fixed rotation additions.

src/Lib/Canvas.vala
giacomoalbe: While reading the doc you provided, I discovered that the suggested way of changing cursor is by using `Gdk.Cursor.from_name (string)` rather than the `Gdk.Cursor.for_display (Gdk.CursorType)`. The difference is in the way in which we choose the cursor: in the former, you need to type the cursor name, in the latter the CursorType, which is an enum of not-so-well-supported X cursor types. Since `grab` and `grabbing` were not present in the `Gdk.CursorType` enum too, I refactored the code to make use of the `from_name` construct. Everything works the same as before, the only difference is the rotation cursor, I wasn't able to find something similar to the previous one. Maybe we can use a custom cursor.

aberba:
What is happening?

Alecaddd:
Hey @giacomoalbe, sorry for the delay in reviewing this.
I pushed a rebased version from master on this branch: https://github.com/akiraux/Akira/tree/giacomoalbe-canvas-scroll-and-pan

Upon testing it the scrolling and panning don't work as expected.
Would you be able to take a look at it and rebase your PR?
Thank you so much.

Alecaddd:
I guess this can be closed, right @giacomoalbe?

PR #225, opened 2019-12-26
Fix segfault crashes

Alecaddd:
I started working on some needed improvements to fix those random segfault.
I found out that some methods were getting called 2 or 3 times per click from multiple sources, and some areas were duplicated, probably due to some leftover code.

I improved the code a bit, being sure methods are called only once, preventing most of those infinite loops causing a segfault.

Still, some crashes are occurring, but they're not always related to the Fill Model, as I've experienced crashes also when commenting that entire area, by simply quickly selecting the same shape, or quickly clicking between multiple shapes.

I will still work on this, but if you guys can load this PR and see if you can figure why the last segfault are happening, it'd be great.

Akira can't crash :smile:

### DIFF
--- a/src/Layouts/LeftSideBar.vala
+++ b/src/Layouts/LeftSideBar.vala
@@ -49,10 +49,17 @@ public class Akira.Layouts.LeftSideBar : Gtk.Grid {
         var style_panel = new Akira.Layouts.Partials.StylePanel ();
         fill_box_panel = new Akira.Layouts.Partials.FillsBoxPanel (window);
 
+        var scrolled_window = new Gtk.ScrolledWindow (null, null);
+        scrolled_window.expand = true;
+        var scrolled_grid = new Gtk.Grid ();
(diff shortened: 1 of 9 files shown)


Alecaddd:
I think I fixed all the other crashes happening except for those related to the fill model.
Random segfaults were happening when setting the cursor on startup, and when quickly clicking on the empty canvas.

Onto figuring out why the fill model causes a crash when clearing the list of items.

Alecaddd:
Done! This is fixed....I think.
Ready for a full review!

Alecaddd:
Thanks for the review.
I commented the debug because I needed to print messages only for the sections I was debugger, and all the other stuff were making a lot of noise.
I prefer to keep it commented so we can enable only the messages we need instead of having a terminal filled with everything.
What do you think?

albfan:
As they are debug, they should only be printed if `G_MESSAGES_DEBUG=all` is set. Let me check what's happening

PR #226, opened 2019-12-28
Initial work on Canvas split: Nobs

Alecaddd:
This is a Draft PR to kick-off the discussion on how to split the canvas in dedicated classes to simplify our life.

The idea here is to create a dedicated class to handle the `Nobs` and delegate the update of their position internally, based on the type.

It's possible to extend the `Goo.CanvasRect` pretty easily, and I was able to create the Nob with a simple line:
```
nobs[i] = new Selection.Nob (get_root_item (), current_scale, i);
```

Unfortunately, I'm not able to call the newly created `update_position()` method with this:
```
nobs[i].update_position (target, select_effect);
```
as I'm getting:
```
error: The name `update_position' does not exist in the context of `Goo.CanvasItemSimple'
```

Do you guys know if it's possible to implement a new method here?
@albfan @giacomoalbe

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -105,7 +105,6 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     private double delta_y;
     private double hover_x;
     private double hover_y;
-    private double nob_size;
     private double current_scale;
     private int holding_id = Nob.NONE;
     private double bounds_x;
@@ -548,11 +547,8 @@ public class Akira.Lib.Canvas : Goo.Canvas {
 
         select_effect.set ("parent", get_root_item ());
 
-        nob_size = 10 / current_scale;
-
         for (int i = 0; i < 9; i++) {
             nobs[i] = new Selection.Nob (get_root_item (), current_scale, i);
-            //  nobs[i].update_position (target, select_effect);
         }
 
         update_nob_position (target);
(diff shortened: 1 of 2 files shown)


albfan:
@Alecaddd the problem with `update_position()` should be fixed. 

I think we just need a `NobManager` separate class with know about canvas and put the nobs around an item. Any nob should know the item it refer to so its movement affect the item (resize, rotate)

We can allow the NobManager to deal with list of items so we can cover multiple selection.

Alecaddd:
@albfan @giacomoalbe 

I moved the position update and cursor setting to the `Nob` class.
What do you guys think about this approach?
I'm thinking to move also the part that deals with Nobs in the `motion_notify_event()` method inside the `Nob` class, in order to remove the `enum Nob` from the canvas, and have all the Nob interaction containerized in that class.

Does it make sense having both the methods to generate the Nob and manage the positions inside the same class? How do you envision the `NobManager` approach?

giacomoalbe:
> @albfan @giacomoalbe
> 
> I moved the position update and cursor setting to the `Nob` class.
> What do you guys think about this approach?
> I'm thinking to move also the part that deals with Nobs in the `motion_notify_event()` method inside the `Nob` class, in order to remove the `enum Nob` from the canvas, and have all the Nob interaction containerized in that class.
> 
> Does it make sense having both the methods to generate the Nob and manage the positions inside the same class? How do you envision the `NobManager` approach?

@Alecaddd take a look at the comment I wrote on #229, the approach is the same as what you've described and prototyped in here, I'm going to make a tiny MVP so show off some of the advantages, I think that aftterwards we can merge our work altogether.

Alecaddd:
Closing this because of the work in progress on #231

PR #227, opened 2019-12-28
Fixed Increments Rotation

giacomoalbe:
Before this PR, rotation wasn't bound to any increments and could have been any real number. With this PR, you can easily set often used angles (15, 30, 45, 90, 180) by pressing CTRL and rotation the shape using the nob.

The default increment is 15deg.

* The fixed increment is _fixed_, so it is not possible to change it. Do we need to create a setting entry for that or is it ok to stick to a fixed one?

## This PR implements the following **features**:

* Fixed rotation while holding CTRL and rotation the shape with the mouse
* Fixed rotation following pointer 
* Capped rotation range to [0, 360[
* Rounded rotation amount to decimal values to prevent subdegree issues

### DIFF
--- a/src/Utils/Math.vala
+++ b/src/Utils/Math.vala
@@ -1,0 +1,32 @@
+/*
+* Copyright (c) 2019 Alecaddd (http://alecaddd.com)
+*
+* This file is part of Akira.
+*
+* Akira is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+
+* Akira is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+
+* You should have received a copy of the GNU General Public License
+* along with Akira.  If not, see <https://www.gnu.org/licenses/>.
+*
+* Authored by: Giacomo "giacomoalbe" Alberini <giacomoalbe@gmail.com>
+*/
+
+public class Akira.Utils.Math {
+    public static double roundDouble(double val, int precision = 2) {
+        int precisionPower = 1;
+
+        for (int i=0; i<precision; i++) {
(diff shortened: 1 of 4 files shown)


albfan:
@giacomoalbe this almost work. But the nob and the cursor should be keeped in correlation. Seems cursor is taking advantage for the nob position each time.

fixed-rotation

Alecaddd:
This works great for me.
@albfan you can do another review and see if the things you requested has been properly addressed.
If so, feel free to squash and merge.

giacomoalbe:
> @giacomoalbe this almost work. But the nob and the cursor should be keeped in correlation. Seems cursor is taking advantage for the nob position each time.
> 
> fixed-rotation

Thanks for the suggestions @albfan, I wouldn't have had the opportunity to spot them, since, as you said, "4 eyes are better than 2" and in this particular case my 2 eyes were already biased and unable to find the glitch, so thanks!

I corrected the issues, what do you think? It seems to me that the nob now follows tightly the cursor, isn't it?

albfan:
@giacomoalbe for my is almost done. There's a conflict with current master, can you rebase on top of that to check the actual changes?

giacomoalbe:
> @giacomoalbe for my is almost done. There's a conflict with current master, can you rebase on top of that to check the actual changes?

Yes I'll take care of merging this with the current master. I'll try to rebase with the current master, in case I couldn't get it to work, I will simply merge master (with --strategy-theirs) and take the logic out of the commit history.

src/Utils/Math.vala
```diff
@@ -0,0 +1,32 @@
...
+* along with Akira.  If not, see <https://www.gnu.org/licenses/>.
+*
+* Authored by: Giacomo "giacomoalbe" Alberini <giacomoalbe@gmail.com>
+*/
+
+public class Akira.Utils.Math {
+    public static double round_double (double val, int precision = 2) {
+        int precision_power = 1;
+
+        for (int i = 0; i < precision; i++) {
```
donadigo: Instead of a loop I would just use `int precision_power = GLib.Math.pow10 (precision);` here.

Alecaddd:
Closing this in favour of #260

PR #228, opened 2019-12-28
Pan action with mouse middle click and Spacebar + click and drag.

giacomoalbe:
- [**DONE**]Right now MainCanvas is 10k x 10k pixels but the Canvas only get noticed about 1k x 1k.
- [**DONE**]Canvas starting position is top-left, maybe centering it would be a better idea in order to enable panning in every direction since the beginning (right now right panning is not possible since the canvas is already at its leftmost edge)
- Canvas scrollbars have been removed, since it would have difficult to grab object near the edges. Now what has become difficult is the ability to understand where we are on the canvas, so we should give a way to the user to understand precisely where he/she is and when the canvas "ends". Maybe a minimap would help?
- We should provide (maybe for the next release, not the 1.0) a way to reset canvas to "center" and to zoom in to certain objects (maybe by selecting them in the right panel). Maybe also "locking" the view to a specific zoom would help, too.

## This PR implements the following **features**:
- [x] Canvas pan action with Spacebar + click
- [x] Canvas pan with middle mouse click and drag
- [x] Center canvas at startup
- [x] Make canvas' items move action aware of canvas entire size
- [x] Zoom with scroll + ctrl
- [x] Vertical scroll with scroll
- [x] Horizontal scroll with shift + scroll

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -89,7 +89,6 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         MODE_SELECTION,
         MODE_INSERT,
         MODE_PAN,
-        MODE_ZOOM,
     }
 
     public enum InsertType {
@@ -173,7 +172,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         //  debug ("canvas temp event x: %f", temp_event_x);
         //  debug ("canvas temp event y: %f", temp_event_y);
 
-        if (edit_mode == EditMode.MODE_PAN || edit_mode == EditMode.MODE_ZOOM) {
+        if (edit_mode == EditMode.MODE_PAN) {
             double tmp_event_x_normalized = temp_event_x;
             double tmp_event_y_normalized = temp_event_y;
 
@@ -540,15 +539,10 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                 delete_selected ();
                 return true;
             case Gdk.Key.space:
-                if (edit_mode != EditMode.MODE_ZOOM) {
-                    edit_mode = EditMode.MODE_PAN;
-                }
+                edit_mode = EditMode.MODE_PAN;
                 return true;
             case Gdk.Key.Control_L:
             case Gdk.Key.Control_R:
-                if (edit_mode != EditMode.MODE_PAN) {
-                    edit_mode = EditMode.MODE_ZOOM;
-                }
                 return true;
         }
 
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -63,6 +63,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     }
     public InsertType? insert_type { get; set; }
 
+
     /*
         Grabber Pos:   8
                      0 1 2
@@ -86,7 +87,9 @@ public class Akira.Lib.Canvas : Goo.Canvas {
 
     public enum EditMode {
         MODE_SELECTION,
-        MODE_INSERT
+        MODE_INSERT,
+        MODE_PAN,
+        MODE_ZOOM,
     }
 
     public enum InsertType {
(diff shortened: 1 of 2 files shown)


Alecaddd:
> Right now MainCanvas is 10k x 10k pixels but the Canvas only get noticed about 1k x 1k.

This happens because the bounds need to be updated after the `construct` method run its course.
I fixed this issue in the `canvas-actions` branch: https://github.com/akiraux/Akira/pull/218/files#diff-cb8c15a0d16ffaff0fa8bb3c83a934bcR45

You can add that portion of the code to your PR as my `canvas-actions` PR needs a lot of rebasing and changes, so it's stale for now.

giacomoalbe:
Ok @Alecaddd, I'm going to take the code you mentioned and add to my PR!

Alecaddd:
> Canvas scrollbars have been removed, since it would have difficult to grab object near the edges. Now what has become difficult is the ability to understand where we are on the canvas, so we should give a way to the user to understand precisely where he/she is and when the canvas "ends". Maybe a minimap would help?

I'm not a fan of minimaps as they're not always used unless for specific edge cases like you mentioned. And most of the times are just hidden widget that we're forced to support and remember to update.

Other design apps like Sketch have these shortcuts to help the user recenter and navigate the "infinite" canvas:
- If an item is selected, zoom actions are anchored to it as center point.
- Zoom at 100% on selected object: <kbd>CTRL</kbd> + <kbd>1</kbd>
- Zoom to fit selected object on screen: <kbd>CTRL</kbd> + <kbd>2</kbd>
- Zoom to fit everything on screen: <kbd>CTRL</kbd> + <kbd>3</kbd>

We could provide these clickable menu items maybe in a context menu from the zoom widget, but these are pretty standard shortcuts for designers, so I wouldn't worry too much about discoverability.

Is this PR ready for a full review or are you still working on it?
I'd recommend to not add too man things in a single PR and keeping them small and focused.

giacomoalbe:
I agree with you on the minimap. 

It is indeed better to have the shortcuts you mentioned working, but I think it's wiser to leave those for another PR, maybe after the Canvas split has been done. 

The PR is ready for a full review, I just don't want to add any more functionality, I was just looking at things of your #218 PR that could have been implemented without much effort, and so I added the last 2 commits. 

So, feel free to start a complete review!

Alecaddd:
Indeed, those zoom actions can be implemented in a follow up PR.
Maybe we could do it in the `canvas-action` PR, which I have to heavily rebase once this lands.
I'll start a review

src/Lib/Canvas.vala
```diff
@@ -673,31 +771,24 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
-                set_cursor (Gdk.CursorType.TOP_SIDE);
-                break;
-            case Nob.TOP_RIGHT:
-                set_cursor (Gdk.CursorType.TOP_RIGHT_CORNER);
-                break;
-            case Nob.RIGHT_CENTER:
-                set_cursor (Gdk.CursorType.RIGHT_SIDE);
-                break;
             case Nob.BOTTOM_RIGHT:
-                set_cursor (Gdk.CursorType.BOTTOM_RIGHT_CORNER);
```
Alecaddd: Any particular reason why you switched to name based cursors?
I have nothing against it, I'm just curious.

giacomoalbe:
I need to test the code with a touchpad/trackpad since I'm using only mouse and keyboard atm. 

Thanks for the suggestions, btw, I'll check later today!

Alecaddd:
>  Two-finger horizontal scroll doesn't work anymore. Normally I'm able to scroll both horizontally and vertically with it.

I fixed it locally and based the scrolling on the delta, in this way the acceleartion is more natural.

> Spacebar + trackpad click + trackpad swipe (both single or two-finger) doesn't move the canvas.

Ignore this, it was a problem of my setup where I had set to disable the trackpad while typing.

I'm merging this as it works great!

PR #231, opened 2019-12-30
[WIP] Canvas split

giacomoalbe:
## Summary

This PR tries to divide the huge amount of logic and code in the Canvas.vala file. 

Goal of this PR is to identify and separate the different pieces that compose the process of interacting with the canvas. 

In particular, this PR introduces a new kind of objects (called _Managers_) that have complete control over a certain and very well defined set of actions.

The communication between them is handled both by the `Canvas` module itself and by the `EventBus` singleton, in order to diminish (or to completely remove) the coupling between the different pieces of the UI and to have a _single point of control_ of the communication inside the application.

The new `Canvas` module is responsible for *state management* and *event handling*, leaving all the other tasks to more specialized agents, responsbile (whenever possible) for just a couple of different tasks, all related to a particular set of actions.

The canvas acts as a dispatcher of events based on the state, which is, in turn, is modified by the user interaction with the UI (mouse move, key press, UI action triggers, etc). 

As a general design pattern, it has been used extensively the EventBus module to communicate between managers, canvas and other UI elements. This way, it is no more necessary to keep references of `Akira.Window` in each class needing to communicate with the "internal world", but instead they can use the `event_bus` singleton (which is now linked to `Akira.Application`, even though connecting it to `Akira.Window` would be more correct) to communicate changes or send information. 
Is then duty of the listener to act accordingly and update itself it needed. 
  
## Managers

Here will follow a thorough description of every new agent introduced by this PR. 

### SelectedBoundManager
This manager is responsible to hold every info related to the items in the current selection. 
It also updates the selection according to the user input (adds or removes items from selection).
It is in charge of transferring the mouse click & move from the canvas to the items in selection. 

Here a utility class called `AffineTransform` will perform geometric actions on the currently selected object. 

The managers itself holds all the information needed to compute and permorm those geometric operations, such as the `intial_event` coordinates and size. 

Whenever a change occurs in the selection (either in terms of *position* and *size* or in the items count) the Manager emits a signal on the `EventBus` to notify the other managers to change accordingly.

Currently, only 1 item selection is provided, but adding multi select either by _click & dragging_ on the canvas or by _shift selecting_ over a collection of items should be straightforward, given that the most challenging task would be to accomplish the requried geometric transformations on the selected elements. 

### NobManager
 This manager is responsible for drawing the `select_effect` and `nob_handles` around the selected item's bound.
  
  When there's only one item, the `select_effect` will follow both the bounds of the selected object and the trasformation applied to it. 

  In the *multi select* case, in turn, the `select_effect` will be drawn according to the effective selection bound, which is the union of each bound of the items in selection.
  Every user transform to the multi select bound will result in a transformation of the bound itself and a corresponding transformation of each selected item according to the center of trasformation.    

### ItemsManager
 This manager holds the canvas' items list and is responsible for each action that modifies this list (add, remove, update). 

 Here the mapping between user set keyboard shortcut related to item addition and the current pressed key is done in order to get the user intention. Currently, those mapping are hard coded and not user selectable, but this can be done easily in a future PR. 

  With regards to *canvas items* a new abstract class' interface has been created. 
  The `CanvasItem` base class (which is **NOT** a class, in order to enable class inheritance from `GooCanvas` basic geometric building blocks - `Goo.CanvasRect`, `Goo.CanvasEllipse` and so on - ) dictate the common interface that each `CanvasItem` should have, with properties and methods that are common between items. 

 Custom properties, methods and behaviour can be added here to increase the set of features our items should have. 

### HoverManager
 This manager is only resposible for each operation related to hovering. 

 It distinguish between item and hovering and acts accordingly (it adds items to the canvas or request the change of the cursor, action that then takes place in the Canvas module). 

## Development Status
Those are the Managers that are planned to be implemented:

- [x] SelectedItemsBoundManager
- [x] NobManager
- [x] ItemsManager
- [x] HoverManager
- [x] Properties Panel link with selected item (two way)
- [x] Canvas zoom 
- [x] Fill panel link with current selected item 
- Raise & lower item [moved to future PR]
- Layer panel update [moved to future PR]
- Folder & names refactor [moved to future PR]

## Known issues

* I'm not totally convinced by the namespace structure I put in place. We have different types of object (like `Layouts`, `Services`, `Utils`, `FileFormat`, `Models`, `Partials`, etc) but I don't think we should keep the Canvas inside the `Lib` folder. Maybe a `Canvas` folder with inside the `AkiraCanvas.vala` main file and `Managers`, `Models` and `Items` subfolders might be a better choice. But maybe with can keep things like that and leave those modification to a more specific PR. 
* Item scaling is not yet perfect, some glitches occur when trying to shrink item to right or bottom wise and the overall mechanism doesn't account for a precise, custom trasform center aware scaling
* EventBus event dispatch should be improved, in order to send different types of arguments to listeners. Calling `event_bus.emit ` function is a good way to have a single point of control for both logging and history management, but right now for each new signal one should need to update a very long `switch-case` block with custom signal calls. Finding a more programmatic way to infer the signal name from the signal's string representation  and to pass custom arguments should be the focus of a next, very needed, PR. 
* There is a lot of coupling between many different modules. This needs to be removed by using signal communication properly and not reference method calls, since this introduces a lot of boilerplate code, the need of long references chains and logic duplication, which is a "refactor" nightmare.


Alecaddd:
> use the `event_bus` singleton (which is now linked to Akira.Application, even though connecting it to Akira.Window would be more correct)

Indeed, technically it would be more correct, but then we'd need to pass with `Window` instance to all the classes that need to connect to events, right?

Alecaddd:
I'll review this later today, as thoroughly as I can.
Did you see my previous comment? @giacomoalbe

src/Layouts/Partials/TransformPanel.vala
```diff
@@ -205,17 +218,54 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
...
+    }
+
+    private void on_selected_items_changed (List<Lib.Models.CanvasItem> selected_items) {
+        if (selected_items.length () == 1) {
+            item = selected_items.nth_data (0);
+        } else {
+            if (item != null) {
+                item = null;
+            }
+        }
```
Alecaddd: This looks pretty dirty. Let's try to always avoid nested conditions and handle the negatives first.
eg:

```
if (!selected_items.length ()) {
  item = null;
  return;
}

item = selected_items.nth_data (0);
```

src/Layouts/Partials/TransformPanel.vala
```diff
@@ -205,17 +218,54 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
...
+        size_ratio = 1.0;
+        size_lock = false;
+
+        x.notify["value"].connect (x_notify_value);
+        y.notify["value"].connect (y_notify_value);
+        width.notify["value"].connect (width_notify_value);
+        height.notify["value"].connect (height_notify_value);
+        rotation.notify["value"].connect (rotation_notify_value);
+        opacity_adj.notify["value"].connect (opacity_notify_value);
+    }
```
Alecaddd: Is this disconnection and reconnection necessary?
if so, please write a comment to explain why. Maybe we can find a cleaner solution here.

src/Lib/Managers/HoverManager.vala
```diff
@@ -0,0 +1,144 @@
...
+    private Goo.CanvasItem hover_effect;
+
+    public HoverManager (Goo.Canvas canvas) {
+        Object (
+            canvas: canvas
+        );
+    }
+
+    construct {
+    }
```
Alecaddd: Do we need an empty constructor here?

src/Lib/Managers/HoverManager.vala
```diff
@@ -0,0 +1,144 @@
...
+            height += line_width * 2;
+
+            if (!item.selected) {
+                hover_effect = new Goo.CanvasRect (
+                    null,
+                    x, y,
+                    width, height,
+                    "line-width", 2.0,
+                    "stroke-color", "#41c9fd",
+                    null
```
Alecaddd: Make the line-width and stroke-color values constants at the beginning of the class, so they're easier to find and update.

src/Lib/Managers/ItemsManager.vala
```diff
@@ -0,0 +1,205 @@
...
+ * along with Akira.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Authored by: Giacomo Alberini <giacomoalbe@gmail.com>
+ */
+
+public enum InsertType {
+    RECT,
+    ELLIPSE,
+    TEXT
+}
```
Alecaddd: Shouldn't be this inside the class?

src/Lib/Canvas.vala
```diff
@@ -140,814 +87,187 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
-                        new_delta_y = delta_y;
-                    } else {
-                        break;
+            default:
+                if (uppercase_keyval <= Gdk.Key.Z && uppercase_keyval >= Gdk.Key.A) {
+                    // Send to ItemsManager to deal with custom user shape
+                    // hotkey preferences from settings
+                    if (items_manager.set_insert_type_from_key (uppercase_keyval)) {
+                        edit_mode = Akira.Lib.Canvas.EditMode.MODE_INSERT;
                     }
```
Alecaddd: I don't understand this part, can you explain it?
Also, let's avoid indented conditions.

src/Lib/Managers/ItemsManager.vala
```diff
@@ -0,0 +1,205 @@
...
+            case "ellipse":
+                insert_type = InsertType.ELLIPSE;
+                break;
+
+            case "text":
+                insert_type = InsertType.TEXT;
+                break;
+        }
+    }
+
```
Alecaddd: These entire file looks a bit convoluted.
Why do we need the `set_insert_type_from_key`, the `insert_item`, and the `set_item_to_insert` methods repeating the switch case condition?
Can we improve this to avoid code repetition?
With this approach, whenever we have to add another shape we need to remember to update multiple methods for the same thing.

src/Lib/Managers/NobManager.vala
```diff
@@ -0,0 +1,345 @@
...
+        event_bus.zoom.connect (on_zoom);
+    }
+
+    public void set_selected_by_name (Nob selected_nob) {
+        this.selected_nob = selected_nob;
+    }
+
+    public Nob get_selected_nob () {
+        return selected_nob;
+    }
```
Alecaddd: Don't use this approach but rather use the getter and setter methods when defining the variable in the class, like we were doing for the selected_item in the Canvas file.

src/Lib/Managers/NobManager.vala
```diff
@@ -0,0 +1,345 @@
...
+        height += line_width * 2;
+
+        select_effect = new Goo.CanvasRect (
+            null,
+            x, y,
+            width,
+            height,
+            "line-width", 1.0 / current_scale,
+            "stroke-color", "#666",
+            null
```
Alecaddd: Set the line width and stroke color variables as constants at the beginning of the class.

src/Lib/Managers/NobManager.vala
```diff
@@ -0,0 +1,345 @@
...
+        set_bound_coordinates (
+            selected_items,
+            ref x, ref y,
+            ref transform,
+            ref line_width,
+            ref width, ref height
+        );
+
+        for (int i = 0; i < 9; i++) {
+            nobs[i] = new Selection.Nob (
```
Alecaddd: If the line is shorter than 120ch, you can keep everything in one line so it's easier to read.

src/Lib/Managers/NobManager.vala
```diff
@@ -0,0 +1,345 @@
...
+        ref double _height
+    ) {
+        if (selected_items.length () == 1) {
+            var item = selected_items.nth_data (0);
+
+            item.get_transform (out transform);
+            item.get ("line_width", out line_width);
+            item.get ("width", out _width);
+            item.get ("height", out _height);
+        } else {
```
Alecaddd: use a `return;` here instead of the `else` statement.

src/Lib/Managers/NobManager.vala
```diff
@@ -0,0 +1,345 @@
...
+          nobs[Nob.LEFT_CENTER].set_transform (transform);
+          if (print_middle_width_nobs) {
+            nobs[Nob.LEFT_CENTER].translate (x - (nob_offset + line_width), y + (height / 2) - nob_offset);
+          } else {
+            nobs[Nob.LEFT_CENTER].translate (x - (nob_size + line_width), y + (height / 2) - nob_offset);
+          }
+          nobs[Nob.LEFT_CENTER].set ("visibility", Goo.CanvasItemVisibility.VISIBLE);
+        } else {
+          nobs[Nob.LEFT_CENTER].set ("visibility", Goo.CanvasItemVisibility.HIDDEN);
+        }
```
Alecaddd: Let's create a method to set the visibility so we don't have to repeat this and it's easier to update if things change.

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -0,0 +1,166 @@
...
+        }
+        set {
+            _selected_items = value;
+
+            update_selected_items ();
+
+        }
+    }
+
+    private unowned List<Models.CanvasItem> _selected_items;
```
Alecaddd: this should be above the `selected_items` declaration for better reading.

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -0,0 +1,166 @@
...
+            var selected_item = selected_items.nth_data (0);
+
+            initial_event_x = event_x;
+            initial_event_y = event_y;
+
+            canvas.convert_to_item_space (selected_item, ref initial_event_x, ref initial_event_y);
+
+            initial_width = selected_item.get_coords ("width");
+            initial_height = selected_item.get_coords ("height");
+        } else {
```
Alecaddd: `return;` here instead of `else`

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    private void update_selected_items () {
+        event_bus.selected_items_changed (selected_items);
+    }
+
+    private void on_request_selection_bound_transform (string property, double amount) {
+        Models.CanvasItem selected_item;
+        selected_item = selected_items.nth_data (0);
+
```
Alecaddd: Can you inline the `type` declaration?

giacomoalbe:
@Alecaddd thanks for the meaningful insights. 

> Lots of things don't work, like the fill panel, drag or spacebar to move the canvas, and the other listed.

Yes, I know, but adding them would have enlarged this PR even more, so I stopped where I found that the canvas-split was done propertly. 

I think that "splitting" this in different PR won't work, since there is a lot of "positive coupling" between all those new "managers", and adding one per time would require a lot of "workaround" code to be added (just adding one feature  would mean to use the canvas module to do the rest, and it's not easy to identify which part does what in the current Canvas module, so it's a diffucult and very time consuming task).

I know for sure that this change is "disruptive", but once in place it would be very easy to tackle all the issues/requests you listed, since the place in which to modifiy the logic is in its own module, and clashes between PR will be less likely to occur (we can work a lot more in parallel). 

So splitting in different PR for me is not a feasible option, what we could do is to document properly the whole process (a lot of it has been done in this PR itself) and to list the features that need to be implemented (which are already present in other active or just merged PR, so it's easy to get them back, in my opinion). 

I can take care of the missing things in future PR, I just didn't want to overload the already present changes with further logic that can be added atomically in narrow scoped PRs.

What we need to "review" in here is the logic behind the split, since it's not the only one possible and might be subjected to change in the future (in order to accomodate different needs), so we have to be sure that, right now, this design is a good candidate for handling the growing complexity of the app in the future.

So once the logic is "approved", I think that adding the "missing parts" would be an easy task.
Furthermore, since many of the features that were working were based on a very tightly coupled design approach, it is actually better to "upgrade" them to be more "event_bus" based.

Obviousely, I firmly want to find the solution that work best for all of us, so I'm completely open to your suggestions, feedback, insights and proposals.

src/Layouts/Partials/TransformPanel.vala
giacomoalbe: Ok, totally feasible!

src/Layouts/Partials/TransformPanel.vala
giacomoalbe: This code was already there, I just copied the logic present in order not to create infinite loops of setters and getters (I suppose). 

I think that there should (and can) be a better way to do this, maybe right inside the LinkedInput widget, with different methods like `reset`, `update`, `bind`, etc, in order to avoid to have to do it in every LinkedInputEntry. Maybe could we leave for a future and more specific PR on LinkedInput? I was facing other issues with them, so maybe we can take care of 'em plus this one alltogheter. 

I'll add a comment to summarize this Github comment.

src/Lib/Managers/HoverManager.vala
giacomoalbe: Nope, just a leftover! Housecleaning necessary :)

src/Lib/Managers/HoverManager.vala
giacomoalbe: Nope, just a leftover!

src/Lib/Managers/ItemsManager.vala
giacomoalbe: Yes, initially I wanted to make them as a separate "model", but in the end they were used just in there, so we can definitely keep them there!

src/Lib/Canvas.vala
giacomoalbe: Sure! Since this event handler is for "every" key on the keyboard, we should handle the case in which we don't have a key associated with a shape. So we need to change to insert mode only when the key pressed is alpha and is associated to a particular shape. This is done inside the ItemsManager, since in there we could load the user settings and do the mapping between user key and shapes. 

The function returns true only if there's a shape associated with the key being pressed.

I will put the first condition inside the method call, in order to avoid nested condition.

src/Lib/Managers/ItemsManager.vala
giacomoalbe: The code repetition can be removed, the reason there are 3 different switch is because it should be taken into account both addition from keyboard AND from filemenu. 

Of course things can be improved, maybe creating a single list-model with all the shapes a user can create, with both the key associated with and the action triggered from the file menu. 

The beauty of this approach is that now `ItemsManager` is the only place in which Item addition happens, so we can refactor it without changing any other file or module (given that we respect the interface we created). 

So I think we can leave this for the future, just track it down in order not to forget it. 

What do you think?

src/Lib/Managers/NobManager.vala
giacomoalbe: Actually, there's no need to have a public property baked by a private one, a simple public variable will work here!

src/Lib/Managers/SelectedBoundManager.vala
giacomoalbe: Sure!

giacomoalbe:
> > use the `event_bus` singleton (which is now linked to Akira.Application, even though connecting it to Akira.Window would be more correct)
> 
> Indeed, technically it would be more correct, but then we'd need to pass with `Window` instance to all the classes that need to connect to events, right?

I've come up with 2 ideas, but I don't truly know which is easier to implement or whether those are feasible at all.
* One is to get, in each item that needs event_bus, the window the item belongs to. This could be (easily?) done by traversing the hierarchy tree upwards till we reach a null parent (which will likely be the window). Then we can attach the event_bus to the window and use it with a tiny overhead (get_current_window ().event_bus)
* The other would be to have to completely different instance of the application running, so two different Gtk.Main () loops, instead of different windows. I don't truly believe a multi window application would work, but this could be a more feasible solution.

PR #233, opened 2020-01-04
[WIP] Fill Panel Connection with Selected Object

giacomoalbe:
## Summary
This PR is intended for restoring the connection between the currently selected object and the fill item panel, with the new EventBus centric approach.

## Things To Do

- [x] Restore connection of fill-item on object selection
- [x] In color file-entry only 0-9A-F chars can be inserted
- [x] Background color change using color picker
- [x] Alpha change using color picker
- [x] Set color in color entry only with hex characters
- [x] Change color and opacity for Ellipse
- [x] Change color and opacity for Text

## Known Issues

* Opacity and alpha rounding: I think we should deal better with opacity and alpha values. I also think that having an integer range between 0 and 100 (backed by an uint) would be enough, as this would improve the rouding of the values between the sliders, the model and the entries. 

* The CanvasItem interface it's not the most suitable place to code the "common" logic, since it's not possible to set and get properties directly. I think we should modify the way in which we inherit the canvas item, maybe by defining a common class (not an interface) with all the common properties set (color, border_color, alpha, opacity, width, height, x, y, etc) and with a custom "item" that should be populated (together with the other specific properties) in the subclasses (like CanvasRect, CanvasEllipse, CanvasText, CanvasImage, CanvasPath, etc). This needs to be better discussed, but I think it's necessary to evolve organically and not being forced to dirty hacks and workarounds.

### DIFF
--- a/src/Layouts/Partials/FillItem.vala
+++ b/src/Layouts/Partials/FillItem.vala
@@ -167,23 +167,19 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
             }
         );
 
-        color_container.key_release_event.connect ((event) => {
-            if (!color_container.text.contains ("#")) {
-                var builder = new StringBuilder ();
-                builder.append ("#");
-                builder.append (color_container.text);
-                color_container.text = builder.str;
-                color_container.set_position (-1);
-            }
+        color_container.delete_text.connect ((start_pos, end_pos) => {
+            string new_text = color_container.text.splice (start_pos, end_pos);
 
-            return false;
+            if (!new_text.contains ("#")) {
+                GLib.Signal.stop_emission_by_name (color_container, "delete-text");
+            }
         });
 
         color_container.insert_text.connect ((_new_text, new_text_length) => {
-            string new_text = _new_text;
+            string new_text = _new_text.strip ();
 
             if (new_text.contains ("#")) {
-                new_text = _new_text.substring (1, new_text.length - 1);
+                new_text = new_text.substring (1, new_text.length - 1);
             }
 
             bool is_valid_hex = true;
@@ -191,16 +187,15 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
             bool char_is_numeric = true;
             bool char_is_valid_alpha = true;
 
-            uint keyval;
+            char keyval;
 
             for (var i = 0; i < new_text.length; i++) {
-                keyval = Gdk.keyval_to_upper (new_text [i]);
+                keyval = new_text [i];
 
                 char_is_numeric = keyval >= Gdk.Key.@0 && keyval <= Gdk.Key.@9;
                 char_is_valid_alpha = keyval >= Gdk.Key.A && keyval <= Gdk.Key.F;
 
-                is_valid_hex &= (char_is_numeric || char_is_valid_alpha);
-
+                is_valid_hex &= keyval.isxdigit ();
             }
 
             if (!is_valid_hex) {
--- a/src/Layouts/Partials/FillItem.vala
+++ b/src/Layouts/Partials/FillItem.vala
@@ -222,7 +222,7 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
         opacity_container = new Akira.Partials.InputField (
             Akira.Partials.InputField.Unit.PERCENTAGE, 7, true, true);
         opacity_container.entry.sensitive = true;
-        opacity_container.entry.text = Math. round((double) alpha / 255 * 100).to_string ();
+        opacity_container.entry.text = Math. round ((double) alpha / 255 * 100).to_string ();
         opacity_container.entry.bind_property (
             "text", model, "alpha",
             BindingFlags.BIDIRECTIONAL | BindingFlags.SYNC_CREATE,
--- a/src/Layouts/Partials/FillItem.vala
+++ b/src/Layouts/Partials/FillItem.vala
@@ -99,6 +99,8 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
         hidden = model.hidden;
         //  blending_mode = model.blending_mode;
         color = model.color;
+
+        debug (@"model.color: $(model.color)");
     }
 
     private void create_ui () {
--- a/src/Layouts/Partials/FillItem.vala
+++ b/src/Layouts/Partials/FillItem.vala
@@ -99,8 +99,6 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
         hidden = model.hidden;
         //  blending_mode = model.blending_mode;
         color = model.color;
-
-        debug (@"model.color: $(model.color)");
     }
 
     private void create_ui () {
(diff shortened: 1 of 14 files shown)


src/Lib/Models/CanvasItem.vala
```diff
@@ -25,15 +25,17 @@ public enum Akira.Lib.Models.CanvasItemType {
     TEXT
 }
 
-public interface Akira.Lib.Models.CanvasItem : Goo.CanvasItem {
+public interface Akira.Lib.Models.CanvasItem : Goo.CanvasItemSimple, Goo.CanvasItem {
     public static int global_id = 0;
 
     public abstract string id { get; public set; }
```
Alecaddd: Do you need to specify a `public` setter? I've never seen this before.

giacomoalbe: No I don't have actually, I was following the compiler's suggestions. In the end even without public or protected the code worked the same, so I removed it!

src/Layouts/Partials/FillsBoxPanel.vala
```diff
@@ -80,5 +81,20 @@ public class Akira.Layouts.Partials.FillsBoxPanel : Gtk.Grid {
...
+        if (selected_items.length () > 0) {
+            if (selected_item == null || selected_item != selected_items.nth_data (0)) {
+                selected_item = selected_items.nth_data (0);
+
+                fills_list_model.add.begin (selected_item);
+            }
+        } else {
+            selected_item = null;
+            fills_list_model.clear.begin ();
+        }
```
Alecaddd: Too verbose, please change this into:
```
if (selected_items.length () == 0) {
    selected_item = null;
    fills_list_model.clear.begin ();
    return;
}

if (selected_item == null || selected_item != selected_items.nth_data (0)) {
    selected_item = selected_items.nth_data (0);
    fills_list_model.add.begin (selected_item);
}
```

src/Lib/Canvas.vala
```diff
@@ -140,6 +140,8 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     }
 
     public override bool button_press_event (Gdk.EventButton event) {
+        focus_canvas ();
+
```
Alecaddd: Do we need to set the focus on the canvas if a button is pressed?
Isn't the very own action of clicking on the canvas moving the focus there?

src/Lib/Managers/ItemsManager.vala
```diff
@@ -21,6 +21,8 @@
 
 
 public class Akira.Lib.Managers.ItemsManager : Object {
+    private const string DEFAUL_BORDER_COLOR = "#ccc";
```
Alecaddd: DEFAULT, you missed a T

src/Lib/Managers/ItemsManager.vala
```diff
@@ -202,7 +207,11 @@ public class Akira.Lib.Managers.ItemsManager : Object {
 
     private void udpate_default_values () {
         border_size = settings.set_border ? settings.border_size : 0.0;
-        border_color = settings.set_border ? settings.border_color: "";
-        fill_color = settings.fill_color;
+
+        string border_color_str = settings.set_border ? settings.border_color : DEFAUL_BORDER_COLOR;
```
Alecaddd: There's no need for a default border color.
If the `settings.set_border` is `false` it means the user doesn't want a default border color.
And if we need to have a default color for the model, let's use the `settings.border_color` without the necessity of defining another constant variable.

src/Layouts/Partials/FillItem.vala
```diff
@@ -159,24 +217,37 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
         opacity_container = new Akira.Partials.InputField (
             Akira.Partials.InputField.Unit.PERCENTAGE, 7, true, true);
         opacity_container.entry.sensitive = true;
-        opacity_container.entry.text = (alpha / 255 * 100).to_string ();
+        opacity_container.entry.text = Math. round ((double) alpha / 255 * 100).to_string ();
```
Alecaddd: Remove extra space after `Math.`

src/Lib/Managers/ItemsManager.vala
```diff
@@ -202,7 +207,11 @@ public class Akira.Lib.Managers.ItemsManager : Object {
...
+        string fill_color_str = settings.fill_color;
+
+        border_color.parse (border_color_str);
+        fill_color.parse (fill_color_str);
```
Alecaddd: No need to define the `fill_color_str` variable if there's no logic or condition attached to it.
Just type hint it when you parse it:
`fill_color.parse ((string) settings.fill_color);`

src/Layouts/Partials/FillItem.vala
```diff
@@ -159,24 +217,37 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
...
-            }, (binding, srcval, ref targetval) => {
-                double src = (double) srcval / 255;
-                targetval.set_string (("%0.1f").printf (src * 100));
+            }, (binding, model_alpha_val, ref entry_text_val) => {
+                double src = (double) model_alpha_val.get_int () / 255;
+
+                src = Math.round (src * 100);
+
+                entry_text_val.set_string (("%f").printf (src * 100));
+
```
Alecaddd: Why the `src` is multiplied by `100` twice?

src/Layouts/Partials/FillItem.vala
```diff
@@ -243,12 +314,7 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
     private void on_color_changed () {
         var selected_color = color_chooser_widget.rgba;
 
-        color = "#%02X%02X%02X".printf (
-            (int) (selected_color.red * 255),
-            (int) (selected_color.green * 255),
-            (int) (selected_color.blue * 255));
-
-        alpha = selected_color.alpha;
+        color = selected_color.to_string ();
```
Alecaddd: No need to declare the `selected_color` variable if it's used only once.
You can directly do:
`color = color_chooser_widget.rgba.to_string ();`

Alecaddd:
> Opacity and alpha rounding.

Yes, let's dot that.

> The CanvasItem interface it's not the most suitable place to code the "common" logic, since it's not possible to set and get properties directly.

I was thinking about that in order to avoid code duplication.
Maybe we could implement a `Controller` class that takes care of all the logic. The various shape classes can then extend it and trigger inherited methods accordingly.

We can discuss that later on, no need to tackle this right now.

src/Lib/Canvas.vala
giacomoalbe: Yes, clicking on the canvas moves the focus on the canvas, but not the "keyboard" focus, so we need to restore this focus on the canvas in order to listen again for the keyboard related actions (delete, insert, etc). This is especially true when use put text into the enties. After your insertion the Canvas "naturally" doesn't receive back the keyboard focus, so we need to force the focus update.

src/Layouts/Partials/FillItem.vala
giacomoalbe: Typo, removed!

src/Lib/Canvas.vala
Alecaddd: Sounds good, thanks for the explanation.

PR #234, opened 2020-01-04
Canvas pan with SPACE + drag and middle mouse drag

giacomoalbe:
## Summary
This PR is intended for restoring canvas pan with space + drag and middle mouse drag.

## Things To Do

- [x] Add MODE_PAN to canvas
- [x] Add PAN with SPACE + drag
- [x] Add PAN with middle mouse button drag

## Known Issues

* `Gdk.CursorType` does not have the "grabbing" icon. In order to have it, we need to change the way in which we set the cursor (like changing from "set_cursor" to "set_cursor_from_name"). It is better to add a "set_cursor_by_name" method (doubling the "current_cursor" logic to prevent looping and repetition) or do we change the way in which we set the cursor?

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -211,6 +211,8 @@ public class Akira.Lib.Canvas : Goo.Canvas {
 
             case EditMode.MODE_PAN:
                 set_cursor_by_edit_mode ();
+
+                canvas_scroll_set_origin (temp_event_x, temp_event_y);
                 break;
         }
 
@@ -260,6 +262,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                 break;
 
             case EditMode.MODE_PAN:
+                canvas_moved (event_x, event_y);
                 break;
         }
 
@@ -303,7 +306,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     }
 
     private void set_cursor (Gdk.CursorType? cursor_type) {
-        debug (@"Setting cursor: $cursor_type");
+        // debug (@"Setting cursor: $cursor_type");
         current_cursor = cursor_type;
 
         var cursor = new Gdk.Cursor.for_display (Gdk.Display.get_default (), cursor_type);
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -47,6 +47,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         MODE_SELECTION,
         MODE_INSERT,
         MODE_PAN,
+        MODE_PANNING,
     }
 
     public Managers.SelectedBoundManager selected_bound_manager;
@@ -91,6 +92,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     }
 
     public void set_cursor_by_edit_mode () {
+        // debug ("Calling set_cursor_by_edit_mode");
         Gdk.CursorType? new_cursor;
 
         switch (_edit_mode) {
@@ -103,7 +105,11 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                 break;
 
             case EditMode.MODE_PAN:
-                new_cursor = holding ? Gdk.CursorType.HAND2 : Gdk.CursorType.HAND1;
+                new_cursor = Gdk.CursorType.HAND2;
+                break;
+
+            case EditMode.MODE_PANNING:
+                new_cursor = Gdk.CursorType.HAND1;
                 break;
 
             default:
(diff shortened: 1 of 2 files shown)


Alecaddd:
@giacomoalbe do you want to give priority to this one?
The method we had before was working pretty nicely, and I did some updates to remove unwanted recursions and loops.

Be sure to keep an eye on these things:
- Use the delta when dealing with 2 fingers swiping to ensure a natural acceleration.
- Do some debug to be sure the events are not called multiple times during a single interaction.
- Try to keep the code small and clean, without bouncing off methods between multiple files.

src/Lib/Canvas.vala
```diff
@@ -262,12 +307,16 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
-            set_cursor (cursor_type);
+            return;
         }
+
+        set_cursor (cursor_type);
     }
 
     private void set_cursor (Gdk.CursorType? cursor_type) {
+        // debug (@"Setting cursor: $cursor_type");
+        current_cursor = cursor_type;
```
Alecaddd: Move this at the end of the function as we don't want to change the current_cursor until the GdkCursor was updated.

src/Lib/Canvas.vala
```diff
@@ -39,8 +36,10 @@ public class Akira.Lib.Canvas : Goo.Canvas {
             return _edit_mode;
         }
         set {
-            _edit_mode = value;
-            set_cursor_by_edit_mode ();
+            if (_edit_mode != value) {
+                _edit_mode = value;
+                set_cursor_by_edit_mode ();
```
Alecaddd: Since you're calling the `set_cursor_by_edit_mode ()` method here, you don't need to call it anywhere else as it will be automatically triggered whenever the edit mode changes.
Right now, if you debug the method, you'll see it gets called constantly at every mouse movement.
Leave the call only here and remove it everywhere else.

src/Lib/Canvas.vala
```diff
@@ -201,10 +230,22 @@ public class Akira.Lib.Canvas : Goo.Canvas {
 
         holding = false;
 
+        if (event.button == Gdk.BUTTON_MIDDLE) {
+            edit_mode = EditMode.MODE_SELECTION;
+        }
+
         //item_moved (selected_item);
         //add_hover_effect (selected_item);
```
Alecaddd: Do we need these? Maybe we can remove them.

src/Lib/Canvas.vala
```diff
@@ -127,6 +129,10 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                 // delete_selected ();
```
Alecaddd: Can we remove this?

src/Lib/Canvas.vala
```diff
@@ -262,12 +307,16 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     private void on_request_change_cursor (Gdk.CursorType? cursor_type) {
         if (cursor_type == null) {
             set_cursor_by_edit_mode ();
```
Alecaddd: You can leave this here since we needed to handle cursor changes for nob selection, when the edit mode doesn't change, but you should add a check to see if the current cursor is identical to the requested cursor, in order to prevent calling this constantly at every mouse motion.

src/Lib/Canvas.vala
giacomoalbe: No, just a leftover! Gonna remove it!

src/Lib/Canvas.vala
giacomoalbe: Yes

src/Lib/Canvas.vala
giacomoalbe: Ok, I think I have to rework the logic in the `HoverManager` in order to emit the event only when the hovered nob changes. 

Right now it triggers way too many events, we could make a flip-flop logic to decrease the event_bus load. 

Thanks for the suggestion, I'll make those changes as soon as I can.

src/Lib/Canvas.vala
giacomoalbe: This is generally true, the main issue there is that I need to call this function when the user clicks because I need to update the cursor in PAN mode when holding. 

To overcome this situation (that I see more as a workaround or bad coding, at least) I think we can have 2 different ´edit_mode´s, one is MODE_PAN and the other MODE_PANNING. The first is only triggered by pressing space on the keyboard, the second is when you are pressing space and clicking or clicking with the MIDDLE_MOUSE_BUTTON. What do you think?

src/Lib/Canvas.vala
giacomoalbe: Ok

src/Lib/Canvas.vala
Alecaddd: Yes, I like it.
That removes the necessity of having a `holding` variable, right?
I think it might be cleaner.

giacomoalbe: Negative, sir! The `holding` variable is still necessary for other tiny things (like hovering, object moving and I think something else too). I'm going for this approach then!

PR #235, opened 2020-01-04
[WIP] SelectionBound scale using Affine Transform

giacomoalbe:
## Summary
This PR is intended to address scaling of SelectedBound object using combination of transforms and not a sequence of translate and width/height editing in order to achieve custom pivot point scaling.

This work can be the foundation for multiple object scaling and transform in general.

##  Things To Do

- [ ] Move translate + width/height modification to combination of scale, transform & rotate matrices
- [ ] Implement scaling around a custom point function


Alecaddd:
> Implement scaling around a custom point function

Let's not do this for now as it might get complicated.
Let's focus on one thing at a time :smiley:

Alecaddd:
Closing this as the original scope was partially covered by #265 and the other edits are not a priority for the alpha release.

PR #237, opened 2020-01-04
Improve EventBus

Alecaddd:
I experimented a lot with the `EventBus` and tried multiple solutions, but ended up going with the easiest and cleaner one by simply calling the signal directly from the `event_bus` singleton.

I think this is the better approach since it doesn't cause us issues in handling a different amount of arguments per signal, and keeps the file clean and easy to update.
I'm not concerned about the lost of logging or history in that file as those can happen directly in the classes receiving those signals.

Just for posterity, here's a list of methods I tried and failed:
- Base everything on anonymous functions to remove the necessity of creating signals.
- Use the `Signal.emit_by_name` to call the signal from its string reference, and passing variable arguments:https://valadoc.org/gobject-2.0/GLib.Signal.emit_by_name.html
- Convert the EventBus into a Subscriber that handles each signal as an anonymous callback method.

All these attempted approaches were kind of OK, but the code was getting pretty verbose and not linear.
I prefer to stick with the default signal emission approach of Vala, which is clean and simple to read and write.

**EventBus as Window's singleton**
Since all our canvas interactions will be based on signals, the `EventBus` can't be a property of the `Gtk.Application` as all the signals will be emitted throughout every single opened Window instance.
I moved the `EventBus` to the `Akira.Window` class and turned it into a singleton.
Now events are containerized and limited to the currently active window.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -31,7 +31,7 @@ public class Akira.Application : Gtk.Application {
     construct {
         flags |= ApplicationFlags.HANDLES_OPEN;
 
-        event_bus = new Akira.Services.EventBus ();
+        event_bus = Akira.Services.EventBus.get_default ();
         settings = new Akira.Services.Settings ();
         windows = new GLib.List<Window> ();
         opened_files = new Gee.HashMap<string, Akira.Window> ();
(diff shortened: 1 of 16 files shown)


Alecaddd:
@albfan let me know if you have time to review this and what do you think about this approach.

giacomoalbe:
I think this approach is good, I found it pretty neat. We need to hold a reference to either window or canvas, and I think that every widget we might imagine would have at least one of those, both sometimes, either.

PR #238, opened 2020-01-06
Fix fill item regressions

Alecaddd:
After the merge of #233, lots of regressions were introduced.

Here's a list of what I discovered so far:
- [x] `reset_colors()` method called multiple times at every item selection.
- [x] `set_button_color()` called multiple times at every item selection.
- [x] Alpha value ignored when edited from color picker.
- [x] Border size and color issues from settings since conversion from `string` to `Gdk.RGBA`.

I'm gonna work on this and tackle these issues.
The majority of these issues are due to the verbose code style of handling items, models, and managers. There's a lot of repetition and things can be simplified a lot.

### DIFF
--- a/src/Layouts/Partials/FillItem.vala
+++ b/src/Layouts/Partials/FillItem.vala
@@ -14,7 +14,7 @@
  * GNU General Public License for more details.
 
  * You should have received a copy of the GNU General Public License
- * along with Akira.  If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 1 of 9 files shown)


Alecaddd:
The `reset_colors()` method was called 4 times at every selection or color change on the item.
This was happening because the method was getting called every time the `fill_color`, `border_color`, `alpha`, or `opacity` were edited.
The problem is that those values are not independent when using the color chooser.
Since all those values are managed by the `FillsItemModel`, listening for the changes on the model allows us to trigger that method only once.

PR #240, opened 2020-01-07
Improve SelectBound and FillPanel

Alecaddd:
Continuing the code improvements started in #238 this PR is taking care of:

- [x] Avoid triggering `reset_selection ()` when clicking on an already selected item.
- [x] Avoid triggering `reset_selection ()` when clicking on the canvas if not item was selected.
- [x] Avoid loop segfault when adding/removing a fill item model.
- [x] Remember the `hidden_fill` option.
- [x] Allow deletion of the Fill model.
- [x] Allow adding a Fill model if the list is empty.
- [x] Show/hide the Fill Panel based on selected item.

I wanted to keep this PR as small as possible, but while working on the `hidden_fill` option, I stumbled upon some regressions, so I decided to tackle those as well since they were causing segfaults.

### DIFF
--- a/src/Layouts/LeftSideBar.vala
+++ b/src/Layouts/LeftSideBar.vala
@@ -22,7 +22,7 @@
 public class Akira.Layouts.LeftSideBar : Gtk.Grid {
     public weak Akira.Window window { get; construct; }
     public Akira.Layouts.Partials.TransformPanel transform_panel;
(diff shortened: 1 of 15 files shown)


Alecaddd:
> I just suggest to split formal changes like copyright in separate commits.

You're right, I need to be more careful with those.
Thanks

PR #241, opened 2020-01-07
Implement Border panel

### LINKED ISSUE #92: Left Panel: Create Border Box
UI Mockup:
border-box

- The border box appears only if an object has the ability to edit this attribute. Eg. Shape, image, and grouped layers can have a border attribute, text layers can't.
- Create the possibility to have multiple borders by clicking on the `+` icon.
- Each border row can independently control its color, stroke type, and thickness.
- For v1, implement the most common stroke types: Solid, Dashed, Dotted.

When a stroke type different from solid gets selected, a second row should appear (or maybe an option button) to allow control over dashed and dotted spacing.

For v1, we will only support centered borders. Outside and Inside borders will be explored in a future version.

Alecaddd:

Add the Border Panel and copy the behaviour of the Fill Panel, obviously with its own dedicated Border Model.

- [x] Change color and border thickness.

~Change border style from `solid` to `dashed`.~
~Offer options to manage dash gap.~

### Fixes
- Fixes #92

### DIFF
--- a/src/Models/ListModel.vala
+++ b/src/Models/ListModel.vala
@@ -22,29 +22,55 @@
 
 public class Akira.Models.ListModel : GLib.Object, GLib.ListModel {
     private GLib.List<Akira.Models.FillsItemModel?> fills_list;
+    private GLib.List<Akira.Models.BordersItemModel?> borders_list;
+    public ListType list_type;
+    public enum ListType {
+        FILL,
+        BORDER
+    }
+
+    public ListModel (ListType type) {
+        list_type = type;
+    }
 
     construct {
-        fills_list = new GLib.List<Akira.Models.FillsItemModel?> ();
+        if (list_type == ListType.FILL) {
+            fills_list = new GLib.List<Akira.Models.FillsItemModel?> ();
+        } else {
+            borders_list = new GLib.List<Akira.Models.BordersItemModel?> ();
+        }
     }
 
     public uint get_n_items () {
-        return (uint) fills_list.length ();
+        if (list_type == ListType.FILL) {
+            return (uint) fills_list.length ();
+        }
+        return (uint) borders_list.length ();
     }
 
     public Object? get_item (uint position) {
+        Object? o = null;
         //  debug ("get item %u", position);
-        var o = fills_list.nth_data (position);
+        if (list_type == ListType.FILL) {
+            o = fills_list.nth_data (position);
+        }
+
         if (o != null) {
             return o as Object;
         }
+
         return null;
     }
 
     public Type get_item_type () {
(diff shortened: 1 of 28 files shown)


Alecaddd:
This PR is getting pretty big, so I will deal with the dashed option in a separated PR.

Alecaddd:
Give a special look to the `Models/ListModel.vala` file (former `FillsListModel.vala`) as I've updated that to handle both Fills and Borders models, trying to not duplicate the code as much as possible.

Alecaddd:
:heart_eyes: :heart_eyes: :heart_eyes: 
akira-fill-border-2

src/Models/ListModel.vala
```diff
@@ -0,0 +1,147 @@
...
+    public enum ListType {
+        FILL,
+        BORDER
+    }
+
+    public ListModel (ListType type) {
+        list_type = type;
+    }
+
+    construct {
```
giacomoalbe: I think that the idea behind this refactor is good (get the logic out of `FillListModel` and put it inside a new object) but this approach is not going to scale well. 

The ListModel object should be used elsewhere too (`LayerPanel` I'm thinking about you!), so hardcoding every possible list types (with the `ListType`) is not going to work, IMHO. 

Maybe we are not going to deal with 30 different types of list, but the code might start to become convoluted and we could introduce a lot of custom logic which is not related to the `ListModel` itself in here (like the action triggered on the removal of the last element, which does something on the BordersItemModel the ListModel is and should not be resposible for).

Moreover, we are creating (and thus instantiating) one list for every "ListType" each time we use the ListModel (in this case, twice, one for each panel - fill & border -).

My suggestion is to create 2 different base class, `ListModel` and `ItemModel` in which we declare only the very specific logic (add, remove, edit, connect, etc) and then inherit the `ItemModel` to handle the custom logic we want associated with the model. 

In this case, we could create a `BorderItemModel` and `FillItemModel` in which we put the custom logic and properties. The "if this is the last element" logic could be implemented in the BorderPanel by connecting to the "on_item_remove" event of `ListModel` (or whatever the name of the event might be), since this check is more related to the BorderPanel than to the general `ListModel`. 

I don't know if this could be the matter of a more narrowed scoped PR, given the fact that the current version of `ListModel` works out just fine, but in my opinion this would benefit the logic a lot more and in the end would provide us a new truly reusable component, with every aspect or concern of the code clearly separated and in the proper context.

Alecaddd: That's a good point, but I wouldn't do that here.
As you suggested, I think we should implement it in a more narrowed scoped PR.
I'll open an issue for it.

PR #247, opened 2020-01-08
meson: Ignore custom vapis depending on vala version

### LINKED ISSUE #243: Removing custom vapis
## Expected Behavior
Since version 0.43.90, vala provide updated vapis from goocanvas-2.0.

Would be great to remove our custom vapis and stick to upstream

Here I have vala 0.46

## Current Behavior
We maintain a custom vapi for goocanvas

## Possible Solution
remove custom vapis

## Context
flatpak still need this as the vala version still do not provides those vapis

@bilelmoussaoui I see io.elementary.App provides vala 0.40.

We need 0.43.90 to be able to remove custom vapis for goocanvas.

What's the roadmap for vala version in elementary SDK?

https://github.com/flathub/io.elementary.BaseApp

- Akira Version: master
- OS Info: arch linux

albfan:
Avoid custom vapis if vala provide them

Keep up to date with upstream if possible

- Fixes #243

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,10 +5,14 @@ akira_prefix = get_option('prefix')
 akira_datadir = join_paths(akira_prefix, get_option('datadir'))
 akira_pkgdatadir = join_paths(akira_datadir, meson.project_name())
 
-add_project_arguments(
-    ['--vapidir', join_paths(meson.current_source_dir(), 'vapi')],
-    language: 'vala'
-)
+vala_version = meson.get_compiler('vala').version()
+if vala_version.version_compare('<=0.43.90')
+    add_project_arguments(
+        ['--vapidir', join_paths(meson.current_source_dir(), 'vapi')],
+        language: 'vala'
+    )
+endif
+
 # Include the translations module
 i18n = import('i18n')
 # Dependencies


Alecaddd:
@albfan thanks for taking care of this.
There's an error in the GitHub CI:
`meson.build:14:4: ERROR:  Missing language definition in add_project_arguments`

meson.build
```diff
@@ -6,9 +6,16 @@ akira_datadir = join_paths(akira_prefix, get_option('datadir'))
...
-    ['--vapidir', join_paths(meson.current_source_dir(), 'vapi')],
     language: 'vala'
 )
+
+vala_version = meson.get_compiler('vala').version()
+if vala_version.version_compare('<=0.43.90')
+    add_project_arguments(
+        ['--vapidir', join_paths(meson.current_source_dir(), 'vapi')]
+    )
+endif
```
Alecaddd: I guess the `add_project_arguments` doesn't add extra arguments but overrides what was defined before, as the `language` argument is missing when the vala version is <= 0.43.90

bilelmoussaoui: The language should be moved to the project definition at the very top of the meson.build file to avoid this.

Alecaddd: Ah wait, the language is already defined in the project definition at the beginning.
If I'm not wrong, the `language` attribute is only necessary if we're defining the `vapi`.
I'll give it a try.

Alecaddd: Yes, I was right.
Does it build normally with an updated version of Vala?

albfan: It does.

PR #256, opened 2020-01-10
Improve ListModel architecture

### LINKED ISSUE #248: Rework ListModel
Once #241 lands, we will have a `ListModel` used for both Fills and Borders.
The implementation done is not the best and it wouldn't scale properly.

We need a better and more modular logic in order to handle different items.
As a reference, here's @giacomoalbe's comment on that PR:

> I think that the idea behind this refactor is good (get the logic out of FillListModel and put it inside a new object) but this approach is not going to scale well.
The ListModel object should be used elsewhere too (LayerPanel I'm thinking about you!), so hardcoding every possible list types (with the ListType) is not going to work, IMHO.
Maybe we are not going to deal with 30 different types of list, but the code might start to become convoluted and we could introduce a lot of custom logic which is not related to the ListModel itself in here (like the action triggered on the removal of the last element, which does something on the BordersItemModel the ListModel is and should not be resposible for).
Moreover, we are creating (and thus instantiating) one list for every "ListType" each time we use the ListModel (in this case, twice, one for each panel - fill & border -).
My suggestion is to create 2 different base class, ListModel and ItemModel in which we declare only the very specific logic (add, remove, edit, connect, etc) and then inherit the ItemModel to handle the custom logic we want associated with the model.
In this case, we could create a BorderItemModel and FillItemModel in which we put the custom logic and properties. The "if this is the last element" logic could be implemented in the BorderPanel by connecting to the "on_item_remove" event of ListModel (or whatever the name of the event might be), since this check is more related to the BorderPanel than to the general ListModel.

Alecaddd:
Improve and rework the `ListModel` in order to make it usable for different models without code duplication.

- Fixes #248

### DIFF
--- a/src/Layouts/Partials/BordersPanel.vala
+++ b/src/Layouts/Partials/BordersPanel.vala
@@ -66,7 +66,7 @@ public class Akira.Layouts.Partials.BordersPanel : Gtk.Grid {
         title_cont.attach (label, 0, 0, 1, 1);
         title_cont.attach (add_btn, 1, 0, 1, 1);
 
-        list_model = new Akira.Models.ListModel (Akira.Models.ListModel.ListType.NONE);
+        list_model = new Akira.Models.ListModel ();
 
         borders_list_container = new Gtk.ListBox ();
         borders_list_container.margin_top = 5;
--- a/src/Layouts/Partials/BordersPanel.vala
+++ b/src/Layouts/Partials/BordersPanel.vala
@@ -66,7 +66,7 @@ public class Akira.Layouts.Partials.BordersPanel : Gtk.Grid {
         title_cont.attach (label, 0, 0, 1, 1);
         title_cont.attach (add_btn, 1, 0, 1, 1);
 
-        list_model = new Akira.Models.ListModel (Akira.Models.ListModel.ListType.BORDER);
+        list_model = new Akira.Models.ListModel (Akira.Models.ListModel.ListType.NONE);
 
         borders_list_container = new Gtk.ListBox ();
         borders_list_container.margin_top = 5;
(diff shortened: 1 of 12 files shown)


Alecaddd:
@albfan @giacomoalbe @bilelmoussaoui anyone available to review this?

src/Models/ItemModel.vala
```diff
@@ -0,0 +1,24 @@
...
+* You should have received a copy of the GNU General Public License
+* along with Akira. If not, see <https://www.gnu.org/licenses/>.
+*
+* Authored by: Alessandro "alecaddd" Castellani <castellani.ale@gmail.com>
+*/
+
+public abstract class Akira.Models.ItemModel : GLib.Object {
+    public Lib.Models.CanvasItem item { get; construct; }
+    public Akira.Models.ListModel list_model { get; set; }
+    public Akira.Utils.BlendingMode blending_mode { get; set; }
```
giacomoalbe: Why do we need to have the blending model in here? 

I think this is a leftover, this should go in the `FillItemModel` and/or in the `BorderItemModel`, since it's not a general attribute of the model. 

Do you agree?

src/Models/FillsItemModel.vala
```diff
@@ -20,7 +20,7 @@
 * Authored by: Alessandro "alecaddd" Castellani <castellani.ale@gmail.com>
 */
 
-public class Akira.Models.FillsItemModel : GLib.Object {
+public class Akira.Models.FillsItemModel : Models.ItemModel {
```
giacomoalbe: The `blending_mode` property should have been declared here, since it's a specific property of this particular model

src/Layouts/Partials/FillsPanel.vala
```diff
@@ -91,16 +92,25 @@ public class Akira.Layouts.Partials.FillsPanel : Gtk.Grid {
...
+            var model_item = create_model ();
+            list_model.add_item.begin (model_item);
             selected_item.reset_colors ();
             add_btn.hide ();
             window.main_window.left_sidebar.queue_resize ();
         });
+
+        list_model.items_changed.connect ((position, removed, added) => {
+            if (selected_item != null) {
+                selected_item.has_fill = (added == 1);
```
giacomoalbe: Maybbe adding a comment to clarify what this code actually does would make it better. 

For me it's not super easy to get the meaning of this at a first glance, so maybe adding some hints would help our future self to understand what's happening. 

What do you think?

src/Layouts/Partials/BordersPanel.vala
```diff
@@ -91,16 +91,25 @@ public class Akira.Layouts.Partials.BordersPanel : Gtk.Grid {
...
+            var model_item = create_model ();
+            list_model.add_item.begin (model_item);
             selected_item.reset_colors ();
             add_btn.hide ();
             window.main_window.left_sidebar.queue_resize ();
         });
+
+        list_model.items_changed.connect ((position, removed, added) => {
+            if (selected_item != null) {
+                selected_item.has_border = (added == 1);
```
giacomoalbe: The same here. Adding a comment would make this statement a bit clearer.

src/Models/ItemModel.vala
Alecaddd: Indeed. I left it there because I was going to deal with it once adding the dashed style for the border. I guess I can change it now.

src/Layouts/Partials/FillsPanel.vala
Alecaddd: Good point.

PR #257, opened 2020-01-10
Allow scale item in any direction

albfan:
Creating new items, right now there's only a direction to scale, bottom/right

Scaling existing items each nob stops at its counter part

This PR allows to scale in any direction
 
## Steps to Test
Create a new item and scale bypassing other border

## Screenshots 

scale-in-any-direction

This still do not reset correctly the size so more refinement is still needed

- Relates to #206

### DIFF
--- a/src/Utils/AffineTransform.vala
+++ b/src/Utils/AffineTransform.vala
@@ -177,6 +177,16 @@ public class Akira.Utils.AffineTransform : Object {
             origin_move_delta_y = 0.0;
         }
 
+        if (new_width < MIN_SIZE) {
+            canvas.window.event_bus.flip_item ();
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


Alecaddd:
I tested this feature on sketch, and I noticed that it handles it by flipping the object, forcing a click on the `flip_horizontal` or `flip_vertical` button, as those a toggle buttons it activates them accordingly.
Would that be something useful to copy?

albfan:
Ah yes, it makes sense for items not symmetric, will add that

Alecaddd:
@albfan, once #270 is merged, you can use `window.event_bus.flip_item ();` to quickly flip the object on resize. That should take care of any problem.

Alecaddd:
I updated the PR, I hope you don't mind.
Super tiny update :smile:

PR #261, opened 2020-01-14
Basic image

donadigo:
##  Summary / How this PR fixes the problem?
This branch adds a barebones image support to Akira. At the moment you can:
* change size/position/rotation of the image
* change it's opacity

This PR is open also to discuss the workflow regarding importing images. Currently this PR just opens a FileChooser dialog that lets you open multiple images at once however there are of course other ways to do this.

## Screenshots
akira-image
 

## Changes in this PR
* `IMAGE` is now a new `CanvasItemType` but it's not an `InsertType`, that is because adding an image doesn't involve changing the cursor and drawing on the canvas, adding the item is separately done in `ActionManager`(?)

* A new method `insert_item_default` in `Canvas` was introduced to enable other components to insert items totally on their own 

* A new interface called `ImageProvider` was introduced to allow implementing new image sources in the future and as a general class to be passed into things that want an image

* Since images are things that take time to load, initially the item will have width and height of `1`. A new signal was added `item_bound_changed` to signal the selected bounds manager that the bounding box of an item changed and it should update it's selection too

* The bigger the image the slower the operations become on it. This is unfortunately a limitation of the implementation of `Goo.CanvasImage` and Cairo in general. The image is drawn by a CPU so it takes a lot of time.

* Placing the image in the current viewport. Currently the image is just placed to a constant x and y of `CANVAS_SIZE/2` 

* Handle errors reading the chosen files. Probably throw a dialog here or something else. May need some new classes for general error reporting.


src/Lib/Models/CanvasImage.vala
```diff
@@ -0,0 +1,81 @@
...
+    
+    public bool has_fill { get; set; default = false; }
+    public int fill_alpha { get; set; }
+    public Gdk.RGBA color { get; set; }
+    public bool hidden_fill { get; set; }
+    public bool has_border { get; set; default = false; }
+    public int border_size { get; set; }
+    public Gdk.RGBA border_color { get; set; }
+    public int stroke_alpha { get; set; }
+    public bool hidden_border { get; set; }
```
Alecaddd: You missed the `public bool has_border_radius { get; set; }` attribute and the build is failing.

src/Lib/Canvas.vala
```diff
@@ -87,6 +87,22 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
     }
 
+    public void insert_item_default (Akira.Lib.Models.CanvasItem item, bool select = false) {
+        items_manager.add_item (item);
+        // TODO: place in the viewport
+        Utils.AffineTransform.set_position (
+            Akira.Layouts.MainCanvas.CANVAS_SIZE / 2,
+            Akira.Layouts.MainCanvas.CANVAS_SIZE / 2,
+            item    
+        );
```
Alecaddd: We could implement a check to see if an item is currently selected, and use its X & Y coordinates to position the image, and if not, put it in the middle of the canvas like you did.

src/Lib/Canvas.vala
```diff
@@ -87,6 +87,22 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
+
+        if (select) {
```
Alecaddd: Is there any scenario where we don't want to select the last inserted image?

src/Lib/Canvas.vala
```diff
@@ -87,6 +87,22 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
+            selected_bound_manager.reset_selection ();
+            selected_bound_manager.add_item_to_selection (item);
+            nob_manager.selected_nob = Managers.NobManager.Nob.BOTTOM_RIGHT;
```
Alecaddd: Do we need to select the Nob here?

src/Lib/Canvas.vala
donadigo: In the code, no, but let's say you import a lot of images and the app should (?) just select the last one.

Alecaddd: Yeah, I think only the last one should be selected, at least for now.
Once we implement the multi-selection option we will update this.
For now, it'd be great to have `select` always `false` unless is the last uploaded image.

src/Services/ActionManager.vala
```diff
@@ -216,6 +218,34 @@ public class Akira.Services.ActionManager : Object {
...
+        switch (response_id) {
+            case Gtk.ResponseType.ACCEPT:
+            case Gtk.ResponseType.OK:
+                SList<File> files = dialog.get_files ();
+                weak Akira.Lib.Canvas canvas = window.main_window.main_canvas.canvas;
+                files.@foreach ((file) => {
+                    var provider = new Akira.Services.FileImageProvider (file);
+                    var item = new Akira.Lib.Models.CanvasImage (window.event_bus, provider, canvas.get_root_item ());
+
+                    canvas.insert_item_default (item, true);
```
Alecaddd: Here we could use `true` only if it's the last item of the list.

PR #265, opened 2020-01-16
Improve Selection Managers and Transform Panel bindings

Alecaddd:
This is gonna be a pretty big PR as I stumbled upon some pretty weird shortcomings and architectural issues we need to solve before moving forward.

- `Nobs` and `SelectEffect` are constantly removed and created at every mouse movement, event if the object selected is always the same, those effects are not simply updated like they should be.
- The `SelectEffect` redraw is getting called manually from the `SelectBoundManager` every time we change a parameter. We should rather listen to the model changes of the `CanvasItem` in order to call the redraw automatically.
- The `TransformPanel` doesn't update its values when the item changes, but it actually constantly removes and adds the selected item to fetch new values. We should update this section to use bidirectional bindings, and deal with freezes and segfault.

- [x] Fix infinite remove/create loop of select effect.
- [x] Fix manual call of select effect redraw.
- [x] Use bindings for the Transform Panel.
- [x] Add <kbd>CTRL</kbd> lock size on item creation.


Alecaddd:
This is getting big and out of hand. I'm gonna split it in smaller PRs

Alecaddd:
I think I fixed everything, sorry for the size of this PR but there were a lot of tiny little pieces interconnected with each other, and changing one thing was breaking other two.

Here's what I end up doing:
- The selection bound and nobs are created only once and not at every mouse movement.
- I was able to bind all the properties, except for X and Y, as those are not directly accessible from the CanvasItem.
- I'm using only affine transform and simplified the the event calling.
- The selection bounds are redrawn automatically when the selected item changes one of its values.

I also created a little bit more separation of concerns, meaning that if we change a method related to the affine transform, it shouldn't destroy everything else with a cascading effect.

Alecaddd:
I'm merging this as it fixes a lot of problems and improve the underlying architecture.

PR #267, opened 2020-01-17
Uneven border radii

albfan:
Allow rectangles with uneven radius borders

Using an open MR from goocanvas 

https://gitlab.gnome.org/GNOME/goocanvas/tree/wip/albfan/rect-radii 

connect uneven border entries to edit corners.

## Steps to Test

This works from flatpak. Using a submodule is another option not implemented

## Screenshots 

uneven-border-radius

- Consider CSS notation where `10/20` allows to set x and y radius (to draw ellipsize corners
- goocanvas branch will end in a goocanvas-3.0 which could be problematic for some long term support distros.

- Fixes #259


Alecaddd:
Awesome implementation for flatpak!
Would you be able to implement the git submodule for mason? Without that I don't feel comfortable in merging it as it break the elementary building process.

src/Lib/Models/CanvasRect.vala
```diff
@@ -89,11 +89,19 @@ public class Akira.Lib.Models.CanvasRect : Goo.CanvasRect, Models.CanvasItem {
     }
 
     public void update_border () {
+        //We only use one value (don't do ellipses)
         if (is_radius_uniform) {
             set ("radius-x", radius_x);
             set ("radius-y", radius_x);
+        } else {
```
Alecaddd: Use a `return;` here instead of an `else`.

albfan: Yes, let me release that on a goocanvas-3.0 branch so we get any other improvement for free.

albfan:
Should we revisit this now goocanvas 3.0 is out?

https://mail.gnome.org/archives/ftp-release-list/2021-January/msg00074.html

See #301

Alecaddd:
Yes, we definitely should!

Alecaddd:
Now that goocanvas 3 is in, this is gonna be a kick-ass feature!

Alecaddd:
Closing this since we dropped goocanvas and we're dealing with this directly in the app with Cairo.

PR #271, opened 2020-01-18
Better input fields

Alecaddd:
This PR has the purpose of improving the usability of the various input fields by using the `Gtk.SpinButton` for the entry element in order to leverage built-in value validations and increase/decrease actions.

**Reasons for this changes**
- The spin button comes with a built in scroll whell detection to increase/decrease value.
- I comes with built in "press-and-hold" detection to increase/decrease value.
- Removes the shenanigans of converting numeric values into strings and viceversa.
- It's a native GTK element and it's easier to style than a custom solution.
- It's a paradigm known to users.

## Screenshots 
spin-buttons

- ~For some reason, manually typing the values doesn't update the model binded to it until the focus moves away, or the value is changed via the spin buttons.~ Fixed
- ~The spin buttons are a bit too big and they add a lot of noise in the UI.~ Kinda Fixed
- ~It's not possible to bind directly the value change as the `SpinButton` doesn't have a `"value"` attribute to listen to.~ That's not true, I'm just not smart.

- [x] Allow better typing without force adding a `0`.
- [x] Non-numeric values are automatically ignored.
- [x] Increase/Decrease values via scroll wheel.

### DIFF
--- a/src/Partials/InputField.vala
+++ b/src/Partials/InputField.vala
@@ -58,7 +58,7 @@ public class Akira.Partials.InputField : Gtk.EventBox {
         entry.width_chars = chars;
         entry.sensitive = false;
 
-        entry.key_press_event.connect (handle_key);
+        entry.key_release_event.connect (handle_key);
 
         switch (unit) {
             case Unit.HASH:
@@ -99,20 +99,20 @@ public class Akira.Partials.InputField : Gtk.EventBox {
     }
 
     private bool handle_key (Gdk.EventKey key) {
-        if (key.state != Gdk.ModifierType.SHIFT_MASK) {
-            return false;
-        }
-
         // Arrow UP
-        if (key.keyval == 65362) {
+        if (key.keyval == 65362 && key.state == Gdk.ModifierType.SHIFT_MASK) {
             entry.spin (Gtk.SpinType.STEP_FORWARD, 10);
             return true;
         }
 
         // Arrow DOWN
-        if (key.keyval == 65364) {
+        if (key.keyval == 65364 && key.state == Gdk.ModifierType.SHIFT_MASK) {
             entry.spin (Gtk.SpinType.STEP_BACKWARD, 10);
             return true;
+        }
+
+        if (key.keyval != 65362 && key.keyval != 65364) {
+            entry.update ();
         }
 
         return false;
--- a/src/Partials/InputField.vala
+++ b/src/Partials/InputField.vala
@@ -1,5 +1,5 @@
 /*
-* Copyright (c) 2019 Alecaddd (http://alecaddd.com)
+* Copyright (c) 2019-2020 Alecaddd (https://alecaddd.com)
 *
 * This file is part of Akira.
 *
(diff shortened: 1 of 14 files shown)


christiankaindl:
One question about that:
> - [x] Increase/Decrease values via scroll wheel.

Does this also trigger when scrolling with a trackpad? And/or does the value only change if the input field is focused?

If the value changes on any scroll and without it being focused, you can't practically scroll the sidebar, because you would change the input fields accidentally all the time (something the current Inkscape suffers from)

Alecaddd:
@christiankaindl indeed, that's one of the shortcomings of the spin button, and yes, it's an annoying issue currently present in this PR.
I'm working on a fix to prevent value change if the field is not focused.
Thanks for the heads up :+1:

Alecaddd:
@christiankaindl I fixed it :smile: 
A simple check to see if the field was focused did the trick.
```
        if (!entry.has_focus) {
            GLib.Signal.stop_emission_by_name (entry, "scroll-event");
        }
```

Alecaddd:
@albfan @giacomoalbe do you guys have time to review this PR today?
I have another couple of PRs I'd like to push that depend on this.

src/Partials/InputField.vala
```diff
@@ -89,138 +93,41 @@ public class Akira.Partials.InputField : Gtk.EventBox {
...
+    public void set_range (double min_value, double max_value) {
+        entry.set_range (min_value, max_value);
     }
 
-    private bool handle_key (Gdk.EventKey key) {
+    private bool handle_key_press (Gdk.EventKey key) {
         // Arrow UP
-        if (key.keyval == 65362) {
-            increase_value (key);
+        if (key.keyval == 65362 && key.state == Gdk.ModifierType.SHIFT_MASK) {
```
giacomoalbe: Here, from a mere code readability point of view, it's better to use the `Gdk.Key.uparrow` and `Gdk.Key.downarrow` enum values, so it's immediately clear which key we are referring to.

src/Partials/InputField.vala
```diff
@@ -89,138 +93,41 @@ public class Akira.Partials.InputField : Gtk.EventBox {
...
+            entry.spin (Gtk.SpinType.STEP_FORWARD, 10);
             return true;
         }
 
         // Arrow DOWN
-        if (key.keyval == 65364) {
-            decrease_value (key);
+        if (key.keyval == 65364 && key.state == Gdk.ModifierType.SHIFT_MASK) {
```
giacomoalbe: The same as before

src/Partials/InputField.vala
```diff
@@ -89,138 +93,41 @@ public class Akira.Partials.InputField : Gtk.EventBox {
...
-        int num = key != null && key.state.to_string () == "GDK_SHIFT_MASK" ? 10 : 1;
-        double src = double.parse (entry.text) - num;
-        entry.text = src.to_string ();
-    }
-
-    public bool button_up_event (Gdk.Event event) {
-        if (event.type == Gdk.EventType.BUTTON_PRESS) {
-            increase_value (null);
+    private bool handle_key_release (Gdk.EventKey key) {
+        if (key.keyval != 65362 && key.keyval != 65364) {
```
giacomoalbe: The same as before.

src/Partials/InputField.vala
```diff
@@ -89,138 +93,41 @@ public class Akira.Partials.InputField : Gtk.EventBox {
...
         return false;
     }
 
-    public bool button_down_event (Gdk.Event event) {
-        if (event.type == Gdk.EventType.BUTTON_PRESS) {
-            decrease_value (null);
+    private bool handle_scroll_event (Gdk.EventScroll event) {
+        // If the input field is not focused, don't change the value.
+        if (!entry.has_focus) {
+            GLib.Signal.stop_emission_by_name (entry, "scroll-event");
```
ebassi: You can return `true` here and it will stop the signal emission at this point, preventing the default scroll event handler to run, without explicitly calling `stop_emission_by_name()`.

Alecaddd: Ah, thanks for the heads up.
If you have time you could create a small PR to fix it 😁

ebassi: Done :smiley:

Alecaddd: Thank you so much

PR #272, opened 2020-01-18
LayersPanel Restoration with ListModel

giacomoalbe:
This PR's goal is to reconnect the LayersPanel with the canvas again.

This implementation does use the `ListModel` approach to handle list update and sync with the canvas.

This enables a better separation of concerns between the various parts involved in the process (the canvas, the panel, the event bus) so that mixed code can be avoided as much as possible.

Right now is possible to create items in the canvas which are not bound nor belong to any `Artboard`.

When we'll introduce the `Artboard` management in the app, we should take care of differentiating the actual item delegate (the widget present in the LayerPanel list) between the 2 possible values (the `Layer` and the `Artboard`).

Each new layer is added on top of the other, in order to reflect the z-index ordering in the canvas in the LayersPanel as well.

## Steps to Test
Just add or remove items in the canvas, they should appear in the LayersPanel.

- Everything related to D&D and layers exchange between Artboard is not working (since we don't have Artboards yet) and may cause crashed. One possible solution would be to prevent drag dropping or to enable it only for z-ordering between layers.

- What does it mean to "lock" a panel? Prevent it from being selected?

- [x] Item insertion on the canvas layer sync
- [x] Item deletion on the canvas layer sync
- [x] Item visibility toggle
- [x] Item (layer)name change
- [x] Layer deleted causes item to be deleted
- [x] Selected item on canvas is in sync with selected layer
- [x] Hover effect on canvas is in sync with Hover item in LayersPanel
- [x] Layers in panel have zebra's stripes effect
- [x] Property `locked` on item prevents it from being _interacted with_ within the canvas


giacomoalbe:
That's what I'm currently working! I need your help and feedback in order to understand where to "stop" this PR (I just don't want it to become too big). Thanks!

albfan:
@giacomo locks means not move or resize.

giacomoalbe:
Ok, so it enough to add a property to `CanvasItem` (locked) to prevent selection in case it's true. Thanks!

Alecaddd:
> Ok, so it enough to add a property to `CanvasItem` (locked) to prevent selection in case it's true. Thanks!

Correct. A locked item is an item that can't be moved, selected, hovered, or interacted with in any way.
If the item is currently selected, and it gets locked, it should be removed from the list of selected items.

src/Layouts/Partials/Layer.vala
```diff
@@ -83,24 +85,23 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
 
     // public Akira.Shape shape { get; construct; }
```
Alecaddd: I think this is an old leftover, it can be removed.

src/Layouts/Partials/Layer.vala
```diff
@@ -551,6 +578,15 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
...
         activate ();
+
+        var new_label = entry.get_text ();
+
+        if (label.label == new_label) {
+            return;
+        }
+
+        label.label = new_label;
+        model.name = new_label;
```
Alecaddd: I think here you can do `label.label = mode.name = new_label;`

src/Models/LayerModel.vala
```diff
@@ -0,0 +1,76 @@
+/*
+* Copyright (c) 2019 Alecaddd (https://alecaddd.com)
```
Alecaddd: `2019-2020`

src/Models/LayerModel.vala
```diff
@@ -0,0 +1,76 @@
...
+        Akira.Models.ListModel list_model
+    ) {
+        Object (
+            item: item,
+            list_model: list_model
+        );
+    }
+
+    construct {
+    }
```
Alecaddd: You can remove the constructor method if you don't need it.

src/Layouts/Partials/LayersPanel.vala
```diff
@@ -49,17 +52,41 @@ public class Akira.Layouts.Partials.LayersPanel : Gtk.ListBox {
...
+        // real items and the layers panel only knows items model
+        item_model_map.@set (new_item.id, model);
+
+        show_all ();
+    }
+
+    private void on_item_deleted (Lib.Models.CanvasItem item_to_delete) {
+        var model_to_delete = item_model_map.@get (item_to_delete.id);
+
+        list_model.remove_item.begin (model_to_delete);
```
Alecaddd: Let's keep the variable names as simple and consistent as possible.
We can call those simply `item` and `model` as we don't need to be that specific inside an already eloquent method.

src/Lib/Managers/ItemsManager.vala
```diff
@@ -110,6 +113,11 @@ public class Akira.Lib.Managers.ItemsManager : Object {
         items.append (item);
     }
 
+    public void on_request_delete_item (Lib.Models.CanvasItem item_to_delete) {
+        item_to_delete.delete ();
+        canvas.window.event_bus.item_deleted (item_to_delete);
+    }
```
Alecaddd: Same thing here for the variable name, just call it `item`.

src/Layouts/Partials/Layer.vala
giacomoalbe: Ok!

src/Layouts/Partials/Layer.vala
giacomoalbe: Done! Thanks for the suggestion, I was not aware of this Vala feature (guess it's more of a C family thing, btw)

src/Models/LayerModel.vala
giacomoalbe: Sure

giacomoalbe: (Also I think we should update the Copyright notice in almost any new file. Could this be done automatically? Better a separate PR)

src/Layouts/Partials/LayersPanel.vala
giacomoalbe: Sure, I will do the same if it happens again!

src/Lib/Managers/ItemsManager.vala
giacomoalbe: Sure, as before!

src/Models/LayerModel.vala
giacomoalbe: Done!

giacomoalbe:
We have a problem with image items, since they are not created in the same way as other elements, so they cannot be inserted on the Canvas with a keyboard shortcut & they do not update the EventBus accordingly. 

Since the only connection between the `Canvas` and the `LayersPanel` which is missing is the `CanvasItem`, I think I won't add it to this PR and I'll create a new PR with the image insertion logic moved in the right place (which would be the `ItemsManager`) and also link image creation/deletion with the `LayersPanel`.

What do you think?

giacomoalbe:
I'd like to discuss a bit how to handle clicking on an item wrt LayersPanel:

* What happens when I click on an item in the `Canvas`?
__Proposed Solution__: in the layers panel, the correspondent item gets selected, so the orange badge it's added to the `LayerItem` in the list
* What happens when I click on a Layer in the LayerPanel? 
__Proposed Solution__: in the `Canvas` the item becomes selected (current selection is replaced with this new item) and it is centered in the view (in the same way it's would be with the `Ctrl 0` shortcut or similar)

I'll update this comment with other question related to `Canvas - LayersPanel` relative actions.

Alecaddd:
Sounds good for the image item.

Another thing I want to implement at a later time is the ability to draw the area where the image will be uploaded, like drawing a rectangle and then importing the image that will take the place and size of that rectangle. But that's for a future PR of course.

Sounds good for both later and item selection, except for centering the canvas.
Better not jump around the canvas at every selection as it can be visually jarring.
The user will be able to do it with the shortcuts.

Something nice to have if it's not hard, otherwise we can do it in another PR are:
- When hovering over a layer, the canvas item gets the hover effect.
- When a canvas item gets the hover effect, a similar effect highlights the layer.
- When selecting a canvas item, the layers panel scrolls (without animation) to the selected layer.

giacomoalbe:
> Sounds good for the image item.

Ok, copy! Next PR would be this one. 

> Another thing I want to implement at a later time is the ability to draw the area where the image will be uploaded, like drawing a rectangle and then importing the image that will take the place and size of that rectangle. But that's for a future PR of course.

Yes this could be nice, given that we respect its size ratio. 

> Sounds good for both later and item selection, except for centering the canvas.
> Better not jump around the canvas at every selection as it can be visually jarring.
> The user will be able to do it with the shortcuts.

Ok no centering. 


> Something nice to have if it's not hard, otherwise we can do it in another PR are:
> 
> * When hovering over a layer, the canvas item gets the hover effect.

Totally feasible

> * When a canvas item gets the hover effect, a similar effect highlights the layer.

Again totally feasible. What does hovering on the layer item mean? Bold layer name? Or maybe some kind of different background? 

> * When selecting a canvas item, the layers panel scrolls (without animation) to the selected layer.
Well, I don't think this is going to require little time, right now I'm not testing it with more than 10/15 items, so scrolling is not necessary. 
I think we can leave this for a next more focused PR.

Alecaddd:
Perfect 👍

Alecaddd:
For the hover effect on layers panel, I'd like the same effect we're using on the canvas item, a blue bordered rectangle.the layers panel has an overlay which allows us to draw things on top.

giacomoalbe:
This PR is big enough even without the canvas - layer panel z-index sync. So I'll stop it here and open a new one for those remaining stuff.

src/Layouts/Partials/Layer.vala
```diff
@@ -587,7 +649,8 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
                 button_hidden.get_style_context ().remove_class ("show");
             }
 
-            item.visibility = active ? Goo.CanvasItemVisibility.INVISIBLE: Goo.CanvasItemVisibility.VISIBLE;
+            model.visibility = !active;
+            //item.visibility = active ? Goo.CanvasItemVisibility.INVISIBLE: Goo.CanvasItemVisibility.VISIBLE;
 
             icon_visible.visible = active;
             icon_visible.no_show_all = ! active;
```
Alecaddd: Remove the space

src/Layouts/Partials/Layer.vala
```diff
@@ -587,7 +649,8 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
                 button_hidden.get_style_context ().remove_class ("show");
             }
 
-            item.visibility = active ? Goo.CanvasItemVisibility.INVISIBLE: Goo.CanvasItemVisibility.VISIBLE;
+            model.visibility = !active;
+            //item.visibility = active ? Goo.CanvasItemVisibility.INVISIBLE: Goo.CanvasItemVisibility.VISIBLE;
```
Alecaddd: Remove the comment if not needed

data/css/layer.css
```diff
@@ -23,9 +26,16 @@
...
 .layer.even:focus,
 .layer.even:selected {
+    border: 1px solid shade (@bg_color, 1);
     background-color: shade (@bg_color, 1);
 }
 
+/* Hovered */
+.layer.hovered,
+.layer.even.hovered {
+  border: 1px solid #41c9fd;
```
Alecaddd: define a variable for this color since it's the primary hovering color we'll be using throughout the app.

src/Layouts/Partials/Layer.vala
```diff
@@ -472,32 +507,50 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
...
+            });
+
+            if (layer_group != null) {
+                artboard.container.selection_mode = Gtk.SelectionMode.NONE;
+                artboard.container.unselect_row (layer_group);
             }
 
-            activate ();
+            //activate ();
+            //window.main_window.right_sidebar.layers_panel.selection_mode = Gtk.SelectionMode.NONE;
```
Alecaddd: Leftover comment?

src/Layouts/Partials/Layer.vala
```diff
@@ -472,32 +507,50 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
...
+
+        return false;
+    }
+
+    public bool on_key_release_event (Gdk.EventKey event) {
+        switch (event.keyval) {
+            case Gdk.Key.Delete:
+                window.event_bus.request_delete_item (model.item);
+                break;
+            default:
```
Alecaddd: No need for a default if there's nothing to declare.

src/Lib/Managers/ItemsManager.vala
```diff
@@ -39,7 +39,7 @@ public class Akira.Lib.Managers.ItemsManager : Object {
     public ItemsManager (Akira.Lib.Canvas canvas) {
         Object (
             canvas: canvas
-        );
+            );
```
Alecaddd: wrong indentation

src/Lib/Managers/ItemsManager.vala
```diff
@@ -120,24 +128,8 @@ public class Akira.Lib.Managers.ItemsManager : Object {
             border_color,
             fill_color,
             root
-        );
+            );
```
Alecaddd: same here

src/Models/LayerModel.vala
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    public string icon {
+        get {
+            return item.layer_icon;
+        }
+    }
+
+    public bool? visibility {
```
Alecaddd: Vala warning:
`warning: Type 'bool?' can not be used for a GLib.Object property`
Are we expecting anything else other than a boolean value here?
Also, maybe it would make more sense call this `is_visible`

src/Models/LayerModel.vala
```diff
@@ -0,0 +1,78 @@
...
+            return item.visibility == Goo.CanvasItemVisibility.VISIBLE;
+        }
+        set {
+            item.visibility = value
+                ? Goo.CanvasItemVisibility.VISIBLE
+                : Goo.CanvasItemVisibility.INVISIBLE;
+        }
+    }
+
+    public bool? locked {
```
Alecaddd: Same Vala warning here, and maybe we could call this `is_locked`

data/css/layer.css
giacomoalbe: Done

src/Layouts/Partials/Layer.vala
giacomoalbe: Well, kinda. It is something related to artboards selection or something, so I'm not totally sure what was the purpose of it. Nevertheless, when we'll tackle the artboard thing we'll account for this, now I simply remove it.

src/Models/LayerModel.vala
giacomoalbe: We are actually expecting a bool, but the problem is that a bool cannot be used as a property. That's why I used a `bool?` which can be used as a property, despite the warning. As you can see I'm also making use of setter and getter to update the item behind the model correcly, with a normal property getter and setter. That is to avoid using directly `model.item.locked` or `model.item.visibility`.

Actually the warning (which is build only) states something that's not completely true, since I'm using those as properties even though they _aren't supposed_ to be working as properties. It's very strange. I think that changing that would require a lot of hackish code, I'd rather stay with this warning and maybe ask help on the Gtk Discourse forum on how to get through this issue (this might also trigger a Vala change somewhere). What do you think? Alternative solution?

Wrt naming, I totally agree with you, we can change them, no problem.

src/Models/LayerModel.vala
giacomoalbe: Ok

src/Models/LayerModel.vala
Alecaddd: I did some research and talked with the Vala maintainer.
This was actually fixed in a recent patch landed in September: https://git.launchpad.net/vala/commit/?id=4e695a4e374df741859651f62e9a554f55991aac

So, future versions of Vala will remove this warning.
Good stuff!

PR #274, opened 2020-01-19
Fix item coordinates problem

Alecaddd:
The X and Y coordinates of an item don't always work as expected when the item is rotated.
It seems that updating those values via the Transform panel causes unexpected outcomes and repositioning.

## Steps to Test
- Create a shape.
- Rotate it (the bigger the rotation the more evident is the problem).
- Change its X or Y position via the Transform Panel multiple times.
- The item gets moved exponentially and the value in the input field doesn't match.

## Screenshots 
coords-fix

- [x] Refresh X & Y when rotating or resizing.
- [x] Prevent callback loops.
- [x] Fix X & Y issue when translating a rotated item.

~Probably regressed by #271, but maybe it was there the whole time and we didn't notice.~

**Update**
~I tested this with an old build and the problem persists.
The `AffineTransform` doesn't account for the current rotation. 
Eg. Increasing the Y it actually means decrease when the item is rotated, and vice-versa. Tricky.~

**Second Update**
Everything is fixed now.
I'm using the Cairo Matrix transform to ensure the translation is always calculated by the world's coordinates and not the (sometimes rotated) object's coordinates.

### DIFF
--- a/src/Layouts/Partials/TransformPanel.vala
+++ b/src/Layouts/Partials/TransformPanel.vala
@@ -224,7 +224,7 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
 
     private void enable () {
         canvas = selected_item.canvas as Akira.Lib.Canvas;
-        on_item_coord_changed.begin ();
+        on_item_coord_changed ();
 
         width.value = selected_item.get_coords ("width");
         height.value = selected_item.get_coords ("height");
@@ -293,7 +293,7 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
             (binding, val, ref res) => {
                 res = val.get_boolean ();
                 window.event_bus.flip_item (true);
-                on_item_coord_changed.begin ();
+                on_item_coord_changed ();
                 return true;
             });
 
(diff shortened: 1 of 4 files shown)


Alecaddd:
@albfan @giacomoalbe 
Hey guys, I tried to fix this but I'm not sure how to handle this issue.
Basically, once rotating, the increase and decrease value happening in the X & Y fields in the Transform Panel don't match anymore the actually X & Y coordinates of the item.

Example:
- Rotate an item 180 deg.
- Now the X coordinates are upside down, so if we increase the value in the field, we should actually decrease the the X value and vice-versa.
- This is not a linear problem since we can have 360 different variations and based on the uneven shape of the object, the coordinates can be all different.

I fixed the issue with a dirty hack, by delaying the update of the values by 1ms, and then forcing the fields to get the actual new values.
I'm not happy at all with this solution, but at least it allows moving the object without breaking it like it was happening before.

If one of you wants to take on this beautiful math problem, I think the solution should be to account for the current item rotation and do our X & Y translation without the rotation, so, even if the item is rotate or titled, X moves it always horizontally and Y moves it always vertically.

What do you think?

Alecaddd:
THIS!
We need to do this: https://www.cairographics.org/cookbook/matrix_conventions/

Alecaddd:
I FIXED IT!!!
In order to move an object by the world's coordinates I'm updating the Cairo Matrix.
The `translate()` method updates the object's coordinates, which are different from the world's coordinates if the object is rotated.

Sweet!

@giacomoalbe @albfan @bilelmoussaoui

PR #277, opened 2020-01-23
Move objects in canvas

Alecaddd:
Enable the ability to move the object with the arrow keys when the focus is on the canvas.

## Steps to Test
- Create an object.
- Use the arrow up, down, right, and left to move it around.
- Do a flip.

## Screenshots 
canvas-arrows


isneezy:
@Alecaddd IDK if you have plans to tackle this but please don't forget to include some sort of a way to move 10px instead of 1px, like when you do that using SHIFT modifier on Adobe Software.

Keep it up with the great work u're doing!
Thanks

Alecaddd:
@isneezy this feature is already present 😊

PR #278, opened 2020-01-24
Set size ratio to item and not canvas

Alecaddd:
Set size ratio to item and not canvas and properly calculate the aspect ratio of imported images.

## Steps to Test
- Import an image
- Resize it and be sure the `size-lock` is on in the Transform panel and the image maintain its aspect ratio.


Alecaddd:
@albfan 
> A side regression about flipping items and fixed ratio is that rotate nob goes to the bottom when resize goes in vertical

I'm OK with the rotation nob being at the bottom as I think the selection bound should follow the current object's coordinates. We should actually update the cursor to be aligned with the world coordinates. I'll do that in a follow up PR.

> fixed ratio do not allow to flip item vertically

Argh, I didn't noticed, I'll try to fix it.

Alecaddd:
@albfan I fixed it. It's a bit flimsy and not as smooth and precise as I'd like, but it's good for alpha.
We'll need some serious polish for this.

PR #281, opened 2020-01-24
Use the signal return value to stop its emission

ebassi:
Event-related signals in GTK widgets have a return value which control
the signal emission chain. Returning a false value means "continue the
signal emission", whereas returning a true value means "I handled the
event, please stop the emission chain here".

Instead of forcibly stopping the signal emission using the
`GLib.Signal.stop_emission_by_name()` method, we can return `true` to
stop the signal emission.


bilelmoussaoui:
Thanks @ebassi

PR #283, opened 2020-01-25
Segfault and CSS fixes

Alecaddd:
I stumbled upon a nasty segfault happening when quickly clicking on a locked shape.
We were triggering the `AffineTransform` before checking if we had a selected item and if the item wasn't locked.

Also, an itty bitty update to use box shadows instead of border colors for layer effects.
Box shadows are faster attributes to animate, don't cause the widget to grow in size, and don't force us to manage the left border color when selected.

## Screenshots 
layers-panel

### DIFF
--- a/src/Layouts/Partials/LayersPanel.vala
+++ b/src/Layouts/Partials/LayersPanel.vala
@@ -43,9 +43,9 @@ public class Akira.Layouts.Partials.LayersPanel : Gtk.ListBox {
 
     public LayersPanel (Akira.Window window) {
         Object (
-            window: window,
-            activate_on_single_click: false,
-            selection_mode: Gtk.SelectionMode.SINGLE
+        window: window,
+        activate_on_single_click: false,
+        selection_mode: Gtk.SelectionMode.SINGLE
         );
     }
 
@@ -60,8 +60,8 @@ public class Akira.Layouts.Partials.LayersPanel : Gtk.ListBox {
             // TODO: Differentiate between layer and artboard
             // based upon item "type" of some sort
             return new Akira.Layouts.Partials.Layer (
-                window,
-                (Akira.Models.LayerModel) item
+            window,
+            (Akira.Models.LayerModel) item
             );
         });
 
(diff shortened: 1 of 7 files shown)


src/Layouts/Partials/LayersPanel.vala
```diff
@@ -43,9 +43,9 @@ public class Akira.Layouts.Partials.LayersPanel : Gtk.ListBox {
 
     public LayersPanel (Akira.Window window) {
         Object (
-            window: window,
-            activate_on_single_click: false,
-            selection_mode: Gtk.SelectionMode.SINGLE
+        window: window,
```
giacomoalbe: Is this indentation right? To me it's better the previous version

src/Layouts/Partials/LayersPanel.vala
```diff
@@ -60,8 +60,8 @@ public class Akira.Layouts.Partials.LayersPanel : Gtk.ListBox {
             // TODO: Differentiate between layer and artboard
             // based upon item "type" of some sort
             return new Akira.Layouts.Partials.Layer (
-                window,
-                (Akira.Models.LayerModel) item
+            window,
```
giacomoalbe: Same in here

src/Models/LayerModel.vala
```diff
@@ -22,11 +22,7 @@
 public class Akira.Models.LayerModel : Models.ItemModel {
     public string name {
         owned get {
-            if (item.name != null) {
-                return item.name;
-            }
-
-            return item.id;
+            return item.name != null ? item.name : item.id;
```
giacomoalbe: I was thinking to change how we give ID's to item, stick to a pure numeric format and marshall the name (when not set by the user) by getting the `default_layer_name` property of each Canvas Item. This way we could make use of incremental numeric ID and have the complete human readable name of the Layer (like it was on the video of the previous version of the LayerPanel) like "Rectangle 0", "Ellipse 0", "Text 0", rather then the capitalized version of the ID which is present today. What do yo think?

Issue #244, opened 2020-01-07
Hide SelectEffect when editing the CanvasItem model properties


Alecaddd:
## Expected Behavior
When the user edits the properties of a `CanvasItem` the `SelectEffect` should disappear until the end of the interaction, with a delay of 2 seconds.

This behaviour is commonly used in other design tools to remove the visual noise of the selection bound when the user needs to edit the visual aspects of an item.

Eg. editing the color of a 1px thick border is pretty hard with the selection bound border in the way.

## Current Behavior
The selection effect is always visible.

## Possible Solution
We should emit a signal whenever a property we know can benefit from not having the selection visible is getting edited by the user.
The `SelectedBoundManager` can then hide itself and set a delay of 2 seconds before turning back visible.

I'm suggesting the use of a delay to avoid visual glitches when quickly editing a property. eg. typing the color or the opacity value.

Alecaddd:
This will also help us to solve the problem of updating the selection bounds when the border thickness changes.

PR #284, opened 2020-01-26
Bug fixes and code clean up

Alecaddd:
This is a house keeping PR that takes care of cleaning up and organizing methods and attributes, and fixes small quirks visual issues here and there.
Nothing new has been implemented.

- Organize the `Canvas.Models.*` attributes.
- Organize the `EventBus` signals.
- Get the `current_scale` from the `Canvas` without a dedicated attribute for each class.
- Update the `Select Effect` when zooming.
- Update the `Hover Effect` when zooming.
- Fix wrong positioning of `Select Effect` and `Nobs`.
- Update the `Rotation Nob` distance based on the `Canvas` scale.
- Fix a segfault when trying to update the nob position when no items has been selected.
- Prevent a zoom out below 10% as a 0% scale factor breaks the canvas.
- Use the native `Granite.AccelLabel` class for HeaderBar popover buttons.

## Screenshot
Screenshot from 2020-01-26 19-26-09

Enjoy!


giacomoalbe:
Great work! The new accels are very very good, I like them! 

I also wanted to clean the code up, and now that's everything is in its own container/box it's a lot easier and no more a merging nightmare!

Btw, I found some bugs related to the hover effect and the rotation nob when selecting. 

I also recorded a screencast to show the problem: when hovering over an item and ctrl scrolling the mouse wheel the hover effect's border also increases in size, but when losing hover and regaining it again the hover is the right size. 

hover_effect_on_zoom

Also the rotating nob is way to far when zooming extra large (to the maximum extent), the distance between the selection and the rotation nob is not always the same. 

Screenshot from 2020-01-27 10-08-27

Now I'm going to review the code, too!

Alecaddd:
> when hovering over an item and ctrl scrolling the mouse wheel the hover effect's border also increases in size.

I left that out as I will need to create extra signals and methods I didn't want to introduce in this PR.
More related to #244 than this.

> Also the rotating nob is way to far when zooming extra large

I'll fix that, thanks.

PR #285, opened 2020-01-27
WIP: Layer Panel Z Ordering with D&D

giacomoalbe:
Goal of this PR is to bring the drag and drop capability back to the LayerPanel, in order to sync the LayerPanel list of items with the current z-index state of the canvas.

Another goal of this PR is to document the process behind the drag&drop which currently is not completely clear. 

For now, I'll focus on bringing back the basic functionality of the panel (such us layer z-index ordering). 
When we'll handle the grouping and artboard features, we'll extend the inner working to also cover those new piece of functionality. 
 
## Steps to Test
* Move the layers in the LayerPanel with D&D to change the z-index ordering of the items in the canvas
* Change z-index ordering with the actions on the HeaderBar to see the changes reflected in the LayersPanel

## Screenshots
layer_z_ordering_d d

## TODO List

- [x] Changes in the z-index of an item is reflected in the LayersPanel
- [x] Drag & Drop in the LayersPanel changes the z-index of the item in the canvas

## Known Issues

- When hovering over the `indicator` the cursor changes, as this widget is seen as a drop target. This could be caused by the fact the the `indicator` is over the `LayersPanel` iteself. We need to enable dropping only on the `LayerPanel`. This could cause problems when the user tries to drop and item on the orange line of the `indicator`, which could be frequent. 
- Trying to drop an item at the beginning of the list could be tricky since the drop area is limited to the `LayerPanel` and it is possible that the cursor is released onto the searchbar. Maybe we could handle this by extending the drop zone to the `RightPanel` instead of having it limited to the LayersPanel.

### DIFF
--- a/src/Layouts/Partials/LayersPanel.vala
+++ b/src/Layouts/Partials/LayersPanel.vala
@@ -68,6 +68,7 @@ public class Akira.Layouts.Partials.LayersPanel : Gtk.ListBox {
         window.event_bus.item_inserted.connect (on_item_inserted);
         window.event_bus.item_deleted.connect (on_item_deleted);
         window.event_bus.selected_items_changed.connect (on_selected_items_changed);
+        window.event_bus.z_selected_changed.connect (on_z_selected_changed);
     }
 
     private void on_item_inserted (Lib.Models.CanvasItem new_item) {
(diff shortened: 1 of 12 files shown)


Alecaddd:
Isn't the order upside down? Layers should be created in the same order as the z-index, with a new layer appearing above existing ones.

src/Layouts/Partials/LayersPanel.vala
```diff
@@ -137,37 +162,48 @@ public class Akira.Layouts.Partials.LayersPanel : Gtk.ListBox {
...
-
         if (source == target) {
             return;
         }
 
-        remove (source);
-        insert (source, new_position);
+        window.event_bus.change_item_z_index (source.model.item, new_position);
+
+        window.main_window.right_sidebar.indicator.visible = false;
```
Alecaddd: We could use a signal here to clean this up.

giacomoalbe:
> Isn't the order upside down? Layers should be created in the same order as the z-index, with a new layer appearing above existing ones.

Well, I think you're right, I need to change the order so that when I create an object on the canvas, in the layers list it's going to be placed above the others.

src/Layouts/Partials/LayersPanel.vala
giacomoalbe: Do you refer to the action of hiding the `right_sidebar.indicator`? For sure!

PR #288, opened 2020-01-30
Initial selecting with mouse dragging

donadigo:
Some initial prototyping to get selecting multiple objects with mouse working (by dragging).

Problems:
- [ ] slow. For some reason setting `selected` on an item is very slow and expensive and the selection slows down when you drag over an item. This is a bad sign.


Alecaddd:
Closing this in favor of #520 since the Canvas architecture changed entirely from this initial PR.

PR #290, opened 2020-02-02
Export dialog UI

### LINKED ISSUE #275: UI: Implement export dialog
This is the first round of UI mock-ups to visually prototype the Export Dialog for the Alpha release.

The initial idea is to offer 3 options when clicking on the export.
export-popup

- **Export Selection:** What is currently selected will be wrapped around the bounds to generate the exported image. This gives the flexibility to select multiple items to export without manually drawing an area, creating a slice, or grouping those items together.
- **Export Artboards:** Allows to select 1 or more Artboards which will be exported in batch.
- **Export Area to Grab:** Much like the screenshot tool, allows to click&drag to select an area to export.
- **Accelerators** are visible so the users can slowly learn quick shortcuts to trigger specific exports.

Based on the selected option, an **Export Dialog** will show up allowing the control of some basic options.
export-dialog

- What was selected will be shown as a preview, and if multiple artboards were selected, each file will have its own editable name underneath its preview.
- The location will be selected with a native file dialog.

This are low fidelity mock-ups, so things might change.
Also, these are the basic features we want to implement for the alpha release. More advanced features and edits will be discussed in the future.

Alecaddd:
This PR implements **only** the UI of the Export Dialog and the GSettings options to properly remember user's settings.
No actual export logic or image manipulation is implemented here.

## Steps to Test
- Click on the `Export` toolbar button and select the last option "Export Area to Grab"
- Or use the keyboard shortcut <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>G</kbd>

## Screenshots 
Screenshot from 2020-02-01 16-37-28

There's a problem I can't seem to solve.
Whenever I use a `Gtk.ComboBox` inside a `Gtk.Dialog`, I get this warning in the terminal:
```
Gtk-CRITICAL: _gtk_widget_get_preferred_size_for_size: assertion 'size >= -1' failed
```

- Fixes #275

### DIFF
--- a/src/Dialogs/ExportDialog.vala
+++ b/src/Dialogs/ExportDialog.vala
@@ -30,8 +30,10 @@ public class Akira.Dialogs.ExportDialog : Gtk.Dialog {
     public Gtk.Adjustment compression_adj;
     public Gtk.Scale compression_scale;
     public Gtk.ComboBoxText file_format;
-    public Gtk.Label jpeg_title;
+    public Gtk.Label jpg_title;
     public Gtk.Label png_title;
+    public Gtk.Label alpha_title;
+    public Gtk.Switch alpha_switch;
 
     public ExportDialog (Akira.Window window) {
         Object (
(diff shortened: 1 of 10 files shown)


Alecaddd:
Rethinking some UI from the rebound made in #291 

Screenshot from 2020-02-03 18-40-45

Screenshot from 2020-02-03 18-40-56

giacomoalbe:
I like the idea of having the button group also for the "Format", as presented in #291. 

The main issue I see there is the fact that if we would like to support more that, let's say, 3/4 options the UI could become cluttered and unclean. In this case, the select it's a better fit. 

The rest of the UI is very good for me, can't wait to see the preview in the center of it!

src/Dialogs/ExportDialog.vala
```diff
@@ -0,0 +1,226 @@
...
+        grid.attach (folder_button, 1, 0, 1, 1);
+
+        // File format.
+        var format_title = section_title (_("Format:"));
+        grid.attach (format_title, 0, 2, 1, 1);
+
+        file_format = new Gtk.ComboBoxText ();
+        file_format.append ("png", "PNG");
+        file_format.append ("jpg", "JPG");
+        //  file_format.changed.connect (update_format_ui);
```
giacomoalbe: It seems that there's a bug in here, the first time the export dialog is opened, both the quaity and the compression scale are shown, even though the value of the `file_format` input is correctly populated from settings. 

Also, I think the `changed` event should be connected to `update_format_ui` (here it's commented). 
I tried to compile it without the comment and the switch works, but only after the first time I manually update the `file_format` field.

Issue #279, opened 2020-01-24
Image import dialog doesn't remember last opened folder


Alecaddd:
After importing a new image, if you try to import another one again, the dialog doesn't open the last folder where the image was imported.
I tried to use `set_current_folder()` but it didn't work.
@bilelmoussaoui do you have time to take a look at this?

albfan:
It should work: https://valadoc.org/gtk+-3.0/Gtk.FileChooserDialog.html

Any open WIP PR

albfan:
Uy

Alecaddd:
I didn't have time to open a WIP PR.
Feel free to tackle it if you want to 😊

bilelmoussaoui:
I can fix it this evening

Alecaddd:
@bilelmoussaoui did you find time to work on this?

bilelmoussaoui:
I just looked at this, the reason why we don't remember the path is that we use the portal to open the files. The portal doesn't give us the file location but it mounts it temporary somewhere. Hence we can't really get the path back again.

In order to fix that, we will have to open a hole in the sandbox to access home for our flatpak build and switch back to a simple GtkFileChooserDialog. Usability is much important in a design app

bilelmoussaoui:
I have a patch for that, but we are using a new granite version now so I have to update the BaseApp first

PR #291, opened 2020-02-03
Export dialog ui rebound

cassidyjames:
This is a rebound on the export dialog. My main goal was to follow elementary HIG while keeping all functionality. In summary:

- Gets rid of hardcoded sidebar and header width request and lets the contents determine the width
- Gets rid of separator grid in favor of valign on an action area grid
- Right-aligns the action buttons like elementary dialogs
- Uses margins instead of CSS padding
- Uses elementary HIG styled right-aligned labels
- Simplifies some of the copy
- Reduces the number of clicks by using ModeButtons instead of comboboxes

## Screenshots 
image

### DIFF
--- a/data/css/layout.css
+++ b/data/css/layout.css
@@ -15,7 +15,6 @@
 
 .sidebar-export {
     background-color: shade (@bg_color, 0.95);
-    padding: 10px 20px 20px;
     border-radius: 0 0 0 4px;
 }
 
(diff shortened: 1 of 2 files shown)


danrabbit:
Something to consider is in Flatpak land you won’t be able to write the file out to a location yourself. You’ll be able to spawn a filechooser dialog. So it might not be worth doing a design that’s a filechooser here

bilelmoussaoui:
> Something to consider is in Flatpak land you won’t be able to write the file out to a location yourself. You’ll be able to spawn a filechooser dialog. So it might not be worth doing a design that’s a filechooser here

I think that's fine as we will probably end up giving Akira at least access to home by default. A lot of usability things that can't be used as remembering the latest opened directory (see #279).

Alecaddd:
Thanks for the suggestions and insights.
I included a couple of your ideas into the main PR, so I'm gonna close this.
Thanks again, this is highly appreciated!

PR #294, opened 2020-02-17
Implement "Export Area to Grab"

Alecaddd:
This is almost ready to got. A couple of things are missing but the overall logic and workflow is there and it works. This wasn't easy to implement and I had to do a lot of refactors to properly make it work and prevent segfaults.
Give it a try and wait for a full PR soon.

## Steps to Test
* Create a shape/shapes or import an image.
* Press <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>G</kbd> or use the Export menu.
* Highlight the area to export.
* Interact with all the options in the dialog.
* Export!

## Screenshots 
export

- [x] Implement a `Gtk.Flowbox` to allow multiple previews in the dialog.
- [x] Implement `Filename` setting.

### DIFF
--- a/src/Lib/Managers/ExportAreaManager.vala
+++ b/src/Lib/Managers/ExportAreaManager.vala
@@ -147,20 +147,20 @@ public class Akira.Lib.Managers.ExportAreaManager : Object {
         SourceFunc callback = init_generate_pixbuf.callback;
 
         new Thread<void*> (null, () => {
-			try {
-				generate_pixbuf ();
-			} catch (Error e) {
-				error ("Could not generate export preview: %s", e.message);
-			}
+            try {
+                generate_pixbuf ();
+            } catch (Error e) {
+                error ("Could not generate export preview: %s", e.message);
+            }
 
-			Idle.add ((owned) callback);
+            Idle.add ((owned) callback);
             export_dialog.generate_export_preview.begin ();
             canvas.window.event_bus.preview_completed ();
             Thread.exit (null);
-			return null;
+            return null;
         });
 
-		yield;
+        yield;
     }
 
     public void generate_pixbuf () throws Error {
@@ -226,7 +226,7 @@ public class Akira.Lib.Managers.ExportAreaManager : Object {
                     (int) area.height / 2,
                     Gdk.InterpType.BILINEAR
                 );
-            break;
+                break;
 
             case 2:
                 scaled_image = pixbuf.scale_simple (
@@ -234,7 +234,7 @@ public class Akira.Lib.Managers.ExportAreaManager : Object {
                     (int) area.height * 2,
                     Gdk.InterpType.BILINEAR
                 );
-            break;
+                break;
 
             case 3:
                 scaled_image = pixbuf.scale_simple (
@@ -242,7 +242,7 @@ public class Akira.Lib.Managers.ExportAreaManager : Object {
                     (int) area.height * 4,
                     Gdk.InterpType.BILINEAR
                 );
-            break;
+                break;
 
             default:
                 scaled_image = pixbuf.scale_simple (
@@ -250,7 +250,7 @@ public class Akira.Lib.Managers.ExportAreaManager : Object {
                     (int) area.height * 1,
                     Gdk.InterpType.BILINEAR
                 );
-            break;
+                break;
         }
 
         return scaled_image;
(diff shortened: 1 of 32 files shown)


calslate:
I'm a little confused about the name… In the long run, this won't matter, but shouldn't it be **'Grab Area to Export'** not **'Export Area to Grab'**?

src/Lib/Managers/ExportAreaManager.vala
```diff
@@ -0,0 +1,287 @@
...
+    public Gdk.Pixbuf pixbuf;
+
+    public ExportAreaManager (Akira.Lib.Canvas canvas) {
+        Object (
+            canvas: canvas
+        );
+    }
+
+    public Goo.CanvasRect create_area (Gdk.EventButton event) {
+        var dash = new Goo.CanvasLineDash (2, 5.0, 5.0);
```
giacomoalbe: We need to account for the zoom here, so the `CanvasLineDash` line width needs to be divided by the current scale otherwise, at higher zoom, the line width is too high.

src/Lib/Managers/ExportAreaManager.vala
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        area.remove ();
+    }
+
+    public void create_export_snapshot () {
+        // Hide the area before rendering.
+        area.visibility = Goo.CanvasItemVisibility.INVISIBLE;
+        // Generate the image to export.
+        generate_pixbuf ();
```
giacomoalbe: Maybe this could be an improvement for next versions, not a bug. Since the `generate_pixbuf' could take a while to finish and in the meantime the canvas will be "freezed", I suggest to create and show the dialog beforehand and then to show a progress indicator (or simply a loader) so inform the user some action is taking place. 

Once the pixbuf generation has been done, the user is presented with the image and then he/she can perform other adjustments and could see the results in the window. 

The same can be said for when the user click on the Export button, right now there's no confirmation that the file has been correctly saved, so we could/should provide a screen to inform the user the action has been completed and that the newly created image is ready. 

Again, all those things can be accounted for in a future PR, but it seemed to me that this could be a good place to start this conversation. 

What do you think?

Alecaddd: Yes, these are very good points.
I think I will update all these method to be `async`, and add some multi threading when necessary, but in a follow up bug.

This is mostly the first barebone implementation of a single export, but once we start dealing with big artboards or multiple exports at once, progress bars and confirmation dialogs are definitely necessary.

src/Lib/Managers/ExportAreaManager.vala
Alecaddd: Good point, I completely forgot.

Alecaddd:
> I'm a little confused about the name… In the long run, this won't matter, but shouldn't it be **'Grab Area to Export'** not **'Export Area to Grab'**?

Yeah, that's true is not very clear.
I was trying to get a consistency with the other labels like "Export Selection" and "Export Artboars".
I guess something like "Highlight Area to Export" would make more sense.

src/Lib/Managers/ExportAreaManager.vala
Alecaddd: Nevermind, I implemented async methods and multithreading to deal with it right away :smile:

Alecaddd:
This is driving me slightly insane.
I implemented a `Gtk.FlowBox` and binded to a `ListStore` and it seems to work, except for some random segfault crash I've no clue why is happening.
Also, I can't seem to bind the entry text to the model filename. Super frustrating.

I will take a break and get back to it with a fresh mind.

Alecaddd:
@giacomoalbe @albfan do you guys have some time to take a look at this PR and see what I'm doing wrong?

Alecaddd:
> For me this is work is simply **AWESOME**!

😊️

> I've seen that the `quality` and `compression` sliders does not trigger a preview update, but I've noticed that they are not even connected to the image generation procedure, I think it's a wise choice for now, we can for sure _do it later_ ©

Yeah, that something I need to do manually as the `Gdk.Pixbuf` doesn't offer a byte estimate of the compressed image before the full export.
I'll write a custom algorithm to get an estimate.

> Again, this new UI looks incredibly cool, I'm completely amazed by the quality of your work and I can see a very bright future for Akira!

Ah, thanks, your PR are super awesome as well and the app is looking super slick also thanks to you.

> The multithreading also looks very promising.

Indeed, async and mult-threading are very easy to use and manage. Heavy operations will most likely all be handled in that way.

> As an aside, a little question: why do we need the `x11_dependency`?

Ah, that's a leftover for when I was trying different multi-threading approached.
Removed.

PR #295, opened 2020-03-02
Improve export dialog

Alecaddd:
The export dialog and overall workflow need a lot of love. 

## Screenshots 
export-size-preview

notification

- [x] Update file size preview based on compression and quality slider.
- [x] Implement async thread to save the buffered image.
- [x] Allow handling multiple exports at once.
- [x] Implement async notifications on the Main Canvas.

### DIFF
--- a/src/Dialogs/ExportDialog.vala
+++ b/src/Dialogs/ExportDialog.vala
@@ -21,7 +21,7 @@
 
 public class Akira.Dialogs.ExportDialog : Gtk.Dialog {
     public weak Akira.Window window { get; construct; }
-    public weak Akira.Lib.Managers.ExportAreaManager manager { get; construct; }
+    public weak Akira.Lib.Managers.ExportManager manager { get; construct; }
 
     public GLib.ListStore list_store;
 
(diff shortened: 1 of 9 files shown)


nisrulz:
Just an observation: 

The name of the export is in. somewhat not very intuitive place. What do you think about moving it to the side bar? The enough space there to use it for an edit field. Unless someone saw this Gif, I believe it is hard to figure out where to edit the name. I have nothing to back it up, so it is more of probably personal view on this. You could also maybe put up a poll for this somewhere or have an issue open where people can vote.

Alecaddd:
@nisrulz 
> The name of the export is in somewhat not very intuitive place. What do you think about moving it to the side bar? There is enough space to use it for an edit field.

The sidebar needs to be empty as it will be filled with repeatable export settings in the future (eg. you will be able to export the same images in different formats and resolutions at once).

This dialog will also be used when exporting multiple images or artboards at once, which will show a series of previews on the right side. Handling each file name in the sidebar when we don't really know if there's gonna be 1 image or 20, is not doable.

> You could also maybe put up a poll for this somewhere or have an issue open where people can vote.

I prefer to stay away from polls related to UX as the answers are mostly based on "I like this better" instead of actual accessibility and usability guidelines.

Nonetheless, this is just the first implementation, and many improvements and fixes will come with time.

nisrulz:
All valid point. Thank you for clarifying.

PR #296, opened 2020-03-06
Handle multiple selection export

Alecaddd:
This PR extends the basic functionalities of the `ExportManager` to allow dynamic exporting and handling of selections and custom areas.
This PR creates a distinction between areas and selections, handling those separately, and implementing a dynamic hide/show of the selection effect.

### Screenshot
export-selection

### DIFF
--- a/src/Dialogs/ExportDialog.vala
+++ b/src/Dialogs/ExportDialog.vala
@@ -22,6 +22,7 @@
 public class Akira.Dialogs.ExportDialog : Gtk.Dialog {
     public weak Akira.Window window { get; construct; }
     public weak Akira.Lib.Managers.ExportManager manager { get; construct; }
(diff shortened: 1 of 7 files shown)


Alecaddd:
@giacomoalbe 
> The only thing I'm noticing is that when I export the selection the generated image contains every item that is inside the bounding box of the selected element, which is not completely correct to me.

Yeah, that's true.
I'm not sure we can solve this problem right now as the generation of the pixbuf from goocanvas can be done exclusively by creating a Cairo Surface snapshot from the current canvas. The snapshot grabs everything that is currently visible, therefore all the elements will be there.

I guess, for future PRs, we could experiment with dynamically hide all the elements before generating the snapshot (exactly what I'm doing for the select effect and nobs).

Issue #122, opened 2019-03-13
How Artboards should behave


Alecaddd:
## What's an Artboard
An Artboard is a a virtual representation of the screen containing the design elements. Artboards are commonly used to represent multiple pages of a website, an ad banner, a business card, or any other medium related tot he same design style, all at once on the same canvas.

Artboards remove the necessity of defining a document size like in traditional graphic tools.

## Expected Behavior
An Artboard can be created like a regular rectangle shape, and the only editable attributes are:
- [x] Name
- [x] Size (WxH)
- [x] Background color (use the same fill panel)
- [x] Visibility
- [x] Lock status

## Layers Panel
- [x] Hover effect on the layer panel.
- [x] Nested items are created inside the dedicated artboard.

## Advanced Artboard features
- [x] Artboards can be selected on the canvas by clicking on their empty edges or on their names.
- [x] Artboards can't be rotated. (hide rotation nob and rotation slider).
- [x] If an Artboard is selected, it shouldn't be possible to hover or select any nested object (gif example below), and clicking & dragging should move the artboard and don't select nested objects.

## Nesting and grouping
- [x] Elements generated inside an Artboard are attached to it with a parent-child relationtship.
- [x] Many elements can belong to only 1 Artboard at a time.
- [x] Artboards cannot be nested/grouped.
- [x] When moving an Artboard that has elements inside, all the elements move accordingly as they were part of a grouped object.
- [x] An object that belongs to an artboard will be masked if is positioned outside the edge of its parent artboard.

## Example from Sketch
![artboard-example](https://user-images.githubusercontent.com/2527103/54255630-2098e980-4516-11e9-9f29-0c9d7f6e9484.gif)

amxmln:
What I love about Figma is that they got rid of the concept of artboards and use "frames" instead, which can be nested in each other. I think their handling of this is pretty much perfect.

What you're suggesting seems like a pretty standard way of handling them, basically a named group with a background color. That approach certainly works, but I'm not sure it's the most flexible one.

sr229:
+1 for @amxmln, I think Figma frames would be more useful in some scenarios than standard artboards.

Alecaddd:
@albfan here's the recap of how Artboard should behave.

Alecaddd:
@amxmln @sr229, thanks for the suggestion.
We will definitely look into the possibility of creating nested artboards, or even completely implementing the "frames" approach of Figma.

For version 1, we're keeping things simple and standard artboards are what we need.

[...]

PR #298, opened 2020-03-08
WIP: Artboard as Containers

giacomoalbe:
This is the initial work to bring _basic_ container functionality to the Artboard item. 
Items created inside an artboard and items moved inside one should become children of the artboard. 
At the same time, the Layers Panel should reflect the children of an artboard and enable the user to actionate them (`hide` and `lock`). 

As always, I'd like to get some feedback on how "far" should I get with this PR. 
I'd like to progress up until the "Item should snap to Artboard's edges" task. 

I have some questions, too:

- How can the user remove an item from an Artboard? 
- If the user enlarges an existing Artboard to contain some new items, should those new items be considered as Artboard's children? In other words, can Artboards _fagocitate_  other items while growing?
- More to come...

## Steps to Test
- Create and artboard
- Create a shape **inside** the artboard
- Move the artboard from the handle. The items should move together with the artboard

## Screenshots 
artboard_content_translation

## Implemented Features
- [x] Creating an item inside an Artboard should make the item a child of the Artboard
- [x] Moving an artboard should move its children in a rigid way
- [x] Child coordinates are relative to Artboard origin
- [x] Deleting an item inside an Artboard should remove it from that Artboard's children
- [x] Item have hover effect when hovering over the Artboard
- [x] Item can be selected & moved by clicking on them

Known Issues / Things To Do
- [ ] Items inside Artboard could be rotated with the nob
- [ ] Artboard should not be allowed to be rotated
- [ ] Artboard's label should be 15px regardless of the scale of the Canvas
- [ ] Artboard cannot be originated when inside another Artboard
- [ ] Dragging and dropping an item outside of an Artboard should remove it from the Artboard's children
- [ ] Moving an item inside an artboard should make the item a child of the Artboard
- [ ] The layers panel should show the children of an artboard
- [ ] Hide & Lock should work also on Artboard layer

- Implements (partially) #122

### DIFF
--- a/data/css/artboard.css
+++ b/data/css/artboard.css
@@ -24,7 +24,7 @@
     color: @fg_color;
 }
 
-artboard:selected {
+.artboard:selected {
     box-shadow: inset 4px 0 0 @primary;
 }
 
--- a/data/css/artboard.css
+++ b/data/css/artboard.css
@@ -8,7 +8,14 @@
     font-weight: bold;
     padding: 6px 10px;
     text-shadow: 0 1px 0 @bg_color;
-    border-bottom: 1px solid shade (@bg_color, 0.8);
+}
+
+.artboard .artboard-handle {
+  border-bottom: 1px solid shade (@bg_color, 0.8);
+}
+
+.artboard.hover .layer-action {
+  opacity: 0.5;
 }
 
 .artboard:hover,
@@ -17,7 +24,7 @@
     color: @fg_color;
 }
 
-.artboard:selected {
+artboard:selected {
     box-shadow: inset 4px 0 0 @primary;
 }
 
--- a/src/Lib/Managers/ItemsManager.vala
+++ b/src/Lib/Managers/ItemsManager.vala
@@ -24,6 +24,7 @@ public class Akira.Lib.Managers.ItemsManager : Object {
     public weak Akira.Lib.Canvas canvas { get; construct; }
 
     private List<Models.CanvasItem> items;
+    private List<Models.CanvasArtboard> artboards;
     private Models.CanvasItemType? insert_type { get; set; }
     private Goo.CanvasItem root;
     private int border_size;
@@ -38,7 +39,9 @@ public class Akira.Lib.Managers.ItemsManager : Object {
 
     construct {
         root = canvas.get_root_item ();
+
         items = new List<Models.CanvasItem> ();
+        artboards = new List<Models.CanvasArtboard> ();
 
         border_color = Gdk.RGBA ();
         fill_color = Gdk.RGBA ();
@@ -49,39 +52,59 @@ public class Akira.Lib.Managers.ItemsManager : Object {
     }
 
     public Models.CanvasItem? insert_item (Gdk.EventButton event) {
-        udpate_default_values ();
+      udpate_default_values ();
 
-        Models.CanvasItem? new_item;
+      Models.CanvasItem? new_item;
+      Goo.CanvasItem parent = root;
 
-        switch (insert_type) {
-            case Models.CanvasItemType.RECT:
-                new_item = add_rect (event);
-                break;
+      foreach (var artboard in artboards) {
+        if (artboard.is_inside (event.x, event.y)) {
+          parent = artboard as Goo.CanvasItem;
+        }
+      }
 
-            case Models.CanvasItemType.ELLIPSE:
-                new_item = add_ellipse (event);
-                break;
+      debug (@"Event.x $(event.x) Event.y $(event.y)");
 
-            case Models.CanvasItemType.TEXT:
-                new_item = add_text (event);
-                break;
+      switch (insert_type) {
+        case Models.CanvasItemType.RECT:
+          new_item = add_rect (event, root);
+          break;
 
-            case Models.CanvasItemType.ARTBOARD:
-                new_item = add_artboard (event);
-                break;
+        case Models.CanvasItemType.ELLIPSE:
+          new_item = add_ellipse (event, root);
+          break;
 
-            default:
-                new_item = null;
-                break;
+        case Models.CanvasItemType.TEXT:
+          new_item = add_text (event, root);
+          break;
+
+        case Models.CanvasItemType.ARTBOARD:
+          new_item = add_artboard (event);
+          break;
+
+        default:
+          new_item = null;
+          break;
+      }
+
+      if (new_item != null) {
+        if (new_item is Akira.Lib.Models.CanvasArtboard) {
+          artboards.append ((Models.CanvasArtboard) new_item);
+        } else {
+          items.append (new_item);
+
+          if (parent != root) {
+            var artboard = parent as Models.CanvasArtboard;
+
+            debug (@"Adding $(new_item.id) to $(artboard.id)");
+            artboard.add_child (new_item, -1);
+          }
         }
 
-        if (new_item != null) {
-            items.append (new_item);
+        canvas.window.event_bus.item_inserted (new_item);
+      }
 
-            canvas.window.event_bus.item_inserted (new_item);
-        }
-
-        return new_item;
+      return new_item;
     }
 
     public void add_item (Akira.Lib.Models.CanvasItem item) {
@@ -103,7 +126,8 @@ public class Akira.Lib.Managers.ItemsManager : Object {
         return artboard as Models.CanvasItem;
     }
 
-    public Models.CanvasItem add_rect (Gdk.EventButton event) {
+    public Models.CanvasItem add_rect (Gdk.EventButton event, Goo.CanvasItem parent) {
+        debug (@"Add rect parent is null: $(parent == null)");
         var rect = new Models.CanvasRect (
             Utils.AffineTransform.fix_size (event.x),
             Utils.AffineTransform.fix_size (event.y),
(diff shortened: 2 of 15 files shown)


Alecaddd:
> I'd like to progress up until the "Item should snap to Artboard's edges" task.

I agree, you can leave the snapping feature for another PR.

> How can the user remove an item from an Artboard?

He can delete it, or drag it outside.
As a general rule, we can use a 2 pixels threshold, meaning that if an object is at least 2 pixels inside the boundaries of an Artboard, the object belongs to that artboard and it's masked.
Once dropped (not dragged, only if dropped) "outside", the object goes to the canvas and it's not bounded to the artboard anymore.

> You should not be allowed to move an item out of an Artboard.

This will be allowed as explained above.

> If the user enlarges an existing Artboard to contain some new items, should those new items be considered as Artboard's children? In other words, can Artboards fagocitate other items while growing?

That's a good question. I'd say yes, but only at the end of the transformation, so once the "holding ends" as we shouldn't run that sort of condition during transformation as it might be taxing.

I don't know if we discussed this, but another important thing to add is the relative coordinates of items inside an artboard.
X and Y of an item should reflect the position inside that artboard and not the canvas.

src/Lib/Models/CanvasItem.vala
```diff
@@ -106,9 +123,103 @@ public interface Akira.Lib.Models.CanvasItem : Goo.CanvasItemSimple, Goo.CanvasI
...
+        case "y":
+          return item_y;
+        default:
+          return 0.0;
+      }
+    }
+
+    public virtual void store_relative_position () {
+      this.initial_relative_x = this.relative_x;
+      this.initial_relative_y = this.relative_y;
```
Alecaddd: This whole section has wrong indentation.

data/css/artboard.css
```diff
@@ -17,7 +24,7 @@
     color: @fg_color;
 }
 
-.artboard:selected {
+artboard:selected {
```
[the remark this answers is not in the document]
giacomoalbe: Yes definitely! I'll correct it

src/Lib/Models/CanvasItem.vala
giacomoalbe: Yes, I had the same feeling. Something has gone wrong with my editor! I'll try to revert to the `editorconfig` settings.

giacomoalbe:
> The app crashes with a segfault when deleting an item that wasn't created inside an artboard.

Thanks! I think my eye was tired enough not to see this case! I'll check what is the cause!

Regarding the rest, do you think it's ok, do you have some feedback on the design or on any other aspect of the implementation?

Alecaddd:
> Regarding the rest, do you think it's ok, do you have some feedback on the design or on any other aspect of the implementation?

It looks pretty good for what I can see.
You did an amazing job in tackling this complicated feature.
Let me know once you tackled those small issues I reported and I'll give it another review, and then merge it :smile:

giacomoalbe:
> The app crashes with a segfault when deleting an item that wasn't created inside an artboard.

Ok, I was able to figure out the issue, it was related to a double parent insertion of the child that probably resulted in a double free segfault during the deletion phase. 

I don't know why sometimes the `meson test` failed locally and sometimes not, but now it's working both on my development env and here on Github, so I'm more than happy to say that I leave this PR to your reviews again :)

src/Lib/Managers/ItemsManager.vala
```diff
@@ -52,18 +54,25 @@ public class Akira.Lib.Managers.ItemsManager : Object {
         udpate_default_values ();
 
         Models.CanvasItem? new_item;
+        Models.CanvasArtboard? artboard = null;
+
+        foreach (var _artboard in artboards) {
+            if (_artboard.is_inside (event.x, event.y)) {
+                artboard = _artboard;
+            }
+        }
```
Alecaddd: Did you try using the `get_item_at (event.x, event.y, true);` method?
Since the `Models.CanvasArtboard` extends the `Goo.CanvasItem` we should inherit that method.
With that method you can avoid looping through artboards, and do something like that:
```
var item = get_item_at (event.x, event.y, true);
if (item != null && item is Models.CanvasArtboard) {
     artboard = item as Models.CanvasArtboard;
}
```

src/Lib/Managers/NobManager.vala
```diff
@@ -350,7 +350,9 @@ public class Akira.Lib.Managers.NobManager : Object {
         if (selected_items.length () == 1) {
             var item = selected_items.nth_data (0);
 
-            item.get_transform (out transform);
+            transform = item.get_real_transform ();
+
+            // item.get ("line_width", out line_width);
```
Alecaddd: Do we need this?

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -62,10 +64,15 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+            selected_item.store_relative_position ();
+
+            delta_x_accumulator = 0.0;
+            delta_y_accumulator = 0.0;
+
             initial_event_x = event_x;
             initial_event_y = event_y;
 
-            canvas.convert_to_item_space (selected_item, ref initial_event_x, ref initial_event_y);
+            // canvas.convert_to_item_space (selected_item, ref initial_event_x, ref initial_event_y);
```
Alecaddd: Why this is commented?

src/Lib/Models/CanvasArtboard.vala
```diff
@@ -168,15 +223,65 @@ public class Akira.Lib.Models.CanvasArtboard : Goo.CanvasItemSimple, Goo.CanvasI
...
+    public unowned GLib.List<Goo.CanvasItem> get_items_at (double x, double y, Cairo.Context cr, bool is_pointer_event, bool parent_is_visible, GLib.List<Goo.CanvasItem> found_items) {
+        var artboard_x = x;
+        var artboard_y = y;
+
+        canvas.convert_to_item_space (this, ref artboard_x, ref artboard_y);
+        if (simple_is_item_at (
+                artboard_x, artboard_y,
+                cr, is_pointer_event)
+           ) {
+            found_items.append (this);
```
Alecaddd: Weird indentation.

src/Lib/Models/CanvasArtboard.vala
```diff
@@ -168,15 +223,65 @@ public class Akira.Lib.Models.CanvasArtboard : Goo.CanvasItemSimple, Goo.CanvasI
...
     }
 
     public override bool simple_is_item_at (double x, double y, Cairo.Context cr, bool is_pointer_event) {
-        var is_on_handle = y < 0;
+        // To select an Artboard you should put the arrow
+        // ontop of the Artboard label
+        var is_on_handle =
+            y < 0
+            && y > - (label_extents.height + LABEL_BOTTOM_PADDING)
+            && x < (label_extents.width);
```
Alecaddd: You can return this directly without declaring a variable.

src/Utils/AffineTransform.vala
```diff
@@ -35,55 +35,108 @@ public class Akira.Utils.AffineTransform : Object {
...
         double delta_y = GLib.Math.round (y - initial_y);
 
-        selected_item.translate (delta_x, delta_y);
+        delta_x_accumulator += delta_x;
+        delta_y_accumulator += delta_y;
+
+        selected_item.move (
+          delta_x, delta_y,
+          delta_x_accumulator, delta_y_accumulator
+        );
```
Alecaddd: Wrong indentation.

src/Utils/AffineTransform.vala
```diff
@@ -184,6 +242,19 @@ public class Akira.Utils.AffineTransform : Object {
...
 
+        delta_x_accumulator += origin_move_delta_x;
+        delta_y_accumulator += origin_move_delta_y;
+
+        selected_item.move (
+          origin_move_delta_x,
+          origin_move_delta_y,
+          delta_x_accumulator,
+          delta_y_accumulator
+        );
```
Alecaddd: Also here :smile:

src/Lib/Managers/ItemsManager.vala
giacomoalbe: Yes, the method `get_item_at` is used by the canvas to determine which item si currently being clicked. 

The problem here is that the `get_item_at` function does make use of the `is_item_at` function, which is overridden and used in the `CanvasArtboard` class to get whether or not the Artboard is selected or hovered, condition that is met only when we are pointing the mouse on the Artboard label, for now, and in the future also on the Artboard's edges. 

The `item` variable, thus, is always null or at least it does represent the item which is _behind_ the item we are trying to create inside the Artboard. 

Looping through artboards seems to be inevitable atm, but we shouldn't be concerned at all, since looping through 20 of them only takes 20-30 microseconds, so we could easily loop though something like 20k Artboard in less than a blink of an eye :)

src/Lib/Models/CanvasArtboard.vala
giacomoalbe: Sure, fixed!

src/Lib/Managers/SelectedBoundManager.vala
giacomoalbe: Now dealing with item's coordinate system is not so easy nor obvious, since the item inside artboard are painted by the Artboard itself, so their coordinate system is bound to their parent. 

Here we need to convert the event x,y to the item rotation, since it is the only affine transform we do to the item, since `relative_x` and `relative_y` are the new coordinate system of the object relative to their parent. 

I think that, by introducing the Artboard feature, we need to rethink the way in which we convert events from the canvas to the object. 

I made some test, but eventually I opted to focus on more urgent tasks (like connecting all the "wires" together) and after take focus on the topic of dealing with affine transform and the canvas. 

I hope it does make sense.

src/Lib/Managers/NobManager.vala
giacomoalbe: Yes, as per the answer I gave to the other comment related to the SelectedBoundManager.ù

We now need to take into consideration whether the item is inside an Artboard or nor. 
In the latter, the transformation does involve using the parent's coordinate system to calculate the proper transform to apply.

PR #300, opened 2020-03-10
Another round of bug fixes

Alecaddd:
A couple of our supporters on discord have pointed out some quirky bugs which I'm tackling in this patch.

- [x] The border radius doesn't evenly update when typing a value.
- [x] The Selection Bounds should ignore the border thickness and stay true to the item's coordinates.
- [x] Export Selection bounds don't account for border thickness.
- [x] Export Artboards bounds should ignore the name element when generating the preview.
- [x] Export Artboards should pre-populate the file name with Artboard's name.
- [x] Show hover effect on item when hovering over a layer.

### DIFF
--- a/src/Dialogs/ExportDialog.vala
+++ b/src/Dialogs/ExportDialog.vala
@@ -272,15 +272,16 @@ public class Akira.Dialogs.ExportDialog : Gtk.Dialog {
 
     public async void generate_export_preview () {
         if (list_store.get_n_items () > 0) {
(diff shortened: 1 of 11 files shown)


Alecaddd:
This is ready for a full review if you have time.
@giacomoalbe it seems that I'm not able to get the Artboard's name by using `artboard.name` as it returns null. Any idea why?
https://github.com/akiraux/Akira/pull/300/files#diff-ac282260754399476dee39f01da43b06R314

PR #303, opened 2020-03-16
Save file format

### LINKED ISSUE #106: Create the git backend for akira files
## Expected Behavior
We want to give the users a way to move back to previous versions of the file without any external tools.  

## Possible Solution
@albfan said we can use `libgit2` for this without having to require the use of git commands!

Alecaddd:
Implement the standard file operations to save, open, and save as, an Akira file.

## Steps to Test
* Create artboards and shapes
* Save the file
* Open the your OS file manager, browse to the newly saved file, view hidden files.
* Enter the `lock` folder and open the `content.json` file to see what was saved.

## Screenshots 
Coming soon...

- [x] Save file `.akira` with all the currently supported objects and properties.
- [x] Allow `Save As...` action to create a copy or override the existing one if selected.
- [x] Open a previously saved file.
- [x] Prevent opening the same file twice.

- Fixes #106


Alecaddd:
This is still a WIP, but all the basic wiring to open and save a file is done, as well as managing opened files attached to their windows.
Now it comes the big part, which is using the `JsonObject` classes to convert everything we have on the main canvas, writing into the json file, and creating a new git commit.

@Philip-Scott are you still around?
Would you be able to give me a little direction or suggestions to properly understand your initial implementation in order to hook it to how the canvas has evolved since then?

albfan:
@Alecaddd  you can reuse [load_item()](https://github.com/akiraux/Akira/commit/6ff494b824ee6dcf60895d036a6d211d8560d5ec#diff-45a33dc04853e50db5d915390ca49fd7R117) and [save_to_file()](https://github.com/akiraux/Akira/commit/6ff494b824ee6dcf60895d036a6d211d8560d5ec#diff-45a33dc04853e50db5d915390ca49fd7R215) from my branch `wip/albfan/serialization`

As long as it is isolated everything is mostly the same and we can refine that. Just use as the data when you need to [write_content_to_file()](https://github.com/akiraux/Akira/blob/save-file-format/src/FileFormat/AkiraFile.vala#L61)

Alecaddd:
@albfan awesome, thanks for the heads up :+1:

Alecaddd:
@albfan @giacomoalbe Hey guys, I'm asking for a review at this point as the initial implementation is completed.
Currently is not possible to open/load a file, and I left out the `Git` integration to keep this small and focused.

I'm having an issue with the `x` and `y` values which are never properly stored in the JSON file.
Do you any idea why? I tried to get the coordinates, used various methods, but those values always return `0.0`.

Follow the steps to save a file and look at the content inside, and let me know if you an figure out why those values are missing.

Alecaddd:
[test.akira.zip](https://github.com/akiraux/Akira/files/4367621/test.akira.zip)

This is a test file with an artboard and a bunch of shapes.
Unzip it and open it in Akira. It won't generate anything, but at least you'll be able to see the hidden folder with the content file.

albfan:
@Alecaddd I went down that road, let me check!

giacomoalbe:
> I'm having an issue with the `x` and `y` values which are never properly stored in the JSON file.
> Do you any idea why? I tried to get the coordinates, used various methods, but those values always return `0.0`.
> Most likely it doesn't really matter as I'm saving the `Cairo.Matrix` with the real coordinates, but it's curious that those values are always 0.

The reason lies in the `CanvasItem.position_item` function, here's the incriminated snipped of code:
```c
// Keep the item always in the origin
// move the entire coordinate system every time
translate (_x, _y);
```

We translate the entire coordinate system for easing the _affine transform_ methods and the conversion between different coordinate spaces. 

So the item coordinate system is translated but the object is always at the 0,0 coordinate (top left).

In order to get the absolute (canvas related) coordinates you can either convert from item space or you can transform the 0,0 point with the item matrix transorm. 

Hope this makes sense :) 

Btw great work, I need to find some time to test it properly!
Getting always closer :100:

Alecaddd:
Ah, that make sense.
Indeed, I don't think is gonna be a problem as I'm saving the Matrix transformation, so I can apply that on file load.

PR #307, opened 2020-03-21
Child items in Artboard layer

giacomoalbe:

This PR will add to Artboard layer the possibiliy to show its children inside its container.

## Steps to Test

* Create an Artboard on the canvas
* Create an item inside it
* In the LayersPanel the layer is created inside the Artboard

## Screenshots

akira-artboard-layer

* [x] Adding an item to an Artboard should add the layer to Artboard's layer
* [x] Deleting an item from the Artboard layer should delete it from the canvas
* [x] Deleting an item from the Canvas should delete it also from the Artboard's layer
* [x] Selecting an item from the Artboard layer should select it from the canvas
* [x] Selecting an item from the Canvas should select it also from the Artboard's layer
* [x] Hovering over an item from the Artboard layer should add the hover effect to it from the canvas
* [x] Hovering over an item from the Canvas should add the hover effect to it also from the Artboard's layer

### DIFF
--- a/data/css/artboard.css
+++ b/data/css/artboard.css
@@ -14,7 +14,7 @@
   border-bottom: 1px solid shade (@bg_color, 0.8);
 }
 
-.artboard.hover .layer-action {
+.artboard.hover > grid > grid > .layer-action {
   opacity: 0.5;
 }
 
--- a/src/Layouts/Partials/Artboard.vala
+++ b/src/Layouts/Partials/Artboard.vala
@@ -194,6 +194,14 @@ public class Akira.Layouts.Partials.Artboard : Gtk.ListBoxRow {
             get_style_context ().remove_class ("hover");
             return false;
         });
+
+        container.bind_model (model.items, item => {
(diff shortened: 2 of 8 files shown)


giacomoalbe:
Hi everyone! 

I leave this PR for review, it is very tiny at it actually does no more than what's here stated.
Thanks for the time, indeed :) 

In my opinion, we are almost there as it comes to the side panel. The only thing which is missing is the ability to show/hide items inside Artboard and to show/hide and lock Artbords, but this is related to the fact that property modification in artboard children are not triggering parent's redraw (the fill color does not work, either), so once this particular issue gets addressed show/hide should work again. 

I would not concentrate, though, on making the d&d stuff work again, since it is going to be very time consuming, we should need to document and formalize lots of different cases and the benefit would be very little, if not none. 

I would, instead, focus on more "urgent" and "useful" features, like alignment, file save, z-index, better affine transform and basic history management. 
What do you guys think?

PR #309, opened 2020-03-22
Restore Z-index functionality

giacomoalbe:

This PR will restore the z-index functionality and will improve the layer representation of items in the Canvas.

In particular, items added directly to the canvas will be separated from the list of artboards and will have a separate z-index stack.

For the sake of clarity, the former cluster is called "free items" and the second "artboards".

Each artboard, then, will share the same z-index, since there's not the possibility to stack artboard on top of each other nor to nest artboards inside others.

Inside the artboard, though, it should be possible to raise or lower the elements relative to the items already present.

## Steps to Test

- Draw a shape on the canvas
- Create an artboard
- Draw another shape in the canvas
- Create another artboard
- The artboard gets positioned right on top of the other. The same for the shape
- Items & artbord's layers does not belong to the same z stack anymore

## Screenshots

akira-artboard-layer

- [x] Creating an item outside of any Artboard should put that item in the "root" z-index stack
- [x] Moving items that are outside artboard should be relative to similar items, not to artboard
- [ ] Changing z-index of an item inside an artboard should put raise/lower that item relative to the other children of the artboard, not relative to other items in other artboards


src/Layouts/Partials/Artboard.vala
```diff
@@ -46,24 +46,24 @@ public class Akira.Layouts.Partials.Artboard : Gtk.ListBoxRow {
...
     public bool editing {
         get { return _editing; } set { _editing = value; }
     }
 
-    public Artboard (Akira.Window window, Akira.Models.LayerModel model) {
+    public Artboard (Akira.Window window, Akira.Lib.Models.CanvasArtboard artboard_model) {
         Object (
             window: window,
-            model: model
+            artboard_model: artboard_model
```
Alecaddd: Since this is the `model` of the `artboard` and we're only dealing with the `item_model`, it's better and more linear to call this simply `model`

src/Layouts/Partials/Layer.vala
```diff
@@ -85,21 +84,24 @@ public class Akira.Layouts.Partials.Layer : Gtk.ListBoxRow {
...
-        Akira.Models.LayerModel model,
-        Layouts.Partials.Artboard? artboard = null
+        Akira.Lib.Models.CanvasItem item_model,
+        Gtk.ListBox? list = null
     ) {
         Object (
             window: window,
-            model: model,
-            artboard: artboard
+            item_model: item_model
```
Alecaddd: Also here, leave it as `model`.

src/Layouts/Partials/LayersPanel.vala
```diff
@@ -236,12 +197,11 @@ public class Akira.Layouts.Partials.LayersPanel : Gtk.ListBox {
...
-                zebra_layer ((Akira.Layouts.Partials.Layer) row);
-                return;
-            }
+        items_list.@foreach (row => {
+            zebra_layer ((Akira.Layouts.Partials.Layer) row);
+        });
 
+        artboards_list.@foreach (row => {
             zebra_artboard ((Akira.Layouts.Partials.Artboard) row);
         });
```
Alecaddd: Here it would be nice to only reload the zebra of the list that was changed, and handling separated zebras per artboards, but that's something we can deal in future PRs.

src/Lib/Managers/ItemsManager.vala
```diff
@@ -51,33 +49,35 @@ public class Akira.Lib.Managers.ItemsManager : Object {
...
-    public Models.CanvasItem? insert_item (Gdk.EventButton event) {
+    public Models.CanvasItem? insert_item (double x, double y) {
         udpate_default_values ();
 
         Models.CanvasItem? new_item;
         Models.CanvasArtboard? artboard = null;
 
-        foreach (var _artboard in artboards) {
-            if (_artboard.is_inside (event.x, event.y)) {
+        root = window.main_window.main_canvas.canvas.get_root_item ();
```
Alecaddd: Why did you move this here? Isn't the root canvas always gonna be the same?
Moving it from the construct will set the variable every time a new item is inserted.

src/Lib/Managers/ItemsManager.vala
```diff
@@ -51,33 +49,35 @@ public class Akira.Lib.Managers.ItemsManager : Object {
         window.event_bus.change_item_z_index.connect (on_change_item_z_index);
     }
 
-    public Models.CanvasItem? insert_item (Gdk.EventButton event) {
+    public Models.CanvasItem? insert_item (double x, double y) {
```
Alecaddd: This change makes the loading of files way easier, great job!

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -160,6 +162,7 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
             canvas.window.event_bus.request_delete_item (item);
         }
 
+
```
Alecaddd: Remove this extra line :P

src/Models/BaseModel.vala
```diff
@@ -14,10 +14,10 @@
 * You should have received a copy of the GNU General Public License
 * along with Akira. If not, see <https://www.gnu.org/licenses/>.
 *
-* Authored by: Alessandro "alecaddd" Castellani <castellani.ale@gmail.com>
```
Alecaddd: Don't delete the previous contributors even if you renamed the file.

src/Layouts/Partials/Artboard.vala
giacomoalbe: Ok, I agree!

src/Layouts/Partials/LayersPanel.vala
giacomoalbe: Yes, I agree on updating only the artboard that has changed.

Btw, each artboard is treated as separated wrt zebra handling, so I don't get your question about treating them separately.

src/Lib/Managers/ItemsManager.vala
giacomoalbe: This is a weird problem. 

Since we moved the `items_manager` initialization at the Window level, the problem now is that we need to have the `canvas` properly set to get the `root` item. 

But the `canvas` itself, in turn, does need an instance of `items manager`, which leads to a recursive import problem: `canvas` need `items_manager` and at the same time `items_manager` needs the `canvas` at init time to get the root item. 

You can see this issue in the `window` construct code, I post it here:
```vala
    construct {
        accel_group = new Gtk.AccelGroup ();
        add_accel_group (accel_group);

        event_bus = new Akira.Services.EventBus ();
        items_manager = new Akira.Lib.Managers.ItemsManager (this);
        action_manager = new Akira.Services.ActionManager (app, this);
        file_manager = new Akira.FileFormat.FileManager (this);
        headerbar = new Akira.Layouts.HeaderBar (this);
        main_window = new Akira.Layouts.MainWindow (this);
        dialogs = new Akira.Utils.Dialogs (this);

        build_ui ();

        move (settings.pos_x, settings.pos_y);

        show_app ();

        main_window.main_canvas.canvas.focus_canvas ();
        event_bus.file_edited.connect (on_file_edited);
        event_bus.file_saved.connect (on_file_saved);
    }
```

## Possible solutions:
1. We could call a method in `items_manager` _after_ both `canvas` and `items_manager` have been created to save the root item of the `canvas` inside the `items_manager`. 
1. We can leave it as is now and simply request the `root` item at every item's creation.
 1. Request `root` item _only_ if it is null (so the first time we create an object). 

What do you think it's the best approach? Is the explanation clean enough? Does it make sense to you?

src/Lib/Managers/ItemsManager.vala
giacomoalbe: Yes, I did in order to make the testing phase easier (Idk how to create an EventButton). 

Good to know it'll help in other contexts as well!

src/Models/BaseModel.vala
giacomoalbe: I think I done it for accident, I don't recall how I deleted that line!

Btw, this file is not going to survive the next PR, since it is no more needed :) 

It's still there only because I didn't want to overload this PR, but since we changed the `ListModel`, both the `ItemModel` and the `BaseModel` are going to be dropped!

src/Lib/Managers/SelectedBoundManager.vala
giacomoalbe: Good catch!

src/Layouts/Partials/LayersPanel.vala
Alecaddd: Really?
The zebra coloring of my layers changes on all the artboards whenever I add a new layer.
Does it happen to you? As a general rule, the first layer inside an artboard should always have a white background.
I'll attach a screenshot.

src/Lib/Managers/ItemsManager.vala
Alecaddd: Method 3 seems the easiest and most linear, go for it and thanks for the explanation, that makes sense.

src/Layouts/Partials/LayersPanel.vala
giacomoalbe: Ok I noticed the issue you reported and I have to admit that I've never been able to spot it, good catch!

I guess it's because your eye is not used as mine at staring at the Layers panels and simply checking layers get created and destroyed after user interaction, that's why is always good to have someone else to look at your code!

I think we could delay this to a future PR, but if you do not agree I could try to sort this out.

src/Lib/Managers/ItemsManager.vala
giacomoalbe: Ok perfect!

src/Layouts/Partials/LayersPanel.vala
giacomoalbe: Never mind, it has been easier than what I expected, I fixed the issue with 2 line (one is an explanatory comment :))

src/Lib/Models/CanvasArtboard.vala
```diff
@@ -132,28 +132,28 @@ public class Akira.Lib.Models.CanvasArtboard : Goo.CanvasItemSimple, Goo.CanvasI
...
     }
 
     public uint get_items_length () {
-        return this.items.length ();
+        return this.items.get_n_items ();
     }
 
     public void remove_item (Models.CanvasItem item) {
-        items.remove (item);
+        items.remove_item (item);
```
Alecaddd, line 143: this is an `async` method, so it should be called as `items.remove_item.begin (item);`

PR #311, opened 2020-03-25
Load saved file

Alecaddd:
Implement the load of a previously saved file.

- [x] Open a previously saved file on a dedicated Window.
- [x] Load all the Artboards.
- [x] Load all the Items.
- [x] Restore the previous selection status.

### Fixes
- Closes #209


Alecaddd:
@giacomoalbe @albfan hey guys, I pretty much have this patch ready, except for an issue that I can't figure out why it's happening.

If a shape is inside an artboard, after it gets loaded, is not possible to move it. The selection bound moves, the resize works, and moving with the arrow keys work, but not with the mouse.
Can you give it a look?

To test it, create an arboard and a shape inside, save the file, close Akira and load that file.
Everything should be in place, but if you try to move that shape with the mouse, no redraw happens.

Alecaddd:
Another issue solved, but another issue found.
It seems that my `int_to_rgba` method is not correct and the colors are not properly loaded.
Can you guys give it a look?

Alecaddd:
I fixed it by saving the RGBA colors in a string format and creating a dedicated method to load them again.

Alecaddd:
Here's a file example you can use to test this patch.
[3-artboards.akira.zip](https://github.com/akiraux/Akira/files/4391506/3-artboards.akira.zip)

And this is how the app should look once loaded, including remembering the currently selected item.
Screenshot from 2020-03-26 21-36-12

giacomoalbe:
I checked out this PR locally and did some tests, but I haven't been able to make the Artboard children to work. I don't know whether it is a known bug or not, but everything seems to be working except for the Artboard children stuff. 

I have those critical errors for the Artboard's children layers:
```
(Akira:11529): Akira-CRITICAL **: 23:01:09.090: akira_models_layer_model_add_child_item: assertion 'self != NULL' failed

(Akira:11529): Akira-CRITICAL **: 23:01:09.090: akira_models_layer_model_get_child_item: assertion 'self != NULL' failed

(Akira:11529): Akira-CRITICAL **: 23:01:09.090: akira_models_layer_model_add_child_item: assertion 'self != NULL' failed

(Akira:11529): Akira-CRITICAL **: 23:01:09.090: akira_models_layer_model_get_child_item: assertion 'self != NULL' failed

(Akira:11529): Akira-CRITICAL **: 23:01:09.090: akira_models_layer_model_add_child_item: assertion 'self != NULL' failed

(Akira:11529): Akira-CRITICAL **: 23:01:09.090: akira_models_layer_model_get_child_item: assertion 'self != NULL' failed
```

And here's a screenshot of the Layers panel after the file import: 

image

As you can see, all the Artboard are expanded, so the children should be visible,  but they're not. 

Idk if this is due to the fact that the Layers panel and ItemsManager are not updated to the Master branch, but as of now they're not working as expected.

Alecaddd:
@giacomoalbe yes, this is normal as the issue is due to the incompatibility with the master branch.
After your latest PR landed, the list model for the free items is different and doesn't actually save anything.
I'll rebase this later this week.

Alecaddd:
@albfan @giacomoalbe this is ready for a full review.

I stumbled upon some quirks which I'm not totally sure why.
Specifically, I had to manually set the `item.bounds` of a loaded Artboard, otherwise those would be 0 when a new item is created and we need to check if the item is inside an Artboard.

src/Lib/Managers/ItemsManager.vala
```diff
@@ -371,4 +354,140 @@ public class Akira.Lib.Managers.ItemsManager : Object {
...
+
+            case "AkiraLibModelsCanvasText":
+                insert_type = Models.CanvasItemType.TEXT;
+                item = insert_item (pos_x, pos_y);
+                break;
+
+            case "AkiraLibModelsCanvasArtboard":
+                insert_type = Models.CanvasItemType.ARTBOARD;
+                item = insert_item (pos_x, pos_y);
+                break;
```
giacomoalbe, line 404: Here I can see the `CanvasImage` item's loading is not yet implemented.

I don't know if this is due to the fact the we have a separated PR for images and that would have caused merge issue, so I state this here as a reminder!

giacomoalbe:
Ok, for me this code is super ok! I just need to take a more detailed look at it (and maybe start tinkering with it) in order to properly understand each stage involved in the process, but overall I think it's very well written and concise!

src/Lib/Managers/ItemsManager.vala
Alecaddd: Yeah, I left the image stuff outside as I'm not yet saving images at all.
I will implement this in a separate PR.

PR #315, opened 2020-04-01
Limit build architectures

popey:
As we depend on the elementary PPA for their SDK packages, we should limit Akira to only build on architectures supported by those packages. That means no s390x mainframe builds and no powerpc builds for you! :D
(honestly, it would be hilarious to run elementary os and akira on an S390 mainframe) :D


Alecaddd:
I'm sure we will get some users requesting support for those architecture 😅

PR #319, opened 2020-04-06
WIP: Artboard Children Z Ordering

giacomoalbe:

This PR is to enable z ordering relative to Artboard

## Steps to Test

* Create an Artboard
* Add item to the Artboard
* Change z-index of the Item using known shortcuts or the z-index buttons in HeaderBar
* Check the Canvas status and Layer panel status

## Screenshots

akira-z-index-change-inside-artboard

* [x] Item inserted last in an Artboard are painted on top of the existing ones
* [x] Changing z-index of an Artboard's children changes Artboard's layer in LayerPanel
* [x] Changing z-index of an Artboard's children is reflected in Canvas Artboard items painting
* [x] HeaderBar button sensitivity works for Artboard childrens too

### DIFF
--- a/src/Models/ListModel.vala
+++ b/src/Models/ListModel.vala
@@ -35,8 +35,8 @@ public class Akira.Models.ListModel<Model> : GLib.Object, GLib.ListModel {
         return list.nth_data (position) as Object;
     }
 
-    public Model? nth_data (uint position) {
-        return list.nth_data (position);
+    public new Model get (uint index) {
+        return list.nth_data (index);
     }
 
     public Type get_item_type () {
--- a/src/Models/ListModel.vala
+++ b/src/Models/ListModel.vala
@@ -35,6 +35,10 @@ public class Akira.Models.ListModel<Model> : GLib.Object, GLib.ListModel {
         return list.nth_data (position) as Object;
     }
 
+    public Model? nth_data (uint position) {
+        return list.nth_data (position);
+    }
+
     public Type get_item_type () {
         return typeof (Model);
     }
--- a/src/Models/ListModel.vala
+++ b/src/Models/ListModel.vala
@@ -79,6 +79,13 @@ public class Akira.Models.ListModel<Model> : GLib.Object, GLib.ListModel {
         items_changed (position, 1, 0);
     }
 
+    public void swap_items (int source_index, int target_index) {
+        // Remove item at source position
+        var item_to_swap = remove_at (source_index);
+        // Insert item at target position
+        insert_at (target_index, item_to_swap);
+    }
+
     public void insert_at (int position, Model item) {
         list.insert (item, position);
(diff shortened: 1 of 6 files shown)


src/Models/ListModel.vala
```diff
@@ -35,6 +35,10 @@ public class Akira.Models.ListModel<Model> : GLib.Object, GLib.ListModel {
         return list.nth_data (position) as Object;
     }
 
+    public Model? nth_data (uint position) {
+        return list.nth_data (position);
+    }
```
Alecaddd, line 40: I noticed you tend to add a lot of these helper functions, but I'm not sure this is an optimal case.
Helper functions are necessary when dealing with repeated methods, conditions, or calculations, that need to be called in multiple locations throughout the application.

This `nth_data()` method basically only removes the necessity of calling the list object.
From `items.list.nth_data (items_length - 1 - i);` to `items.nth_data (items_length - 1 - i);`.
I don't see any benefit to this, and it can also be quite confusing for other devs or even ourselves in the future since this method is named exactly like the original method.

We should try to avoid adding helper functions without any login within.
What do you think?

giacomoalbe: Yes, I tend to create helper methods, when they're needed. 

Here I totally agree with you that this thing isn't very useful. 

But it tries to solve a lot of quirks that happended to accumulate on this particular spot eheh. 

First, the list inside the ListModel is a private member and I think for a good reason, so I just don't want to expose it and use directly. 

Second, the `get_item` method in ListModel is required by the GLib.ListModel interface to return an GLib.Object and I didn't want to cast the item I get from the ListModel here.

The third issue is that it seems that Vala does not expose any bracket indexing API. I just wanted to the the i-th element of a list by using the [] operator, but, at a first glance, I've been able to find the way to overload the indexing operator in ListModel to provide this way of getting the items inside with the proper type. 

So I created this tiny wrapper around the list `nth_data`, just to accomplish this super simple task that Vala is not providing. 

I think I'll dig deeper trying to find a way to do this using only Vala, otherwise I'll simply use the `get_item` with casting!

giacomoalbe: It has been as simple as that, I've just added these lines to the `ListModel` class:

```vala
    public Model get(uint index) {
        return list.nth_data (index);
    }
```

You simply need to add the `get` (and the corresponding `set` method) in order to enable the bracket indexing of the item itself! 

Super cool! 

I found it here, I think you might be interested: https://naaando.gitbooks.io/the-vala-tutorial/en/5-advanced-features/methods-with-syntax-support.html

giacomoalbe:
Ok, I think I'm done with this! It has been truly rewarding to see that I could change the z-index ordering, I'm not able to see why :laughing: 

This is ready for a full review, so have a look at it and please share some feedback (or just approve and merge it)!

PR #320, opened 2020-04-10
Fix segfault when deleting an artboard

Alecaddd:
It seems we were trying to remove the same item multiple times as the code wasn't preventing an extra `delete()` when dealing with Artboards.
This should fix it.


giacomoalbe:
Ok I tested it locally and it worked as expected! I was facing the same issue once, when I was recording a gif, I said to myself "I will fix this later" but you've been quicker than me :)

PR #323, opened 2020-04-14
Artbord improvements

### LINKED ISSUE #325: Broken paint method when items are partially outside an artboard
## Expected Behavior
If an item is partially outside an Artboard, that portion should be invisible.

## Current Behavior
The item is fully visible, and a filling trail is left behind if the Artboard is moved.

## Possible Solution
I'm not sure if this could be solved by forcing a call to the `simple_paint` method of the main Canvas when an artboard is moved.

## Screenshots of the Problem
paint-issue

Alecaddd:
A round of improvements and fixes to make the Artboard implementation even more awesome.

## Screenshots 
items-in-artboards

- [x] Artboards shouldn't rotate.
- [x] Fix accidental dragging of selected elements when clicking on a locked item.
- [x] Dropping an item outside its Artboard should remove it from it.
- [x] Dropping an item inside an Artboard should add it to it.
- [x] Moving an item from an Artboard to another should change its parent.
- [x] Adapt drop shadow based on Light/Dark mode.
- [x] Scale the label based on the current Canvas scale.

### Fixes

Fixes #325


Alecaddd:
@giacomoalbe I've almost completed this PR, but I'm stumbling upon a problem when trying to move a free item into an artboard.
It seems as the item gets duplicated, like it gets connected to the artboard twice.

A related issue is that if you try to then move that item outside, it gets pushed to 10k coordinates.
So strange.

Would you be able to take a look at it and maybe find the issue with fresh eyes?

Alecaddd:
I think I fixed the issue of the duplicate item when adding a free item to an artboard, but the wrong positioning of an item getting added and then removed from an artboard persists.

src/Lib/Canvas.vala
```diff
@@ -387,4 +390,92 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
+                var new_artboard = artboard.dropped_inside (item) ? artboard : null;
+                change_artboard (item, new_artboard);
+            }
+        }
+    }
+
+    /**
+     * Add or remove an item from an artboard.
+     */
+    public void change_artboard (Models.CanvasItem item, Models.CanvasArtboard? new_artboard) {
```
giacomoalbe, line 424: Maybe we can move this logic out to the `ItemsManager` or `SelectedBoundManager` just to let the Canvas to deal _only_ with Canvas issues and business.

What do you think?

Alecaddd:
I left out the ability to control the background color of the canvas, and also a method to change the label colour based on light/dark theme as I will deal with those in a separated PR.

giacomoalbe:
> This is ready for a full review.
> I left out the ability to control the background color of the canvas, and also a method to change the label colour based on light/dark theme as I will deal with those in a separated PR.

Ok, 100% agree on the "split the PR mood". 
I'd like to add just another _quirk_ that we need to face related to Artboards: the name label is always 12pt tall, regardless of the zoom level. Maybe it would be nice to make it relative to the zoom, so that Artboard's name are always visible and sized the same at every zoom level.

isneezy:
@Alecaddd what do you think about removing the right and bottom borders from the artboard since the background of the canvas creates a nice contrast with the artboard. Sometimes it is tricky to be sure that the shape has the correct size specially when you zoom in and the shape as a similar color with the artboard borders. May be the way to go is not really remove those borders but play around with different options like trying to have the border on all sides. or try with shadows I don't know. Just want to know if you fill the same / what do you think about it.

akira3
Akira1
Akira2

Alecaddd:
@giacomoalbe indeed, I will.add that in the next PR.

@isneezy the original idea was to have a light drop shadow to improve the contrast, but technical limitations don't make it easy to implement.
I'll see what I can do in the next PR, and if not possible, I'll remove it.

Alecaddd:
> I'd like to add just another quirk that we need to face related to Artboards: the name label is always 12pt tall, regardless of the zoom level. Maybe it would be nice to make it relative to the zoom, so that Artboard's name are always visible and sized the same at every zoom level.

I fixed it :smile:

Alecaddd:
I decided to remove the emulated drop shadow and increase a little bit the contrast with the container.
Like states above, having a hard rectangle as a drop shadow doesn't work very well when aligning content or looking at zoomed in previews.
And since Cairo doesn't really support blur or drop shadow without some pretty hacky and resource intensive operations, I think it's better to remove it for now and deal with that later.

`Goo.CanvasItem` apparently support a `CanvasStyle` property which allows to define CSS attributes. I tried to add a drop shadow but it doesn't work.
I created an issue on GitLab, let's see if we can overcome this limitation: https://gitlab.gnome.org/GNOME/goocanvas/-/issues/9

Alecaddd:
I'm merging this as it's a blocker for the files drag and drop, and also because it brings a lot of needed fixes.

PR #332, opened 2020-04-19
Create Release Dialog

### LINKED ISSUE #329: Create a fancy but scary warning dialog for the Alpha release
The alpha release is upon us, which means probably many users will see this app popping out in their app stores, and upon using it the first will most likely wonder why so many basic features are missing.

We should implement a warning dialog that triggers automatically when the app is opened the first time and warns the user that this is an alpha and if you use it too much a black hole might open i your bathroom...or something like that.

Nonetheless, if should be a nice and stylish dialog, with quick links to the repo, donations, and issue tracker.

Alecaddd:
Implement a Release Dialog that gets triggered at startup automatically if the version of the App doesn't match the latest stored version number in the user's GSettings.
We will initially use this to warn the user about the alpha status of the software.

## Screenshots 
Screenshot from 2020-04-19 19-17-08

- [x] Trigger a fancy dialog with a cute header.
- [x] Show a scary warning message.
- [x] List the features from the changelog (is it even possible?)
- [x] Don't show it again after the first time.

- Fixes #329


Alecaddd:
It's a bit jarring and not super cute, but it's on purpose in order to catch the user attention.
It looks way better without the warning message.

PR #338, opened 2020-04-27
Layers Panel Drag & Drop

Alecaddd:
Enable the usual Gtk Drag and Drop for Layers and Artboards in the Layer panel.

## Steps to Test
1. Create a bunch of layers and artboards.
2. Reorder them with drag and drop.
3. Be sure the z-index order was properly updated.

## Screenshots 
dnd-layers

- [x] Enable Layer drag motion and don't interfere with click event box.
- [x] Show drop indicator on drag motion for layers, artboards, search field, and empty panel.
- [x] Reorder z-index on drop.
- [x] Change artboard accordingly on drop.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .vscode
 src/config.vala
 build/
+builddir/
 _build/
 .flatpak-builder/
 *~
--- a/data/css/artboard.css
+++ b/data/css/artboard.css
@@ -34,10 +34,6 @@
     color: @fg_color;
 }
 
-.highlight-drop-area .artboard:last-child {
-    box-shadow: inset 0 -2px 0 shade (@bg_color, 0.5);
-}
-
 .artboard-container {
     background-color: @bg_color;
 }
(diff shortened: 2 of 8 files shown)


Alecaddd:
This is ready for a review, even if it doesn't include everything.
I haven't implemented yet the drop on the Search area or the empty layers panel, as well as the full drag and drop of artboards.
I'll do those in later PRs to not make this one too big and hard to review.

Alecaddd:
>     * z-order change does not work on free items, when using d&d

Yep, the free items ordering is pretty messed up also on master. I'll deal with that in a dedicated PR.

>     * When dragging an item, the zebra coloring is not updated. I don't know whether this is a bug or intended, but it doesn't feel completely ok to me

Good point. I think the only solution is to not hide the dragged layer, instead of recalcualting the zebra coloring.

For all the other issues you pointed out, I wrote the comment above:
> I haven't implemented yet the drop on the Search area or the empty layers panel, as well as the full drag and drop of artboards.
> I'll do those in later PRs to not make this one too big and hard to review.

PR #353, opened 2020-06-30
Disable flipping for CanvasArtboard (fixes #342)

### LINKED ISSUE #342: ArtBoards shouldn't have flip options
## Expected Behavior
Flip options should be disabled for ArtBoards

## Current Behavior
User can flip ArtBoards vertically and horizontally 

1.Launch Akira
2.Create an ArtBoard
3.Click on flip buttons

## Screenshots of the Problem
Artboards flip

- Akira Version: 0.0.1
- OS Info: Manjaro 20

TimHal:

Disabling hflip/vflip button if a Lib.Models.CanvasArtboard is selected.

- Fixes #342

### DIFF
--- a/src/Layouts/Partials/TransformPanel.vala
+++ b/src/Layouts/Partials/TransformPanel.vala
@@ -189,6 +189,14 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
         }
 
         selected_item = selected_items.nth_data (0);
+
+        if (selected_item is Lib.Models.CanvasArtboard) {
+            hflip_button.hide ();
+            vflip_button.hide ();
+        } else {
+            hflip_button.show ();
+            vflip_button.show ();
+        }
     }
 
     private void disconnect_previous_item () {
--- a/src/Layouts/Partials/TransformPanel.vala
+++ b/src/Layouts/Partials/TransformPanel.vala
@@ -70,8 +70,8 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
             height.enabled = has_item;
             width.enabled = has_item;
             rotation.enabled = has_item && !(_selected_item is Lib.Models.CanvasArtboard);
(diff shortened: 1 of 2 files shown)


src/Layouts/Partials/TransformPanel.vala
```diff
@@ -189,6 +189,14 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
...
 
         selected_item = selected_items.nth_data (0);
+
+        if (selected_item is Lib.Models.CanvasArtboard) {
+            hflip_button.hide ();
+            vflip_button.hide ();
+        } else {
+            hflip_button.show ();
+            vflip_button.show ();
+        }
```
Alecaddd, line 199: Thanks for the contribution.
I prefer to change the button sensitivity instead of hiding them.
Already the Rotation field stays insensitive, so we keep consistency and we prevent sudden changes in the UI by hiding items that we never hide.

Alecaddd:
This PR should also take care of stopping the flip command if an Artboard is selected, since we're handling that with accelerators as well.

src/Layouts/Partials/TransformPanel.vala
TimHal: Hi, thanks for the quick reply!
I see, I will try to change the sensitivity instead then. 

> This PR should also take care of stopping the flip command if an Artboard is selected

Sorry about that, I have not even realized there was a shortcut for it

Alecaddd: No need to apologize, it's a big code base and the simple fact that you're willing to contribute to it is already amazing.
I'm here to help you figure out all the nits of the source code.
Thank you so much :smile:

TimHal: Awesome, thank you a lot :) I have made some changes again, hope the PR will be updated soon.
By the way, do you have a mailing list/IRC/discord for this project?

PR #377, opened 2020-07-22
Fix items coordinates

### LINKED ISSUE #372: Items position doesn't account for rotation when inside Artboard
## Expected Behavior
The X and Y position of an item should reflect the real position of its bounds.

## Current Behavior
The position of an item inside an artboard is incorrect and doesn't account for its rotation.
This issue also causes rotated items to be wrongly positioned when dropped on top of artboards. 

## Screenshots of the Problem
rotate-position

Alecaddd:

This PR takes care of improving the ongoing rewrite of the Affine Transform, focusing mostly on scale events and accurate coordinates.

## Screenshots 

TBD

## This PR fixes/implements the following bugs/features

- [x] Implement private methods for coordinate constraints during scale event.
- [x] Make the item properly scale if the size ratio is locked.


Alecaddd:
I'm merging this since it improves a lot of things and I don't want to make this PR grow too much.

PR #379, opened 2020-08-01
Fix item bounds

### LINKED ISSUE #372: Items position doesn't account for rotation when inside Artboard

Alecaddd:
Implement a new `GhostBoundsManager` object to keep track of global coordinates, no matter if the item is inside or outside an artboard, or if it's part of a group.
This also allows us to use this "ghost" item for future rulers overlay.

## Steps to Test
- Create an item inside an artboard.
- Rotate the item and see the X & Y coordinates updating properly.
- Hold <kbd>ALT</kbd> when an item is selected to reveal its ghost.

## Screenshots 
item-bounds

- [x] Create a ghost item to keep track of real world item position.
- [x] Use the relative coordinates only to translate items inside artboards.
- [x] Improve the `get_position` and `set_position` of the `AffineTransform` object.
~Properly position rotated items when dragged inside and outside artboards.~

### Fixes

- Fixes #372

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -239,7 +239,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                         return true;
                     }
 
-                    // Item has been selected
+                    // Item has been selected.
                     selected_bound_manager.add_item_to_selection (clicked_item as Models.CanvasItem);
                 }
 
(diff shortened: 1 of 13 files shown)


Alecaddd:
> Properly position rotated items when dragged inside and outside artboards.

I think I'm gonna do this in a dedicated PR as I don't want to mess things up with what currently works.
Also, this PR is already pretty big on its own.

PR #383, opened 2020-08-02
Fix rotation

albfan:
- correct atan parameters
- reference points to item space
- invert calculated rotation (this can be simplified in atan calculation)

## Steps to Test

Rotate now behaves correctly

## Screenshots 

akira-rotation

- Rotate again after mouse release does an unexpected jump (corner case I guess)

### DIFF
--- a/src/Lib/Managers/SelectedBoundManager.vala
+++ b/src/Lib/Managers/SelectedBoundManager.vala
@@ -103,9 +103,8 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
 
             case Managers.NobManager.Nob.ROTATE:
                 Utils.AffineTransform.rotate_from_event (
-                    selected_item,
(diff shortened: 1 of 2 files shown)


Alecaddd:
Another thing missing from this PR is the rotation for items inside an artboard

Alecaddd:
> Rotate again after mouse release does an unexpected jump (corner case I guess)

I found the solution for that.
Since at every click we update the initial event X and Y, the radians difference was always 0 at first rotation.
We need to pass those initial coordinates as reference and account for the current item rotation.

I'll tweak the code an merge this.

Alecaddd:
I partially fixed the rotation of items inside artbaords.
I'm pushing on this PR because it's easier.

The fixed 15deg rotation doesn't seem to work, after I fix that, I'll merge.

PR #389, opened 2020-08-05
Translation: Update zh_CN.po

whriedplanck:

Update the Chinese Simplified translation.
## Steps to Test
 
Modified the zh_CN.po, and copied it to `po` folder, then followed the Compile Guide in the README.md to compile and run it. 
## Screenshots 

 - Interface")
^ Before translating (1) - Interface
 - About dialog")
^ Before translating (2) - About Dialog
 - Interface")
^ After translating (1) - Interface
 - About dialog")
^ After translating (2) - About Dialog

There are two strings in the file that I'm unable to accurately translated.
 

None

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -1,4 +1,666 @@
 msgid ""
 msgstr ""
+"Project-Id-Version: Akira 0.0.11 alpha\n"
+"POT-Creation-Date: 2020-08-05 20:25+0800\n"
+"PO-Revision-Date: \n"
+"Last-Translator: WhiredPlanck <fungdaat31@outlook.com>\n"
+"Language-Team: \n"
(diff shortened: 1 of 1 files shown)


albfan:
We don't have a i18n team so we mostly will trust you in these.

Just squash the changes and we should be good to go

whriedplanck:
> We don't have a i18n team so we mostly will trust you in these.
> 
> Just squash the changes and we should be good to go

Thank you for your approval. : D

Because Akira is still under development, many features are unstable and some buttons are unable to use/press, I can confirm the translation for some strings, but not for others. 

But I believe if users are able to select their native language to use the software, it may attract more people to pay attention to Akira, and even make an important contribution to it. So although Akira is not ready for production, I still want to help in translation. 

Thanks again!

PR #390, opened 2020-08-07
Fix items selection in artboards

Alecaddd:
We're currently leveraging the `simple_is_item_at` to hover and select items inside an Artboard, but this approach doesn't fully work as it seems we can't properly account for the filled area of an item, causing the selection based on its bounding box.

This PR implements the approach of using the duplicated ghost item, currently used to properly represent the X & Y coordinates of the item's bounding box.
The ghost item is a perfect duplicate which lives as invisible item on the canvas and it's not selectable.
We can use that for items inside an Artboard since that object retains the correct cairo fill, therefore returning the proper pointer status.

## Screenshots 
Before:
before

After:
after

## This PR implements
- [x] Update the goocanvas VAPI to match the updated [VAPI](https://gitlab.gnome.org/GNOME/vala/-/commit/7b45f83365e969141de8a2f7e86b47079156c9eb)
- [x] Implement coordinate getters for the `bounds_manager` 
- [x] Implement the `duplicate_item ()` method
- [x] Use the duplicated item for artboard's child selection
- [x] Remove the `simple_is_item_at` method from the CanvasItem
- [x] Cleanup unused z-index methods

We need to evaluate the performance of this approach since we're basically creating 2 extra items for each item we generate:
- The duplicate item
- The bounding ghost item

Both extra items are not rendered by default and get updated whenever the user interacts with the original item.
This might bite us in the ass once we implement multiselect or for large projects.

**Z-INDEX**
Another issue with this approach is that the z-index of items inside artboards don't properly reflect the duplicate ghost items.
We could solve this by mirroring the sorting of those items with their duplicates.

### DIFF
--- a/src/Layouts/HeaderBar.vala
+++ b/src/Layouts/HeaderBar.vala
@@ -85,24 +85,23 @@ public class Akira.Layouts.HeaderBar : Gtk.HeaderBar {
 
         move_up = new Akira.Partials.HeaderBarButton (window, "selection-raise",
             _("Up"), {"<Ctrl>Up"}, "single");
-        move_up.button.clicked.connect (() => {
-            window.event_bus.change_z_selected (true, false);
-        });
+        move_up.button.action_name = Akira.Services.ActionManager.ACTION_PREFIX
+            + Akira.Services.ActionManager.ACTION_MOVE_UP;
+
         move_down = new Akira.Partials.HeaderBarButton (window, "selection-lower",
             _("Down"), {"<Ctrl>Down"}, "single");
-        move_down.button.clicked.connect (() => {
-            window.event_bus.change_z_selected (false, false);
-        });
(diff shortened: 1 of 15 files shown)


Alecaddd:
This is broken on purpose to offer a simple reproducer for the Vala maintainers.
https://gitlab.gnome.org/GNOME/vala/-/issues/1057

Alecaddd:
I'm not sure about this as it creates a lot of problem with the z-index, and also I introduced some random segfaults probably due to the duplication method.

Alecaddd:
So, this PR got a little bit too big and it comes with a lot of small fixes that should be pushed on dedicated PRs.
I'm gonna split it and push the minor things first, then delete this and rework it.

Alecaddd:
Closing this as it's now outdated due to other merged PRs, and the proposed solution doesn't work well.

PR #396, opened 2020-08-11
Use map iterator to handle accelerators

### LINKED ISSUE #392: Avoid potential crash inside set_accels_for_action()
Something fishy is going on inside `set_accels_for_action()` when its second argument is passed from the `Gee.MultiMap<string, string>`: apparently the `string[] accels` gets garbled somehow so the program crashes inside the `strlen()` call shortly after.  I've observed this on FreeBSD, and it happens regardless of the compiler being used (Clang 10.0.1 or GCC 7.5.0).

Changing the second argument to `{ typing_accelerators[action].to_array ()[0] }` gets rid of the crash.  Another option would be to switch to map iterator, which also is a little more efficient and easy on eyes, e.g.:
```vala
  var iter = action_accelerators.map_iterator ();
  while (iter.next ()) {
     app.set_accels_for_action (ACTION_PREFIX + iter.get_key (), { iter.get_value () });
  }
```
See also: https://github.com/Alecaddd/sequeler/issues/96

Alecaddd:
Use `map_iterator ()` to handle setting accelerators for actions.

- Fixes #392

### DIFF
--- a/src/Services/ActionManager.vala
+++ b/src/Services/ActionManager.vala
@@ -153,8 +153,9 @@ public class Akira.Services.ActionManager : Object {
         actions.add_action_entries (ACTION_ENTRIES, this);
         window.insert_action_group ("win", actions);
 
-        foreach (var action in action_accelerators.get_keys ()) {
-            app.set_accels_for_action (ACTION_PREFIX + action, action_accelerators[action].to_array ());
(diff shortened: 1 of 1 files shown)


Alecaddd:
Merging this based on the reporter feedback, which suggested the solution to this random segfault.

PR #402, opened 2020-08-13
Add Spanish translation

fitojb:
Added a Spanish translation. I used Poedit to extract strings from the source code, as there is no POT available. Looks good after building, but some texts are invisible in Adwaita (such as in the Export dialog) (the elementary theme isn’t available on Ubuntu).

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -1,5 +1,666 @@
 msgid ""
 msgstr ""
+"Project-Id-Version: \n"
+"POT-Creation-Date: 2020-08-12 22:06-0500\n"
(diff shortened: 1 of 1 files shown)


Alecaddd:
Thanks for the translation.
We haven't created dedicated pot files, nor connected the repo to a more automated system like weblate or gitlocalize, because being in alpha we will constantly change strings and structure, and we didn't want to open it to translators for then force them to push updates once a week.

I'll merge this anyway, but be aware that things might break very frequently in your language, sorry.

PR #409, opened 2020-08-16
Copy/paste of shapes

### LINKED ISSUE #358: Beta Release Wishlist
## Setting up expectations for a Beta release

Here's a list of must-have features that will be implemented during the ongoing alpha cycle.
This list is a **Work in Progress** and it can change based on availability and technical challenges.

### General UI
- [ ] Create missing symbolic icons.
- [ ] Allow swapping the order of the Options Panel and Layers Panel.

### Canvas
- [ ] Shapes should snap to Artboard's edges.
- [ ] Click and drag on an empty canvas area to select multiple shapes.
- [ ] Zoom should always use currently selected items as center point.
- [ ] Selecting an item from the Layers Panel should move the Canvas to visually center it.

### Shapes
- [ ] Copy/Paste of shapes.
- [ ] Duplicate shape.
- [ ] Implement the Line shape.
- [ ] Implement the Curve/Pen tool.
- [ ] Implement a Global Color Picker.

...

nishp77:

It's still a feature to be implement as of yet

## Steps to Test
 
Just try to use the normal copy/ paste controls for the feature to used

## Screenshots 

Initially made a rect, then a circle. Selected the rect, then copied it

image

After that, use the paste action, 

image

- Fixes #358 (Beta wish list)

### DIFF
--- a/build-aux/flatpak/com.github.akiraux.akira.json
+++ b/build-aux/flatpak/com.github.akiraux.akira.json
@@ -19,10 +19,7 @@
       "--share=ipc",
       "--socket=fallback-x11",
       "--socket=wayland",
-      "--talk-name=org.gtk.vfs.*",
-      /* dconf */
-      "--filesystem=xdg-run/dconf", "--filesystem=~/.config/dconf:ro",
-      "--talk-name=ca.desrt.dconf", "--env=DCONF_USER_CONFIG_DIR=.config/dconf"
+      "--talk-name=org.gtk.vfs.*"
    ],
    "modules": [
       {
--- a/src/Services/ActionManager.vala
+++ b/src/Services/ActionManager.vala
@@ -64,6 +64,8 @@ public class Akira.Services.ActionManager : Object {
     public const string ACTION_FLIP_V = "action_flip_v";
     public const string ACTION_ESCAPE = "action_escape";
     public const string ACTION_SHORTCUTS = "action_shortcuts";
+    public const string ACTION_COPY = "action_copy";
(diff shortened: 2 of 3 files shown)


Alecaddd:
Hi, thanks for contributing to Akira.

A couple of general guidelines to follow:
- If your PR is a still a work in progress, open it as a Draft PR. I updated it already so no problem for now.
- Don't write that a PR fixes an issue if that issue doesn't have a 1-to-1 relationship with the PR. In this case, you're stating that this PR fixes the Beta Whishlist, but that's incorrect.

Unfortunately, this is not the right approach.

You're assuming that the items to paste are those currently selected. That's not always the case.
You're limiting the copy/paste to 1 item at a time, but that's not scalable and should account for all selected items, therefore, this PR should be postponed after the multiselect is implemented.
With this approach you will not properly inherit all the attributes of the duplicated item (colors, borders, rotation, etc.).
You're currently reaching properties and methods by going through the parent->child hierarchy of all the classes, which is cumbersome and not scalable. You should use events for that.

Thanks again for your contribution, but I recommend to spend a little bit more time in trying to get comfortable with Vala before submitting a PR.
Also, Akira has a fairly complex code base, so you should get yourself acquainted with it before submitting a PR to be sure you're following the same paradigms and architecture.

Alecaddd:
Another couple of fundamental things to consider:
- When copying, the cached elements should belong to the App and not the Window, so users can copy a shape from a window and paste it in another.
- When pasting, we need to detect what the user has saved in the clipboard and handle those cases (images, svg icons, text, etc.)

PR #411, opened 2020-08-17
Artboard code cleanup

Alecaddd:
Not fixing any specific issues here but rather improving the code in some areas and preventing unnecessary calls.

- Call the `GhostBoundsManager.update()` method only when necessary and in specific sections to account for resize and movement.
- Remove the `MainWindow` size allocation since it always returns 1px as it hasn't been created yet.
- Remove the Artboard `bounds` custom attributes since those are never used.
- Set the `Cairo.Matrix` transform to items inside Artboards only when they're converted into free items.
- Account for the scale of the Canvas for the padding bottom of Artboards' labels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,8 @@ ninja && sudo ninja install
 
 ## 🤔 Questions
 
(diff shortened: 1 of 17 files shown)


Alecaddd:
Merging this as it doesn't change or introduce any feature but simply optimizes the code a little bit.

PR #417, opened 2020-08-22
Solved Zoom to mouse issue/request #406

### LINKED ISSUE #406: Feature: Zoom should zoom to cursor, not center of view.
## Expected Behavior
If I zoom via the mouse wheel, it should zoom to the cursor. So if an item is in the upper left corner of my screen and I place the mouse cursor there and zoom (ctrl+wheel) the cursor should stay on the item and the view port should zoom in on the item (The mechanics themselves have nothing to do with the item, it is just easier to imagine this way)

This is standard behavior in many applications, like inkscape and figma (though, to my annoyance, not in balsamiq)

## Current Behavior
Currently, the zoom is on center. This means the view is made "larger" while its center stays in place

## Possible Solution
Implement zoom to cursor. This means that the viewport needs to shift depending on mouse position while zooming. 

## Context
When you zoom in, you usually want to zoom in on something. The only way to do this right now is to pan the object of interest into the center of the viewport and then zoom. This is cumbersome. Zoom-to-cursor would just mean you place the cursor on what you are interested in and zoom in. 
(this is also coherent with pinch-to-zoom; while zoom to cursor pans the viewport in a way that your cursor stays on its position within the canvas/document, while pinch-to-zoom pans in a way so that your fingers or the point between them stays on the same position in the canvas/document)

## Zoom-To-Cursor in Inkscape

Peek 2020-08-13 22-03

- Akira Version: v0.0.12 - alpha
- OS Info: Ubuntu 20.04

tenzens:
zoom demo

It solves the issue of only zooming to the center of the view. Now you can zoom to and from mouse position.

There might be an easier way to solve this but this is what I was able to come up with. Also this is my first commit to any project at all so please keep that in mind when giving me feedback :)

- Fixes #406

### DIFF
--- a/src/Layouts/MainCanvas.vala
+++ b/src/Layouts/MainCanvas.vala
@@ -115,8 +115,36 @@ public class Akira.Layouts.MainCanvas : Gtk.Grid {
         if (delta_y < -SCROLL_DISTANCE) {
             // Scroll UP
             if (is_ctrl) {
+                double old_zoom = canvas.get_scale ();
+                //The regular zoom mode shifts the visible viewing area 
+                //to center itself (it already has one translation applied) 
+                //so you cannot just move the viewing area by the distance
+                //of the current mouse location and the new mouse location
+
+                //If you want to zoom to your mouse you need to find the
+                //difference between the distances of the current mouse location
+                //in the current view scale to the left view border and the new
+                //mouse location that has the new canvas scale applied to the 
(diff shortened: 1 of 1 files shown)


albfan:
We can review and suggest refactors, but would be interesting to upload an screencast of what current implementation does.

tenzens:
Thanks for the reply. I updated my text with a gif. If required I can also upload a full res video via google drive.

PR #432, opened 2020-08-30
Implement a global Color Picker

isneezy:

Implement a global Color Picker ported from this awesome little tool https://github.com/RonnyDo/ColorPicker

## Steps to Test
 
1. Draw a and select a shape
2. Click on the first delete icon from the left to right of the Fills panel
3. Pick a color

## Screenshots 

Screen record from 2020-08-31 11 24 03

- When used on multi-monitor setup the application window should be on the primary monitor otherwise it won't pick colors from other monitors
- Wayland compatibility, mentioned by @bilelmoussaoui [here](https://github.com/akiraux/Akira/pull/432#issuecomment-683389423)

- Related to #358

### DIFF
--- a/src/Layouts/Partials/BorderItem.vala
+++ b/src/Layouts/Partials/BorderItem.vala
@@ -224,9 +224,7 @@ public class Akira.Layouts.Partials.BorderItem : Gtk.Grid {
         eyedropper.show_all ();
 
         eyedropper.picked.connect ((picked_color) => {
-            alpha = ((int)(picked_color.alpha * 255));
             color_chooser_widget.set_rgba (picked_color);
-            on_color_changed ();
             eyedropper.close ();
         });
 
--- a/src/Layouts/Partials/BorderItem.vala
+++ b/src/Layouts/Partials/BorderItem.vala
@@ -17,12 +17,14 @@
  * along with Akira. If not, see <https://www.gnu.org/licenses/>.
  *
  * Authored by: Alessandro "alecaddd" Castellani <castellani.ale@gmail.com>
+ * Authored by: Ivan "isneezy" Vilanculo <ivilanculo@gmail.com>
  */
 
 public class Akira.Layouts.Partials.BorderItem : Gtk.Grid {
     public weak Akira.Window window { get; construct; }
 
     private Gtk.Grid color_chooser;
+    private Gtk.Button eyedropper_button;
     private Gtk.Button hidden_button;
     private Gtk.Button delete_button;
     private Gtk.Image hidden_button_icon;
@@ -31,6 +33,7 @@ public class Akira.Layouts.Partials.BorderItem : Gtk.Grid {
     public Akira.Partials.ColorField color_container;
     private Gtk.Popover color_popover;
     private Gtk.Grid color_picker;
+    private Akira.Utils.ColorPicker eyedropper;
     private Gtk.ColorChooserWidget color_chooser_widget;
 
     public Akira.Models.BordersItemModel model { get; construct; }
@@ -164,6 +167,15 @@ public class Akira.Layouts.Partials.BorderItem : Gtk.Grid {
         color_chooser.attach (color_container, 1, 0, 1, 1);
         color_chooser.attach (tickness_container, 2, 0, 1, 1);
 
+        eyedropper_button = new Gtk.Button ();
+        eyedropper_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
+        eyedropper_button.get_style_context ().add_class ("button-rounded");
+        eyedropper_button.can_focus = false;
+        eyedropper_button.valign = Gtk.Align.CENTER;
+        eyedropper_button.set_tooltip_text (_("Pick color"));
+        eyedropper_button.add (new Gtk.Image.from_icon_name ("preferences-color-symbolic",
+            Gtk.IconSize.SMALL_TOOLBAR));
+
         hidden_button = new Gtk.Button ();
         hidden_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
         hidden_button.get_style_context ().add_class ("button-rounded");
@@ -189,18 +201,36 @@ public class Akira.Layouts.Partials.BorderItem : Gtk.Grid {
         color_popover.add (color_picker);
 
         attach (color_chooser, 0, 0, 1, 1);
-        attach (hidden_button, 1, 0, 1, 1);
-        attach (delete_button, 2, 0, 1, 1);
+        attach (eyedropper_button, 1, 0, 1, 1);
+        attach (hidden_button, 2, 0, 1, 1);
+        attach (delete_button, 3, 0, 1, 1);
 
         set_color_chooser_color ();
         set_button_color ();
     }
 
     private void create_event_bindings () {
+        eyedropper_button.clicked.connect (on_eyedropper_click);
         delete_button.clicked.connect (on_delete_item);
         hidden_button.clicked.connect (toggle_visibility);
         model.notify.connect (on_model_changed);
         color_chooser_widget.notify["rgba"].connect (on_color_changed);
+    }
+    
+    private void on_eyedropper_click () {
+        eyedropper = new Akira.Utils.ColorPicker ();
+        eyedropper.show_all ();
+
+        eyedropper.picked.connect ((picked_color) => {
+            alpha = ((int)(picked_color.alpha * 255));
+            color_chooser_widget.set_rgba (picked_color);
+            on_color_changed ();
+            eyedropper.close ();
+        });
+
+        eyedropper.cancelled.connect (() => {
+            eyedropper.close ();
+        });
     }
 
     private void on_model_changed () {
--- a/src/Layouts/Partials/BorderItem.vala
+++ b/src/Layouts/Partials/BorderItem.vala
@@ -216,7 +216,7 @@ public class Akira.Layouts.Partials.BorderItem : Gtk.Grid {
         model.notify.connect (on_model_changed);
         color_chooser_widget.notify["rgba"].connect (on_color_changed);
     }
-    
+
     private void on_eyedropper_click () {
         eyedropper = new Akira.Utils.ColorPicker ();
         eyedropper.show_all ();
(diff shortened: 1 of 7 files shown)


bilelmoussaoui:
The current implementation won't work under wayland. The proper solution would be using the xdg portals to pick the color and fallback to this in case the dbus interface is not available. See https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-method-org-freedesktop-portal-Screenshot.PickColor

The XDG portals are not Flatpak specific, they are just a bunch of dbus interfaces (as a specification) that have to be implemented by the desktop environment, there are two implementation currently, kde & gtk which should cover most use cases.

isneezy:
@bilelmoussaoui thanks for your quick feedback. I'll first finish up this implementation and get it working and then I'll move forward into the XDG implementation as you suggested. I'm new to vala and Linux programming so in the meantime, I'll ping you for help.

bilelmoussaoui:
Sure, no problem :) thanks for the contribution! let us know if we can help somehow & good luck in your journey.

isneezy:
@bilelmoussaoui thanks. Do you know the name or how I can get the eyedropper icon? I Can't find any.

Alecaddd:
> @bilelmoussaoui thanks. Do you know the name or how I can get the eyedropper icon? I Can't find any.

The elementary icon theme doesn't currently have a symbolic color picker.
I will create a dedicated icon for it. For now don't worry about it and you can temporarily use the `preferences-color-symbolic`.

Thank you so much for your contributions, you're helping us a lot!

src/Lib/ColorPicker.vala
```diff
@@ -0,0 +1,332 @@
...
+* GNU General Public License for more details.
+
+* You should have received a copy of the GNU General Public License
+* along with Akira. If not, see <https://www.gnu.org/licenses/>.
+* 
+* Authored by: Ivan "isneezy" Vilanculo <vilanculoivan@gmail.com>
+* Ported from: https://github.com/ColorPicker/RonnyDo
+*/
+
+public class Akira.Lib.ColorPicker : Gtk.Window {
```
Alecaddd, line 23: This feels more like a utility tool which we might start using in multiple locations.
Let's move this into `/Utils`

src/Layouts/Partials/FillItem.vala
```diff
@@ -166,6 +168,14 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
         fill_chooser.attach (color_container, 1, 0, 1, 1);
         fill_chooser.attach (opacity_container, 2, 0, 1, 1);
 
+        eye_dropper_button = new Gtk.Button ();
+        eye_dropper_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
+        eye_dropper_button.get_style_context ().add_class ("button-rounded");
+        eye_dropper_button.can_focus = false;
+        eye_dropper_button.valign = Gtk.Align.CENTER;
+        eye_dropper_button.add (new Gtk.Image.from_icon_name ("user-trash-symbolic",
+            Gtk.IconSize.SMALL_TOOLBAR));
```
Alecaddd, line 177: Let's add a `tooltip_text` to this button that reads "Pick color".
I'll need to add tooltips for all the other icons.

src/Layouts/Partials/FillItem.vala
```diff
@@ -191,20 +201,38 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
...
+        attach (eye_dropper_button, 1, 0, 1, 1);
+        attach (hidden_button, 2, 0, 1, 1);
+        attach (delete_button, 3, 0, 1, 1);
 
         set_color_chooser_color ();
         set_button_color ();
     }
 
     private void create_event_bindings () {
+        eye_dropper_button.clicked.connect (on_eye_dropper_click);
```
Alecaddd, line 213: Instead of triggering this from within the fill item, I think we should have a globally accessible action so users can quickly copy a color with an handy shortcut.
The shortcut should be `Alt+C`, and the action should check if not items are currently selected.
Once the color is picked, the method should loop through all the currently selected items and apply the same color (since we're working on multiselect, it's better to have things in place right at the start).

src/Layouts/Partials/FillItem.vala
isneezy: Yah I thought about it, I was thinking on making a PR adding a tooltip on everything there. But no prob, I'll

src/Layouts/Partials/FillItem.vala
isneezy: The color picker must work with every part that contains colors, such as fills and borders, if we add it as a globally accessible action how do we know where it should be applied, borders or fills?

Alecaddd: By default it should interact with fills, but if the user holds the SHIFT key down when selecting a color, it'll be applied to the border.

isneezy: @Alecaddd I'm having a problem setting the color after picking via global action, it is not being reactive.
Screen record from 2020-08-31 18 57 31

Alecaddd: I need to look at the code of the action to check potential issues.

isneezy: I've pushed the code with the global action, please advise. thanks.

src/Layouts/Partials/BorderItem.vala
```diff
@@ -191,20 +203,38 @@ public class Akira.Layouts.Partials.BorderItem : Gtk.Grid {
...
     }
 
+    private void on_eyedropper_click () {
+        eyedropper = new Akira.Utils.ColorPicker ();
+        eyedropper.show_all ();
+
+        eyedropper.picked.connect ((picked_color) => {
+            alpha = ((int)(picked_color.alpha * 255));
+            color_chooser_widget.set_rgba (picked_color);
+            on_color_changed ();
```
Alecaddd, line 229: This is not necessary since the changes of the RGBA of the color chooser is binded to that method.

src/Layouts/Partials/FillItem.vala
```diff
@@ -193,20 +205,38 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
...
     }
 
+    private void on_eyedropper_click () {
+        eyedropper = new Akira.Utils.ColorPicker ();
+        eyedropper.show_all ();
+
+        eyedropper.picked.connect ((picked_color) => {
+            alpha = ((int)(picked_color.alpha * 255));
+            color_chooser_widget.set_rgba (picked_color);
+            on_color_changed ();
```
Alecaddd, line 231: Same thing here

src/Services/ActionManager.vala
```diff
@@ -452,6 +456,45 @@ public class Akira.Services.ActionManager : Object {
...
+                    }
+                }
+            } else {
+                // TODO change current selected shape border color
+                // window.main_window.main_canvas.canvas.selected_bound_manager.delete_selection ();
+            }
+            color_picker.close();
+        });
+    }
+
```
Alecaddd, line 497: I updated the method with this: 
```suggestion
    private void action_pick_color () {
        weak Akira.Lib.Canvas canvas = window.main_window.main_canvas.canvas;
        // Interrupt if no item is selected.
        if (canvas.selected_bound_manager.selected_items.length () == 0) {
            return;
        }

        foreach (var item in canvas.selected_bound_manager.selected_items) {
            // Hide the ghost bound manager.
            item.bounds_manager.hide ();
        }

        bool is_holding_shift = false;
        var color_picker = new Akira.Utils.ColorPicker ();
        color_picker.show_all ();

        color_picker.key_press_event.connect (e => {
            is_holding_shift = e.keyval == Gdk.Key.Shift_R;
            return true;
        });

        color_picker.key_release_event.connect (e => {
            is_holding_shift = e.keyval == Gdk.Key.Shift_R;
            return true;
        });

        color_picker.cancelled.connect (() => {
            color_picker.close ();
        });

        color_picker.picked.connect (color => {
            foreach (var item in canvas.selected_bound_manager.selected_items) {
                if (is_holding_shift) {
                    item.border_color_string = Utils.Color.rgba_to_string (color);
                } else {
                    item.color_string = Utils.Color.rgba_to_string (color);
                }
                item.load_colors ();
            }
            color_picker.close ();
        });
    }
```

It properly updates the fill color, but it doesn't update the fill or border widget in the options panel as those models are binded the other way around (model -> item), instead here we're doing the opposite (item -> model).

I think I'm fine with merging this as it is, as I'm not sure we can directly edit the fill item or the fill item model.
Maybe a restructure of that section is necessary, or maybe not, I don't know :sweat_smile:

Alecaddd: Oh, this suggestion won't work without a tiny update to the `Color.vala`

```
public static string rgba_to_hex (string rgba_string) {
        var rgba = Gdk.RGBA ();
        rgba.parse (rgba_string);

        return rgba_to_string (rgba);
    }

    public static string rgba_to_string (Gdk.RGBA rgba) {
        return "#%02x%02x%02x".printf (
            (int) (rgba.red * 255),
            (int) (rgba.green * 255),
            (int) (rgba.blue * 255)
        );
    }
```

src/Utils/ColorPicker.vala
```diff
@@ -0,0 +1,332 @@
...
+
+* You should have received a copy of the GNU General Public License
+* along with Akira. If not, see <https://www.gnu.org/licenses/>.
+* 
+* Authored by: Ivan "isneezy" Vilanculo <vilanculoivan@gmail.com>
+* Ported from: https://github.com/ColorPicker/RonnyDo
+*/
+
+public class Akira.Utils.ColorPicker : Gtk.Window {
+        public signal void picked (Gdk.RGBA color);
```
Alecaddd, line 24: This whole file is wrongly indented.

src/Utils/ColorPicker.vala
```diff
@@ -0,0 +1,332 @@
+/*
+* Copyright (c) 2019-2020 Alecaddd (https://alecaddd.com)
```
Alecaddd, line 2: Just 2020 as this file was created this year.

src/meson.build
```diff
@@ -94,6 +94,8 @@ sources = files(
     'Lib/Models/CanvasArtboard.vala',
 
     'Lib/Selection/Nob.vala',
+
+    'Utils/ColorPicker.vala',
```
Alecaddd, line 98: Move this underneath the other `Utils/` files.

Alecaddd (approved):

PR #435, opened 2020-09-01
Improve canvas zoom

Alecaddd:
Improve the various zoom interactions to offer a better user experience.

## Steps to Test
1. Zoom with the <kbd>CTRL</kbd> + scroll
2. Zoom with <kbd>CTRL</kbd> + <kbd>+</kbd> or <kbd>-</kbd>
3. Zoom with the HeaderBar buttons.

## Screenshots 
TBD

- Zoom by 10% when using the keyboard shortcut.
- Zoom by 50% when using the HeaderBar button.
- Zoom by delta variation when using the mouse scroll.

### DIFF
--- a/src/Layouts/MainCanvas.vala
+++ b/src/Layouts/MainCanvas.vala
@@ -115,6 +115,11 @@ public class Akira.Layouts.MainCanvas : Gtk.Grid {
         if (delta_y < -SCROLL_DISTANCE) {
             // Scroll UP.
             if (is_ctrl) {
+                // Divide the delta if it's too high. This fixes the zoom with
+                // the mouse wheel.
+                if (delta_y <= -1) {
+                    delta_y /= 10;
+                }
                 // Get the current zoom before zooming.
                 double old_zoom = canvas.get_scale ();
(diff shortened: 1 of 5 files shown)


Alecaddd:
@isneezy would you be willing to give this PR a review, since you've been very active lately and you worked on some aspects of the zoom architecture?

isneezy (commented):
@Alecaddd  I see you have prevented the canvas from shrink below 2% which is great and almost everything looks fine to me.
What bothers me is the fact that when you zoom out to 2% and then zoom in again it goes like by 12%, 22%, etc. IDK if we want to keep it like this or do some kind of correction.

Alecaddd:
@isneezy that's the tricky part of zooming with the mouse, as it always returns a delta of 1 or -1, so I'm dividing it by 10 to prevent zooming in an out by 100%.
Scrolling with a trackpad is more natural as it actually returns the real acceleration.
I'm not sure if there's a method to get the proper delta acceleration of the mouse wheel.

Alecaddd:
Fixed!

PR #443, opened 2020-09-21
Fix permissions when saving/opening Akira files

### LINKED ISSUE #441: Akira only opens design files when running as superuser
## Expected Behaviour
Akira should be able to open design files regardless of the user's permissions.

## Current Behaviour
Akira needs root permissions to access the content.json file when decompressing it from the design file (possibly by compressing it without setting permissions and preserving them when decompressing).

## Possible Solution
Change file permissions when compressing and decompressing (for older versions with this bug).

1. Follow the instructions in the README.md file (Compile & Run section)
2. Open Akira...
3. Insert any shape
4. Save the file
5. Close Akira, then reopen
6. Try to open the previously saved file...

## Context
A design application must be fully functional when opening and saving files.
I tried to design something, save it to continue later, but the design didn't show, at first I was scared because I thought it didn't save, then I checked the log, got the permission denied error for the content.json file (ZipArchiveHandler.vala:62), tried to run Akira as superuser and it "worked".

- OS Info: Ubuntu 20.10

xoascf:
Fixes file permissions by compressing and decompressing the Akira file format.

- Fixes #441


src/FileFormat/ZipArchiveHandler.vala
```diff
@@ -259,6 +259,7 @@ public class Akira.FileFormat.ZipArchiveHandler : GLib.Object {
         Archive.Result last_result;
 
         while ((last_result = archive.next_header (out entry)) == Archive.Result.OK) {
+            entry.set_perm (0644);
```
albfan, line 262: In libarchive basic examples it does this too:

https://github.com/libarchive/libarchive/wiki/Examples#A_Basic_Write_Example

> archive_entry_set_perm(entry, 0644);

so probably it's ok to do that. Not sure

albfan (approved):

src/FileFormat/ZipArchiveHandler.vala
Xoas: When unzipping, the permissions for the unzipped files (0000) are changed with the new permissions (0644) so ​​that they can be accessed.

albfan (commented):
Libarchive vapi comes bundled with vala.

You need to remove it.

src/FileFormat/ZipArchiveHandler.vala
```diff
@@ -329,12 +330,11 @@ public class Akira.FileFormat.ZipArchiveHandler : GLib.Object {
 #if VALA_0_42
                     entry.set_size ((Archive.int64_t) file_info.get_size ());
                     entry.set_filetype (Archive.FileType.IFREG);
-                    entry.set_perm (Archive.FileType.IFREG);
 #else
                     entry.set_size (file_info.get_size ());
                     entry.set_filetype ((uint) Posix.S_IFREG);
-                    entry.set_perm (0644);
 #endif
+                    entry.set_perm (0644);
```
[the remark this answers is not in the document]
xoascf: My first test was on Ubuntu 20.10 with Vala 0.50.0.4-69cc1 and Libarchive 3.4.3.
Then I used Elementary OS Hera with Vala 0.40.23 and Libarchive 3.2.2.
Both the first and second commits compile without issues in both environments.

`entry.set_perm (0644)` should be safe to use with stable versions of Libarchive.
I will revert the second commit, no problem.

src/FileFormat/ZipArchiveHandler.vala
xoascf: I put it there for the people who saved the file and compiled Akira like me so they can then open when the file doesn't have permissions, you don't read the comments? (I can't blame you, GitHub doesn't update them, you have to refresh the page).

xoascf:
> Libarchive vapi comes bundled with vala.
> 
> You need to remove it.

I know, what happens is that previous versions of the Libarchive vapi have problems with regard to arguments and methods not taken advantage of by the vapi (which is why this conditional compilation had to be done).

albfan:
Fixed on #445

PR #445, opened 2020-09-23
Handle fix in libarchive permissions

### LINKED ISSUE #441: Akira only opens design files when running as superuser

albfan:
from version vala 0.46 b8e1a4e5 libarchive set_perm is fixed upstream, we need to deal with it conditionally.

## Steps to Test
Save/open a file works without any permissions problem.

Ensure how much backward compatible we want to be.

- Fixes #441

### DIFF
--- a/src/FileFormat/ZipArchiveHandler.vala
+++ b/src/FileFormat/ZipArchiveHandler.vala
@@ -259,6 +259,11 @@ public class Akira.FileFormat.ZipArchiveHandler : GLib.Object {
         Archive.Result last_result;
 
         while ((last_result = archive.next_header (out entry)) == Archive.Result.OK) {
+#if VALA_0_46
+            entry.set_perm (0644);
+#else
+            entry.set_perm (Archive.FileType.IFREG);
+#endif
             entry.set_pathname (Path.build_filename (location.get_path (), entry.pathname ()));
 
             if (extractor.write_header (entry) != Archive.Result.OK) {
@@ -329,11 +334,15 @@ public class Akira.FileFormat.ZipArchiveHandler : GLib.Object {
 #if VALA_0_42
                     entry.set_size ((Archive.int64_t) file_info.get_size ());
                     entry.set_filetype (Archive.FileType.IFREG);
-                    entry.set_perm (Archive.FileType.IFREG);
 #else
                     entry.set_size (file_info.get_size ());
                     entry.set_filetype ((uint) Posix.S_IFREG);
+#endif
+
+#if VALA_0_46
                     entry.set_perm (0644);
+#else
+                    entry.set_perm (Archive.FileType.IFREG);
 #endif
 
                     if (archive.write_header (entry) != Archive.Result.OK) {


xoascf:
Please don't!

xoascf:
Please don't!

xoascf:
Search first!!
https://github.com/search?l=&p=1&q=Archive.FileType.IFREG+language%3AVala&ref=advsearch&type=Code

xoascf (commented):
VIM swap file?

src/FileFormat/ZipArchiveHandler.vala
```diff
@@ -259,6 +259,11 @@ public class Akira.FileFormat.ZipArchiveHandler : GLib.Object {
         Archive.Result last_result;
 
         while ((last_result = archive.next_header (out entry)) == Archive.Result.OK) {
+#if VALA_0_46
+            entry.set_perm (0644);
+#else
+            entry.set_perm (Archive.FileType.IFREG);
```
xoascf, line 265: No, Archive.FileType.IFREG is not a type for permission (that will result in no permissions).
Archive.FileType.IFREG is a file type constant for a regular file.

xoascf:
We can discuss by some other method how we can decide a better solution on this (I suggest by Twitter).

src/FileFormat/ZipArchiveHandler.vala
albfan: Yes, this is not about code something that is correct, but something that will build on older systems.

albfan:
There's no better place to discuss than here.

You can downgrade SDK on flatpak till vala is below 0.46 to see how this will fail to build.

We don't want to maintain dependencies here, like libarchive if possible but report upstream.

src/FileFormat/ZipArchiveHandler.vala
albfan: FTR: https://gitlab.gnome.org/GNOME/vala/-/commit/b8e1a4e591fde375dc46c72d27ed31e2d06ef072#2b1b0611bd0ea56b6e797abed1e21fd93418ca6a_436_443

Maintaining is harder than just "fix for me", and yes make you do weird things like this.

When we are aware that every possible distro is upon vala 0.46 we can remove conditional code.

So the question here is: @xoascf do this PR solve your case, it should.

Xoas: Ok... will try

xoascf: Um... You didn't really read my comments in the previous PR, did you?
It's the same [link I wrote in a commentary](https://gitlab.gnome.org/GNOME/vala/-/commit/b8e1a4e591fde375dc46c72d27ed31e2d06ef072), and I quoted you [why](https://github.com/akiraux/Akira/issues/441#issuecomment-696686869) this [shouldn't be done](https://github.com/akiraux/Akira/issues/441#issuecomment-697016611).
A bad approach and this didn't even compile well ._.
Even worse, the version that compiled, has the same issue, won't save with the correct file permissions.

The best way not to complicate people is with an upstream version of the Libarchive vapi (fixed without compatibility issues, unlike the one provided by Vala in versions prior to 0.46), because that way it compiles with any previous version (from version 0.40 to the latest current one) and it doesn't need any of that unnecessary conditional compilation bs.


A similar case used upstream vapis in the past because of that same issue with versions, gnome-boxes.
And then, when those stable versions of vapis were already distributed in distros with that same fixed upstream version, they were removed later from the repo, and this is not the case, at least not in Elementary OS (that still uses Vala 0.40 and those outdated vapis).

And indeed, you fixed my problem (on my main machine with Ubuntu 20.10), but on my computer with elementary OS 5.1 Hera...
```
../src/FileFormat/ZipArchiveHandler.vala:265.29-265.44: error: The name `FileType' does not exist in the context of `Archive'
            entry.set_perm (Archive.FileType.IFREG);
                            ^^^^^^^^^^^^^^^^
../src/FileFormat/ZipArchiveHandler.vala:345.37-345.52: error: The name `FileType' does not exist in the context of `Archive'
                    entry.set_perm (Archive.FileType.IFREG);
                                    ^^^^^^^^^^^^^^^^
Compilation failed: 2 error(s), 0 warning(s)

```

And the other one with the updated vapis, Akira does not save or open files...

xoascf:
Please [read](https://github.com/akiraux/Akira/pull/445#discussion_r494479772).

I want to discuss. If you are willing to do so, we can do it by Telegram ([@xoascf](https://t.me/xoascf)), I don't have a problem, I think you also speak Spanish, and anyway I would like to talk to you about this and other topics/issues, otherwise, well... this was interesting.

src/FileFormat/ZipArchiveHandler.vala
albfan: Welcome to the downstream maintenance hell.

That's elementary OS 5.1 vala version?

xoascf: Yep

xoascf: Interesting system, poor maintenance by the developers.

albfan: Elementary 5.1 is based on Ubuntu lts 18

Comes with vala 0.40 and the libarchive vapi is:

```
public void set_perm (Posix.mode_t mode);
```

Thank you God for flatpak!

xoascf: Flatpak seems to be the least of the problems (in fact I don't think it's a problem per se _like snap it is_).
But now the [elementary OS AppCenter Akira App](https://appcenter.elementary.io/com.github.akiraux.akira/)... would have to compile with another method, or with the updated vapis (?)

albfan: flatpak is a sandbox so you release exactly what you build. Of course we can left downstream to patch the build in each system, but as we focus on elementary OS we just try to prevent downstream package maintainers to bother with this details.

albfan:
we have a public irc channel, this is the matrix bridge

https://app.element.io/#/room/#akiraux:matrix.org

But in general the best idea is to open an issue for each topic (if there's any action to take on it)

albfan:
Let's work from here. Our CI build on top of https://hub.docker.com/r/elementary/docker and build is passing.

We can continue from there

PR #449, opened 2020-10-04
Generate and upload AppImage

### LINKED ISSUE #3: Appimage package
Look forward to using your app once it comes out as i'm currently using balsamiq and pencil for UI designing and would suggest that you package it as an appimage to ensure it easily runnable on any linux distro.

https://github.com/AppImage/AppImageKit

probonopd:
Generate and upload AppImage, closes #3.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,12 +15,10 @@ jobs:
     - name: Install Dependencies
       run: |
         apt update
-        apt install -y libarchive-dev libgoocanvas-2.0-dev valac meson appstream-util xvfb at-spi2-core
+        apt install -y libarchive-dev libgoocanvas-2.0-dev valac meson appstream-util xvfb at-spi2-core wget librsvg2-bin appstream elementary-theme elementary-icon-theme libgtk-3-common
     - name: Build
-      # env:
-      #   DESTDIR: out
       run: |
-        meson build
+        meson build --prefix /usr
         ninja -C build
         ninja -C build install
         xvfb-run -a -s "-screen 0 1024x768x24" \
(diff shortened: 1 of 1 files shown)


albfan:
Personally rather prefer that you keep appimage in your own external repository. If there's any need to fix something (like this)  we will just block the fix without really reviewing anything, as we are not experts on appimage.

@Alecaddd will now better what to do.

probonopd:
> Personally rather prefer that you keep appimage in your own external repository. 

* The PR will build an AppImage for each git commit, which is very useful for testing new features and bugfixes. This only works if the AppImage is generated in this repository.
* AppImage is for "upstream packaging". It is a format that allows the original authors to distribute applications to end users _directly_. In fact, we are strongly advising users against downloading binaries from random 3rd parties.

So it would be highly beneficial for the AppImage to be created here.

albfan:
not sure if turn appimage into a nigthly testing feature is what users expect from it.

Users want to try releases.

On contrary, when there's a bug fix detected on appimage, fix appimage setup turn us into blockers (we probably didn't notice if there's any fix needed to be applied to appimage, or if we break something with new commits on appimage).

Maintain a release system at zero cost basing it on deliver master is just an illusion that will cause poor expectations on users.

We have a tag versioning to not be afraid to commit on master and share bleeding edge changes between developers. keep appimage here is breaking that, just to allow something that an user do not want.

I say all this based on other examples, like maintaining glade and getting reports of snatcraft not working, and suggestion of changes we cannot test and just blindy accept. Is a mess.

Is it possible to create github actions listening to tags in other repo? Maybe we can link that. I can only see for same repo:

https://github.community/t/how-to-run-github-actions-workflow-only-for-new-tags/16075/2

As a summary: am I maintaning appimage? As the answer is no, there's no point to keep files here.

probonopd:
> As a summary: am I maintaning appimage? As the answer is no, there's no point to keep files here.

I don't quite understand. Are you saying the Akira project is not interested in providing and supporting official AppImages?

Alecaddd:
We're interested in offering Akira in Appimage version, but I think a couple of things need to change in order to be properly handled.

First, we should hook the building of the Appimage only when there's a merge into a specific branch, for example `master`, as it happens right now for the edge channel of the Snap packaging.
This will ensure the possibility for use to target a specific branch after we move past the alpha cycle, and having a dedicated `release` or `deploy` branch where we can push stable releases.

Second, we need someone willing to support and fix Appimage specific issues. As Alberto said, we're not expert of this format and we wouldn't know how to properly support it. We already have a similar issue with our snaps, and when something goes wrong we stumble in the dark for a bit, so, not great for maintainability.

Last thing, it seems the that building process encountered an error.

probonopd:
> possibility for use to target a specific branch after we move past the alpha cycle

Sounds like this is a matter of putting the relevant code into the right place within GitHub Actions. I am not very experienced with GitHub Actions yet (having used mainly Travis CI so far). Do you know how to do this?

> Second, we need someone willing to support and fix Appimage specific issues.

That'd be me, then.

> Last thing, it seems the that building process encountered an error.

"Resource not accessible by integration" just means that a pull request is not authorized to upload artifacts to GitHub Releases. It is not a real error. We can script around it by not trying to upload anything in the case of a pull request.

https://github.com/akiraux/Akira/pull/449/checks?check_run_id=1205738857#step:8:26

Alecaddd:
> I think it should be possible to tell GitHub Actions to run the AppImage action only on tagged releases.
(However, you'd lose a great feature - namely, having AppImages for each commit, which can be very handy for testing!)

We prefer to create releases for specific branches in order to target builds when new features and fixes have been tested on master by the core team.

> "Resource not accessible by integration" just means that a pull request is not authorized to upload artifacts to GitHub Releases. It is not a real error. We can script around it by not trying to upload anything in the case of a pull request.

Could you update the action to do that?

Alecaddd:
Any change of getting back to this?

Alecaddd:
Would be possible to have this in a dedicated GitHub repository that you manage?
Something like https://github.com/flathub/com.github.akiraux.akira, so we keep packages and releases separated and we don't leave the burden of supporting this to the main maintainers of this repo.
Closing this as we can't support it on our own.

probonopd:
Hi @Alecaddd, AppImage is designed as a technology that enables the original applcation author to put out and support official binaries. It is generally discouraged to download AppImages from random third parties. I am not interested in distributing other people's applications, because I don't understand how they work and hence cannot support them. But I am more than happy to support application developers who want to offer AppImages.

albfan:
Something is really wrong when a distribution management forces upstream source to keep the config files for it on main repo.

Looks nearly normal because no other distribution system (flatpak, ubuntu, archlinux, fedora...) ask you to do the same. If so you will see the pattern and look for better alternatives.

Is not you distributing other applications, is whatever appimage is under with (their shop, appstore, whatever) which should care about releasing the next thing for each application.

We, akira maintainers have same concern about keeping files we don't know how they work in our repo, totally out of context as a distribution should only exist on last release day, not a permanently broken config files that we fix for each release.

Break the expectation about "maintainers knows what all files in their repos are for" is deeply wrong.

If this help to move things on appimage great, I guess you hear all this before.

PR #456, opened 2020-10-25
Fix LibArchive permission for vala 40. Regressed by #445

Alecaddd:
PR #445 introduced a regression for elementary OS 5 users, or distribution with Vala v40.
`Archive.FileType.IFREG` doesn't exist in that version, and if we restrict the `entry.set_perm (0644);` to v46+, we prevent opening previously saved files in v40.

Let me know if this breaks anything for newer versions of Vala.


albfan:
That's basically the suggestion from reporter in #445, so let's merge that.

albfan (approved):

PR #465, opened 2021-01-15
Remove support for Vala version < 0.42

Alecaddd:
Very simple PR to remove the support for Vala code prior to v0.42.

Please, be sure to test that the saving and opening of `.akira` files works.


albfan (approved):
Yes, I think we shouldn't care about older versions, that's job for downstream distros

PR #466, opened 2021-01-15
Fix Arabic Translation

AliSamehGalal:
# correct the word البيانت to البيانات


Alecaddd:
Thanks for your contribution.

PR #469, opened 2021-01-19
Fix loading items with negative origin

### LINKED ISSUE #468: Items lost when opening a saved file
## Expected Behavior
All the items should be regenerated and put in the proper place when opening a saved file

## Current Behavior
Items that have their origin outside the boundaries of a Canvas Artboard disappear upon reopening.

## Possible Solution
Account for negative origin points when saving a file, even if we don't show negative values in the transform panel.
Maybe we should start supporting negative values for items inside artboards.

1. Create an Artboard
2. Create a shape inside the Artboard.
3. Move the shape in order to have its top-left origin point outside the Artboard, but the rest of the body inside.
4. Save the file and close Akira.
5. Reopen the saved file.

## Screenshots of the Problem

This is how the file should open back, but if you save the file with a shape partially outside the Artboard, it gets lost upon reopening.
Screenshot from 2021-01-17 22-12-51

Alecaddd:
Fixes the loading of items inside Artboards that were saved with a negative origin.

- Fixes #468

### DIFF
--- a/src/Layouts/Partials/TransformPanel.vala
+++ b/src/Layouts/Partials/TransformPanel.vala
@@ -189,7 +189,6 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
         }
 
         selected_item = selected_items.nth_data (0);
-
     }
 
     private void disconnect_previous_item () {
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -157,6 +157,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
             case Gdk.Key.Right:
             case Gdk.Key.Left:
                 window.event_bus.move_item_from_canvas (event);
+                window.event_bus.detect_artboard_change ();
                 break;
         }
 
@@ -283,9 +284,12 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                 edit_mode = EditMode.MODE_SELECTION;
                 break;
 
+            case EditMode.MODE_SELECTION:
+                window.event_bus.detect_artboard_change ();
+                break;
+
             default:
                 edit_mode = EditMode.MODE_SELECTION;
-                window.event_bus.hold_released ();
                 break;
         }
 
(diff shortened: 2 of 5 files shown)


albfan (changes requested):
Sounds to me like what we should do is check coordinates+height and width are inside artboard.

Even store which items are inside and artboard (in artboard) or if item has an artboard (in item) will allow to detect this without maths involved.

Alecaddd:
> Sounds to me like what we should do is check coordinates+height and width are inside artboard.
Even store which items are inside and artboard (in artboard) or if item has an artboard (in item) will allow to detect this without maths involved.

We do that already.
This is what this code is about when loading an item: https://github.com/akiraux/Akira/blob/master/src/Lib/Managers/ItemsManager.vala#L390-L401

If the item belongs to an Artboard, we recreate it in the Canvas at the proper position, accounting for the Artboard's boundaries.
We need to use the global Canvas coordinates as origin point as we can't create an item at position 4px top and 4px left of the parent Artboard.

If the TOP-LEFT origin position is outside the artboard upon creation, the item won't belong to that artboard.

Alecaddd:
I tried to improve the code based on your suggestion, but I'm encountering a big performance issue.
I'm currently running the event `detect_artboard_change ()` on 2 conditions:
- After an item has been moved with the mouse
- After an item has been moved with the arrow keys

The method runs for all the currently selected items to see if they were dropped inside or outside an artboard, and change things accordingly.

The problem when loading the file, is that we don't have all the items selected, and we're updating the attributes (x, y, width, height, etc.) directly without passing through a mouse or keyboard interaction.

The only way I can think of listening for all the changes on load is by triggering the `detect_artboard_change ()` event at every attribute change for every item, even if they're not selected. 
That's unfortunately a very expensive method to run which causes a segfault.

I don't know if you have a better approach in mind.

albfan:
Here is the problem:

https://github.com/akiraux/Akira/blob/master/src/Lib/Managers/ItemsManager.vala#L103

that position is outside the canvas, because we are not taking into account the item height and width. Let me fix it

albfan (commented):
Suggest to close in favour of #476

Alecaddd:
Closing this in favour of #476

PR #476, opened 2021-01-21
Use metadata to detect items added to an artboard

### LINKED ISSUE #468: Items lost when opening a saved file

albfan:

- Allow position to have negative values
- Reuse akira file metadata to detect if an item is under artboard

## Steps to Test
Load example file: 
[artboard-negative-item.zip](https://github.com/akiraux/Akira/files/5845992/artboard-negative-item.zip)

result:

Captura de pantalla de 2021-01-21 01-13-43
 
Probably check if artboard is null is dead code, but still not sure about new inserted items

- Fixes #468 
- Supersedes to #469

### DIFF
--- a/src/Layouts/Partials/TransformPanel.vala
+++ b/src/Layouts/Partials/TransformPanel.vala
@@ -94,9 +94,9 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
         hexpand = true;
 
         x = new Akira.Partials.LinkedInput (_("X"), _("Horizontal position"));
-        x.input_field.set_range (0, Akira.Layouts.MainCanvas.CANVAS_SIZE);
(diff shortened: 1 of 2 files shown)


Alecaddd (approved):
Yes, this is a great solution.
The condition to check if the artboard is null is good as we don't run the foreach loop if we don't need it, for example when opening a saved file.

PR #477, opened 2021-01-21
Improve artboard interactions

Alecaddd:
Adding some code improvements and clean up to the artboard changing method.

## Steps to Test
- Create a file with multiple artboards and multiple objects.
- Move the objects between artboards and the emtpy canvas.
- Use the arrow keys to move objects around.
- Move the layers between artboards in the layers panel.

- Converts the artboard changing method to async.
- Prevents running the method multiple times if already running (this will probably need some reworking after the multi selection is implemented).
- Change artboard when the item is moved outside with the arrow keys.
- Exclude the artboard label from the artboard's dropping area.


albfan (approved):
Looks good to me

PR #478, opened 2021-01-24
Implement State Managers to handle coordinates

Alecaddd:
Implement a State Manager to keep track of the item coordinates and update the various layout elements via Events.
This approach is to solve the problems caused by the 2-way-binding approach explained here: https://github.com/akiraux/Akira/discussions/470

This is only necessary for the X and Y values since `GooCanvas` doesn't offer those attributes to be directly accessible like it does for the others. E.g.: for the width we can simply do a `selected_item.notify["width"]` and the 2 way binding works in a bidirectional way out of the box.

## Steps to Test
- Create multiple items
- Move the items from the canvas
- Move the items from the transform panel

It doesn't work super well as sometimes values are skipped and fields don't update properly.
There are still a lot of conditions and events called in some sketchy loops.
I'm planning to track those things down and clean them up in follow up PRs.


Alecaddd:
@albfan @giacomoalbe this is a WIP PR, so nothing is final, but what do you guys think about this approach?

Alecaddd:
@jtrees pinging you to get your feedback on this first implementation.

With this changes the `State/PositionManager` handles the X & Y coordinates.

- When a `Shape` is changed in the canvas, it sends an event to the `PositionManager` which updates the state values and sends an event to the `Transform Panel`.
- When the value is changed in the `Transform Panel` a signal is sent to the `PositionManager` which updates the state values and sends an event to the `Shape` telling it to update itself.

### A little problem
In order to emit a signal when the X & Y values are changed in the `Transform Panel` I need to connect an event listener. This listener triggers also when the value is changed from within the code, not only when the user interacts with the input field.
I'm preventing the emission of another signal in case the X value of the `Transform Panel` is identical to the X value of the state in the `PositionManager`, but I feel like this is not a great solution and my implementation is not exactly as it should be.

Alecaddd:
Ignore this for now as I found another approach which should solve almost all of our problems.

Basically the `State/PositionManager` will hold the X and Y coordinates of the item in the lowest TOP and LEFT position, (this is to handle multiple selected items) and whenever the user interacts with the input fields in the Transform panel, or manually moves the shape with the mouse or arrow keys in the canvas, first we update the X & Y in the state manager, and those value changes trigger the actual update of the Shape and Transform Panel.

No more direct manipulation of the Shapes via the Affine Transform, as everything needs to pass through the State Manager.

jtrees:
Sorry, my availability has changed since we last talked. I don't expect to get round to reviewing this any time soon.

Alecaddd:
> Sorry, my availability has changed since we last talked. I don't expect to get round to reviewing this any time soon.

No worries at all and thanks for the heads t.
I'm still working on this and it's 50% done.
Slowly getting there 😉

Alecaddd:
Closing this in favor of a new upcoming branch which tackle this in smaller, more compartmentalize way.

PR #480, opened 2021-01-24
CI: use a custom elementary image

bilelmoussaoui:
The image contains the elementary base app along with the gnome sdk, this should reduce the build time a bit.


Alecaddd:
This is awesome!

bilelmoussaoui:
We saved ~1 minute, not as much as I expected but it's already a step in the right direction i guess :P

PR #482, opened 2021-02-04
Implement State Managers to handle coordinates

Alecaddd:
Implement a State Manager to keep track of the item coordinates.
This approach is to solve the problems caused by the 2-way-binding approach explained here: https://github.com/akiraux/Akira/discussions/470

This is only necessary for the X and Y values since `GooCanvas` doesn't offer those attributes to be directly accessible like it does for the others. E.g.: for the width we can simply do a `selected_item.notify["width"]` and the 2 way binding works in a bidirectional way out of the box.

## Steps to Test
- Create multiple items
- Move the items from the canvas
- Move the items from the transform panel

## What issue does this solve
- Items have the proper coordinates upon creation, as before they all started with 0 or negative coordinates for items inside artboards.
- Properly handles the difference between updating an item from the transform panel or from the canvas.
- Reduces the infinite loop of events when the coordinates change.

## Things to do
This is the first of many PRs that aim to rebuild and simplify our architecture to remove all those unnecessary Caitro.Matrix calls and start using built in GooCanvas methods.
It's gonna take us a while, but this is the first step to simplify things.

### DIFF
--- a/src/Layouts/Partials/TransformPanel.vala
+++ b/src/Layouts/Partials/TransformPanel.vala
@@ -179,7 +179,6 @@ public class Akira.Layouts.Partials.TransformPanel : Gtk.Grid {
         attach (opacity_grid, 0, 10, 3);
 
         window.event_bus.selected_items_changed.connect (on_selected_items_changed);
-        //  window.event_bus.item_coord_changed.connect (on_item_coord_changed);
     }
 
     private void on_selected_items_changed (List<Lib.Models.CanvasItem> selected_items) {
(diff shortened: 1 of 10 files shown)


Alecaddd:
@albfan this is ready!
I update the State coordinates manager to use built-in `GooCanvas` methods and dropped the `Cairo.Matrix` transformation were were wrongly doing to translate the items.
All the items outside artboards now work like a charm.

Items inside artboards are a bit messed up right now, but that's expected and I'm planning to rebuild the `CanvasArtboard` item next to stick to native `GooCanvas` methods.

albfan (approved):
Looks cleaner. LGTM

albfan (changes requested):
x and y are wrong calculated if you create an item directly inside and artboard.

It's a regression, previously it doesn't happen

Alecaddd:
> x and y are wrong calculated if you create an item directly inside and artboard.
> It's a regression, previously it doesn't happen

Yeah, I wrote it in the message before that the items inside artboards are messed up, but I'm planning to rebuild artboards from scratch, which will use the same coordinates methods.

Anyway, I put back the handling of coordinates for items inside artboards.

PR #485, opened 2021-02-08
Drop inheritance, use components

### LINKED ISSUE #405: Bug when resizing rotated items in Artboards
## Expected Behavior
Resizing rotated items inside artbaords should behave exactly as it does for free items.

## Current Behavior
When resizing rotated items inside artboards the X and Y coordinates change unexpectedly.

## Possible Solution
This might be an issue due to our `compute_transform` method and how we handle painting of items inside artboards.

## Screenshots of the Problem
artboard-item-resize

- OS Info: elementary OS 5.1.7

### LINKED ISSUE #455: Text not visible when dropped to Artboard
## Expected Behavior
When Text is dropped to an Artboard it should remain visible.

## Current Behavior
When Text is dropped to an Artboard it is getting white Color/ isn't rendered.

## Screenshots of the Problem
text_white_when_dropping_to_artboard_bug

- Akira Version: 0.0.13
- OS Info: Arch

Alecaddd:
#### 2. This code doesn't change anything in the App right now as the new items are not connected to the canvas and can't be created, yet.

---

- Drop the `inheritance` approach in favor of a `components` oriented approach.
- The goal is to create a more modular architecture that enables developers to create new items by extending multiple components instead of a single `Interface`, therefore avoiding the rewriting of all the inherited attributes.
- This also removes the necessity of updating each item class in case a new attribute is defined in the base class.
- This also removes the unnecessary `Cairo.Matrix` "abuse" we currently have in our `Models`, allowing a more solid implementation that relies on native `GooCanvas` methods.

## Steps to Test
Just look at the code as this can't be used, yet.

- [x] Create base classes and proof of concept.
- [x] Create all the `Components` to cover all the attributes we currently use.
- [x] Create the new `Items` with all the new `Components`.
- [x] Update the `Canvas` to use the new `Items`.
- [x] Update the UI to use the new `Items`.
- [x] Delete the old `Models`.

Related to #470
Fixes #405
Fixes #455

## Breakages

This PR breaks the previous save and load architecture due to the new structure of the items.
Not a big deal since we're still in alpha, and anyway the current saving and loading is a bit too cumbersome and needs to be rewritten.


Alecaddd:
@donadigo here's a first proof of concept, let me know what do you think, whenever you have time to take a look at it.
No rush.

donadigo (commented):
This is definitely going in the right direction, however I feel like you took the concept *too far*. It is very important to be reasonable about what can be it's own component to strike balance that allows you for easy maintanence in the future:

* Components such as `Name` or `Type` seem too basic or too core to be their own components. These can easily be incorporated into the core CanvasItem class, without much worry they will interfere with anything else, mostly because you'd expect an Item to have a name and a type. Even if it does not have a name, you can safely make a nullable property in the CanvasItem class `public string? name { get; set; default = null; }`. The type class also seems like something that could be done diferently. Vala already exposes you the type of each component as well as the item via `typeof (object)` or via the `is` comparator. Using this functionality, you can determine the true type of an object, even if it's downcasted into `CanvasItem`. As such, this information should not be used to determine the name or id of the item. Instead, you should define your own enum with available item types and then have a `to_string` method that translates it to a user facing name. Otherwise, your app will depend on the fact that the class ID is exactly starting with `AkiraLibItemsCanvas`. What happens if you rename any these namespaces?

* Components such as `Fills` or `Borders` that only hold a list of Fill components also seem unneded in this context. You can simply add individiual `Fill` and `Border` components directly to the `CanvasItem` object and then have a plural method that retrieves all components according to the type (pseudocode):
```
public Gee.ArrayList<T> get_components<T> () {
       var list = new Gee.ArrayList<T> ();
       foreach (Component comp in components) {
            if (comp is T) {
                   list.add (comp);
            }
        }
        
        return list;
}

// Usage:
`var fills = item.get_components<Fill> ();
```

I will keep more comments coming once I'm able to take a closer look.

Alecaddd:
@donadigo, thank you so much for looking at it and for your suggestions.

>  Even if it does not have a name, you can safely make a nullable property in the CanvasItem class public string? name { get; set; default = null; }

I'm not sure I can do that as the `CanvasItem` is an interface.
This is a limitation of GooCanvas I think, since in order to create a Rectangle, I can only extend the `Goo.CanvasRect` + an interface.
I know you suggested in your initial example to use a `RectangleRenderer` class to actually generate the shape, but GooCanvas doesn't like that and I would need to manually bind every property of a rectangle to the `CanvasItem` class.

> Components such as Fills or Borders that only hold a list of Fill components also seem unneded in this context.

Indeed, those are very basic right now, but we have some features in the pipeline that will require specific methods for these list of colors. Stuff like blending modes, sorting of multiple colors, reusable global colors, and so on. So I was planning to use those classes to manage these things.

Nonetheless, I think you're 100% right and the goal here is to streamline and remove redundancy.
Already at this stage I simplified a lot of things compared to my initial commits, and I'm continuing to remove unnecessary middle steps.

I'm very looking forward to more suggestions and improvements.
Thank you so much.

Alecaddd:
This is ready for a review, with 2.3K lines it's basically a full Canvas architecture rewrite!
Of course, I don't expect you to read every single line of code, that would be insane, so I simply invite you to use the app and try to spot bugs or crashes.

The only thing that doesn't work and it's broken is the Saving and Loading of files.
I left that out as I will take care of it in a follow up PR.

PR #501, opened 2021-02-27
Improve rotation Nob

Alecaddd:
Improve the positioning of the Rotate Nob and add a line to connect it tot he select effect.

## Steps to Test
- Create an item
- Rotate it, scale it, and change the canvas zoom.
- Confirm the rotation line is in a consistent position.

## Screenshots 
rotate-nob

- Better calculate the Rotate Nob position.
- Add polyline to connect Rotate Nob to select effect.


albfan (approved):
Not much fan of that rotate line, I would probably disconnect it from the item itself.

but looks good.

PR #502, opened 2021-02-28
WIP: Object Snapping

mbfraga:
This is my branch for the object snapping implementation.

Todo:
- [x] Generate snap grid
- [ ] Find the best snap matches for a selection
- [ ] Use snap matches to adjust an affine transform

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -55,6 +55,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     public Managers.ExportManager export_manager;
     public Managers.SelectedBoundManager selected_bound_manager;
     private Managers.NobManager nob_manager;
+    private Managers.SnapManager snap_manager;
     private Managers.HoverManager hover_manager;
 
     public bool ctrl_is_pressed = false;
@@ -82,6 +83,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         export_manager = new Managers.ExportManager (this);
         selected_bound_manager = new Managers.SelectedBoundManager (this);
         nob_manager = new Managers.NobManager (this);
+        snap_manager = new Managers.SnapManager (this);
         hover_manager = new Managers.HoverManager (this);
 
         window.event_bus.update_scale.connect (on_update_scale);
@@ -345,6 +347,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
             case EditMode.MODE_INSERT:
             case EditMode.MODE_SELECTION:
                 var selected_nob = nob_manager.selected_nob;
+                snap_manager.generate_snap_grid(selected_bound_manager.selected_items);
                 selected_bound_manager.transform_bound (event.x, event.y, selected_nob);
                 break;
 
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -355,6 +355,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                     // if snapping occured, we need to update the grid before painting
                     snap_manager.generate_snap_grid(selected_bound_manager.selected_items);
                 }
+                snap_manager.populate_decorators();
                 break;
 
             case EditMode.MODE_PANNING:
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -314,7 +314,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                 break;
 
             case EditMode.MODE_SELECTION:
-                snap_manager.reset();
+                snap_manager.reset ();
                 window.event_bus.detect_artboard_change ();
                 window.event_bus.detect_image_size_change ();
                 break;
@@ -349,7 +349,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
             case EditMode.MODE_SELECTION:
                 var selected_nob = nob_manager.selected_nob;
                 selected_bound_manager.transform_bound2 (event.x, event.y, selected_nob, snap_manager);
-                snap_manager.populate_decorators();
+                snap_manager.populate_decorators ();
                 break;
 
             case EditMode.MODE_PANNING:
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -348,7 +348,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
             case EditMode.MODE_INSERT:
             case EditMode.MODE_SELECTION:
                 var selected_nob = nob_manager.selected_nob;
-                selected_bound_manager.transform_bound2 (event.x, event.y, selected_nob, snap_manager);
+                selected_bound_manager.transform_bound (event.x, event.y, selected_nob, snap_manager);
                 snap_manager.populate_decorators ();
                 break;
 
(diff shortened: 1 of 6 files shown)


src/Lib/Canvas.vala
```diff
@@ -345,7 +348,14 @@ public class Akira.Lib.Canvas : Goo.Canvas {
...
                 var selected_nob = nob_manager.selected_nob;
-                selected_bound_manager.transform_bound (event.x, event.y, selected_nob);
+                snap_manager.generate_snap_matches(selected_bound_manager.selected_items);
+                selected_bound_manager.transform_bound (event.x, event.y, selected_nob, snap_manager.snap_match_data);
+
+                if (snap_manager.snap_match_data.wants_snap()) {
+                    // if snapping occured, we need to update the grid before painting
+                    snap_manager.generate_snap_grid(selected_bound_manager.selected_items);
+                }
+                snap_manager.populate_decorators();
```
Alecaddd, line 358: Wouldn't be better to have all of this directly in the `SelectBoundManager` file?
Instead of getting the selected items from that file and then passing the `snap_match_data` to it, could we do all of this directly inside the `transform_bound()` method?
The `SelectBoundManager` file has the `Canvas` reference directly available on construct so you can tap the SnapManager easily.

src/Lib/Managers/SnapManager.vala
```diff
@@ -0,0 +1,419 @@
...
+* along with Akira. If not, see <https://www.gnu.org/licenses/>.
+*
+* Authored by: Martin "mbfraga" Fraga <mbfraga@gmail.com>
+*/
+
+public class Akira.Lib.Managers.SnapManager : Object {
+    private const string STROKE_COLOR = "#ff0000";
+    private const double LINE_WIDTH = 2.0;
+    private const double DOT_RADIUS = 3.0;
+    private const double SENSITIVITY = 4.0;
```
Alecaddd, line 26: I wonder if the SENSITIVITY should also account for the canvas_scale.
Maybe if you zoomed in you don't need 4 px but less, and if you're zoomed out you want the items to be more snappy and have higher sensitivity.
Just a thought, we can investigate this later.

src/Utils/AffineTransform.vala
```diff
@@ -36,14 +36,15 @@ public class Akira.Utils.AffineTransform : Object {
...
         double event_x,
         double event_y,
         ref double initial_event_x,
         ref double initial_event_y
     ) {
         // Calculate the delta between the initial point and new mouse location.
-        var delta_x = fix_size (event_x - initial_event_x);
-        var delta_y = fix_size (event_y - initial_event_y);
+        var delta_x = fix_size (event_x - initial_event_x + data.horizontal_data.snap_offset());
+        var delta_y = fix_size (event_y - initial_event_y + data.vertical_data.snap_offset());
```
Alecaddd, line 47: Here, to prevent the jitter we can do something like this:
```PHP
        var delta_x = fix_size (event_x - initial_event_x);
        var delta_y = fix_size (event_y - initial_event_y);

        if (
            (delta_x < 0 && delta_x >= -data.horizontal_data.snap_offset()) ||
            (delta_x > 0 && delta_x <= data.horizontal_data.snap_offset())
        ) {
            delta_x = 0;
        }
        if (
            (delta_y < 0 && delta_y >= -data.vertical_data.snap_offset()) ||
            (delta_y > 0 && delta_y <= data.vertical_data.snap_offset())
        ) {
            delta_y = 0;
        }
```

The thing that is missing here is to check if the actual matrix X and Y of the moved item match exactly the X and Y locations of the guides.
If they don't match, we should subtract or add the delta between those coordinates to the matrix to make the item snap in place.

What do you think?

src/Lib/Canvas.vala
mbfraga: Yeah, this could be moved for sure. I wasn't sure what would be better.

src/Lib/Managers/SnapManager.vala
mbfraga: Yeah, I do think it should be have like that. but that does mean the match calculation needs to be updated on zoom in / out while these operations are happening. But in general how we hook up snapping to the transform operations needs to be thought out a bit. Atm for move you can calculate the snaps from the selection. For other transforms I think only a single snap should be tracked (e.g., which nob is being moved).

src/Utils/AffineTransform.vala
mbfraga: I think this is pretty close. I actually added some useful data in my current branch (need to push) to help with this. I got something very similar working:
```
        if (data.horizontal.type == SnapManager.MatchType.FUZZY) {
            delta_x = data.horizontal.snap_offset();
        }
        else if (data.horizontal.type == SnapManager.MatchType.EXACT) {
            delta_x = 0;
        }
```

This snaps correctly and stays in place...but the issue is that it never snaps out (black hole effect!). I need some kind of reference point to the original starting point of the drag--otherwise it's impossible to restore the correct position of the object relative to the mouse.

The downside to an approach like this, compared to my suggestion of a shadow of the selection, is that if we unsnap in a move tick, then an immediate snap would not trigger. It may not matter so much in practice...maybe? i.e., there is one frame where a potential snap may be missed.

mbfraga: I added some information to make this work. I pushed the changes. I think working this relative to the previous positions will always be really flaky.

The snap calculation needs to be based on the theoretical position of the selection to allow the system to be deterministic.

src/Utils/AffineTransform.vala
```diff
@@ -45,6 +48,32 @@ public class Akira.Utils.AffineTransform : Object {
...
+        if (data.horizontal.wants_snap() && delta_x.abs() <= 4) {
+            if (data.horizontal.type == SnapManager.MatchType.FUZZY) {
+                delta_x = data.horizontal.snap_offset();
+                snap_x = true;
+            }
+            else {
+                delta_x = 0;
+                snap_x = true;
+            }
+        }
```
Alecaddd, line 63: Interesting, I think this is looking like a good solid approach.
We can rewrite this to be a bit shorter like:
```PHP
if (data.horizontal.wants_snap() && delta_x.abs() <= 4) {
    delta_x = data.horizontal.type == SnapManager.MatchType.FUZZY ? data.horizontal.snap_offset() : 0;
    snap_x = true;
}
```
And same thing for the vertical condition.

src/Utils/AffineTransform.vala
```diff
@@ -59,8 +88,12 @@ public class Akira.Utils.AffineTransform : Object {
...
         }
 
-        initial_event_x = event_x;
-        initial_event_y = event_y;
+        if (!snap_x) {
+            initial_event_x = event_x;
+        }
+        if (!snap_y) {
+            initial_event_y = event_y;
+        }
```
Alecaddd, line 96: I didn't try the updated code, so apologies if my question is silly, but wouldn't this prevent the delta from returning the real mouse acceleration from the user?
We kinda want the snap to be a bit magnetic unless the user moves quickly out of range.
Did you put this in place otherwise the snap is locked?

src/Utils/AffineTransform.vala
Alecaddd: Also, we should get the `SENSITIVITY` from the SnapManager instead of manually writing it, so we're safe if that value changes for whatever reason.

mbfraga:
I think this PR is fully encapsulated in the other PR, so it's worth closing.

Alecaddd:
Closed in favor of #507

PR #504, opened 2021-03-03
Use absolute positions in move operation

mbfraga:
Snapping requires more robust movement of items. This PR movees
the move_from_event to the SelectedBoundManager.

The SelectedBoundManager stores drag data regarding the start
position of the event and of the item in order to set new positions
based on those instead of relative to an accumulator as before.

This will allow for much nicer snapping

## Steps to Test
Dragging items should work as before.

### DIFF
--- a/src/Lib/Managers/SelectedBoundManager.vala
+++ b/src/Lib/Managers/SelectedBoundManager.vala
@@ -43,6 +43,13 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
     private double initial_width;
     private double initial_height;
 
+    // drag move state
+    private double initial_drag_press_x;
+    private double initial_drag_press_y;
+    private bool   initial_drag_item_registered = false;
+    private double initial_drag_item_x;
+    private double initial_drag_item_y;
+
     public SelectedBoundManager (Akira.Lib.Canvas canvas) {
         Object (
             canvas: canvas,
@@ -65,6 +72,10 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
     public void set_initial_coordinates (double event_x, double event_y) {
         initial_event_x = event_x;
         initial_event_y = event_y;
+
+        initial_drag_press_x = event_x;
+        initial_drag_press_y = event_y;
+        initial_drag_item_registered = false;
 
         if (selected_items.length () == 1) {
             var selected_item = selected_items.nth_data (0);
@@ -91,10 +102,7 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
 
         switch (selected_nob) {
             case Managers.NobManager.Nob.NONE:
-                Utils.AffineTransform.move_from_event (
-                    selected_item, event_x, event_y,
-                    ref initial_event_x, ref initial_event_y
-                );
+                move_from_event ( selected_item, event_x, event_y );
                 break;
 
             case Managers.NobManager.Nob.ROTATE:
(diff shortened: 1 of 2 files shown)


src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -43,6 +43,13 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
     private double initial_width;
     private double initial_height;
 
+    // drag move state
+    private double initial_drag_press_x;
+    private double initial_drag_press_y;
+    private bool   initial_drag_item_registered = false;
```
Alecaddd, line 49: we don't align attribute names, always use 1 space after the type declaration.
Also, what about calling this `initial_drag_registered`?

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -43,6 +43,13 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
     private double initial_width;
     private double initial_height;
 
+    // drag move state
```
Alecaddd, line 46: Better be a bit more explicit with comments.
Something like "Attributes to keep track of the mouse dragging coordiantes."

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -66,6 +73,10 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
         initial_event_x = event_x;
         initial_event_y = event_y;
 
+        initial_drag_press_x = event_x;
+        initial_drag_press_y = event_y;
+        initial_drag_item_registered = false;
```
Alecaddd, line 78: I think we need a comment here to explain why we're setting this variable to false when a new event is triggered.

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -91,10 +102,7 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
 
         switch (selected_nob) {
             case Managers.NobManager.Nob.NONE:
-                Utils.AffineTransform.move_from_event (
-                    selected_item, event_x, event_y,
-                    ref initial_event_x, ref initial_event_y
-                );
+                move_from_event ( selected_item, event_x, event_y );
```
Alecaddd, line 105: No spaces within the parenthesis.

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -310,4 +318,48 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+            initial_drag_item_registered = true;
+            initial_drag_item_x = item.bounds.x1;
+            initial_drag_item_y = item.bounds.y1;
+        }
+
+        // Keep reset and delta values for future adjustments.
+
+        // Calculate values needed to reset to the original position
+        var reset_x = item.bounds.x1 - initial_drag_item_x;
+        var reset_y = item.bounds.y1 - initial_drag_item_y;
```
Alecaddd, line 337: Better avoid using the `item.bounds` since those don't account for grouped items like the Artboard, which the bounds are not representative of the actual size of the item (eg. if an Artboard child is outside the artboard boundaries it expands the Artboard bounds).
Use the `item.transform.x` and `item.transform.y` which are the coordinates coming from the `Transform` component, which takes care of properly returning the correct values.

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -310,4 +318,48 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+
+        item.set_transform (matrix);
+
+        // Any adjustments like snapping would be computed here, and the item would
+        // then be translated again.
+
+        // If the item is an Artboard, move the label with it.
+        if (item is Lib.Items.CanvasArtboard) {
+            ((Lib.Items.CanvasArtboard) item).label.translate (first_move_x, first_move_x);
+        }
```
Alecaddd, line 363: This doesn't work.
Create an Artboard and move it around, you'll see how the label is a bit wonky.

Alecaddd (changes requested):

src/Lib/Managers/SelectedBoundManager.vala
Alecaddd: Ah, I found the error. You're using `first_move_x` also for the Y :P

Alecaddd (approved):
Looks good, thank!

PR #505, opened 2021-03-03
Handle multiple colors

Alecaddd:
Implement the alpha mixing of multiple fill and border colors.

## Steps to Test
- Create a shape
- Add multiple colors and fills
- Play with the alpha of the extra colors

## Screenshots 
color-blending

~The color gets reset once the item is selected, which doesn't make sense.~ Fixed!
For now we only handle 1 border size, so we get the bigger and go with it.
In the future we will probably implement child items just to handle different border sizes and also different offset once implemented in goocanvas.

- Blend colors by their alpha channel for fills and borders.
- Use the color picker icon for the color picker button.
- Move the color picker close to the color button and style them to make them related.

### DIFF
--- a/src/Utils/Color.vala
+++ b/src/Utils/Color.vala
@@ -90,13 +90,12 @@ public class Akira.Utils.Color : Object {
         // Simple blend using Alpha channels. In the future we will support
         // different blending modes.
         double alpha = 1 - (1 - added_color.alpha) * (1 - base_color.alpha);
+        double ar1 = added_color.alpha / alpha;
+        double ar2 = (1 - added_color.alpha) / alpha;
 
-        double red = added_color.red * added_color.alpha /
-            alpha + base_color.red * base_color.alpha * (1 - added_color.alpha) / alpha;
-        double green = added_color.green * added_color.alpha /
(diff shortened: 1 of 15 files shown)


Alecaddd:
Pinging @mbfraga to see if he has time to test this.

src/Utils/Color.vala
```diff
@@ -69,4 +72,38 @@ public class Akira.Utils.Color : Object {
...
+        // the only one visible.
+        if (added_color.alpha == 1.0 || base_color.alpha == 0) {
+            return added_color;
+        }
+
+        // Simple blend using Alpha channels. In the future we will support
+        // different blending modes.
+        double alpha = 1 - (1 - added_color.alpha) * (1 - base_color.alpha);
+
+        double red = added_color.red * added_color.alpha /
```
mbfraga, line 94: You can stroe the alpha rations for performance.

double ar1 = added_color.alpha / alpha;
doouble ar2 = (1 - added_color.alpha) / alpha

mbfraga (approved):

src/Utils/Color.vala
Alecaddd: Ah, good catch.

PR #510, opened 2021-03-07
Add transform boundaries

Alecaddd:
Implement the `x1, x2, y1, y2` attributes to the `Transform` component, which are used to return the correct boundaries of an item excluding the stroke width or masked child items.

## Steps to Test
- Create multiple items and Artboards with items partially outside.
- Move the items around and see how the snaps appear in the proper position.
- Change the stroke width to confirm that the snaps ignore it. 

- Implement new `Transform` coordiantes.
- Keep the usage of bounds for the export selection in order to include borders during export.
- Fix the snapping for Artboards with items partially outside the Artboard's boundaries.
- Move the Artboard's label when a scale event affects the origin coordinates.

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -504,8 +504,8 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         if (show) {
             ghost = new Goo.CanvasRect (
                 null,
-                item.transform.x, item.transform.y,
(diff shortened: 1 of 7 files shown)


mbfraga (approved):

mbfraga:
Tested several features (including snapping), and code looks fine.

PR #519, opened 2021-03-11
Use absolute scale when resizing

mbfraga:
This PR addresses scaling using absolute position of the mouse rather than increments. This is both a much more solid and reliable approach, as well as allowing for future integration with snapping.

This PR also enables resizing beyond the 0,0 point of the bounding box.

Tries to force snapping to pixels. If original has subpixel position or size, will adjust to snap to pixel grid.

## Steps to Test
Resize object using all nobs, including different rotations.

Methods take a lot of arguments, not sure what is the philosophical approach to this.

Small detail is that the scaling is relative to mouse down position rather than the precise nob position. This is not a regression, but something to be addressed in the future--specially for snapping to work correctly.


src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -123,14 +124,7 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
             default:
-                Utils.AffineTransform.scale_from_event (
-                    selected_item,
-                    selected_nob,
-                    event_x, event_y,
-                    ref initial_event_x, ref initial_event_y,
-                    ref delta_x_accumulator, ref delta_y_accumulator,
-                    initial_width, initial_height
-                );
+                scale_from_event ( selected_item, selected_nob, event_x, event_y );
```
Alecaddd, line 127: nit: no blank space inside parenthesis

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -421,4 +415,76 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+        // values even if the item is rotated.
+        canvas.convert_to_item_space (item, ref rel_event_x, ref rel_event_y);
+        canvas.convert_to_item_space (item, ref rel_press_x, ref rel_press_y);
+
+        // Calculate the change based on the event.
+        var delta_x = rel_event_x - rel_press_x;
+        var delta_y = rel_event_y - rel_press_y;
+
+        double item_x = item.bounds.x1;
+        double item_y = item.bounds.y1;
```
Alecaddd, line 448: use the `transform` component instead of the bounds.

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -421,4 +415,76 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+
+        bool ratio_locked = canvas.ctrl_is_pressed || item.size.locked;
+
+        // these values will be populated
+        double inc_width = 0;
+        double inc_height = 0;
+        double inc_x = 0;
+        double inc_y = 0;
+        bool h_flip = false;
+        bool v_flip = false;
```
Alecaddd, line 459: I think we can safely drop these as we don't want to actually flip the item on resize.

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -421,4 +415,76 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+
+        Utils.AffineTransform.calculate_size_adjustments (
+            selected_nob,
+            initial_width,
+            initial_height,
+            delta_x,
+            delta_y,
+            initial_width / initial_height,
```
Alecaddd, line 467: The size ratio can be accessed from the `Size` component without needing of calculation.
`item.size.ratio`

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -421,4 +415,76 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+            ref inc_y,
+            ref inc_width,
+            ref inc_height,
+            ref h_flip,
+            ref v_flip
+        );
+
+        var reset_width = item.size.width - initial_width;
+        var reset_height = item.size.height - initial_height;
+
```
Alecaddd, line 480: nit: remove extra blank line

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -421,4 +415,76 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+
+        Cairo.Matrix new_transform;
```
Alecaddd, line 482: nit: I prefer to call this `new_matrix` since we have the `item.transform` component.
It's easier to read and to distinguish.

src/Utils/AffineTransform.vala
```diff
@@ -27,10 +27,262 @@ public class Akira.Utils.AffineTransform : Object {
...
+
+        bool pure_v = (nob == NobManager.Nob.TOP_CENTER || nob == NobManager.Nob.BOTTOM_CENTER);
+        bool pure_h = (nob == NobManager.Nob.RIGHT_CENTER || nob == NobManager.Nob.LEFT_CENTER);
+
+        // handle vertical adjustment
+        if (NobManager.is_top_nob (nob)) {
+            inc_height = -delta_y;
+            local_y_adj = delta_y;
+        }
+        else if (NobManager.is_bot_nob (nob)) {
```
Alecaddd, line 87: nit: put the else inline the closing curly bracket.

Alecaddd (changes requested):
This is great and it works like a charm.
Just a few marginal fixes for the code, but overall this is awesome.

Two things to consider:
1. When resizing an Artboard below 1px, the lable should be resized back on top of it.
2. When resizing an Artboard from the TOP or LEFT nobs, the label doesn't properly follow.

src/Lib/Managers/SelectedBoundManager.vala
mbfraga: I preferred using the initial values for accuracy, otherwise resize weirdness could cause issues? I am not sure if it will, but this feels a bit more robust. But I can change it if you think it's better.

Alecaddd: That's okay with me.

Alecaddd:
@mbfraga this looks perfect IMO. Do you need to do anything else or am I good to merge it?

Alecaddd:
Mhhh...there's something wrong with the code but I can't find where.
Screenshot from 2021-03-16 19-15-24

As you can see form this screenshot, sometimes the items are created with the border aligned to the grid and not the actual bounds which should only account for the fill/body of the item and ignore the border.

Also, when creating a new item with the visible pixel grid, the grid doesn't get properly updated to always be on top.
This doesn't happen on the `master` branch.

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -422,4 +420,88 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+        Lib.Items.CanvasItem item,
+        Managers.NobManager.Nob selected_nob,
+        double event_x,
+        double event_y
+    ) {
+        if (!initial_drag_registered) {
+            item.get_transform (out initial_item_transform);
+            initial_drag_registered = true;
+            initial_drag_item_x = item.bounds.x1;
+            initial_drag_item_y = item.bounds.y1;
```
Alecaddd, line 434: This is wrong, it should be `item.transform.x1` and `item.transform.y1`.

src/Lib/Managers/SelectedBoundManager.vala
```diff
@@ -422,4 +420,88 @@ public class Akira.Lib.Managers.SelectedBoundManager : Object {
...
+            var dummy_x = 1.0;
+            var dummy_y = 1.0;
+            initial_item_transform.transform_distance (ref dummy_x, ref dummy_y);
+            if (Utils.AffineTransform.epsilon_equals_d (1.0, GLib.Math.fabs (dummy_x), double.EPSILON * 10)
+                && Utils.AffineTransform.epsilon_equals_d (1.0, GLib.Math.fabs (dummy_y), double.EPSILON * 10)) {
+                scale_item_x_adj = Utils.AffineTransform.fix_size (initial_drag_item_x) - initial_drag_item_x;
+                scale_item_y_adj = Utils.AffineTransform.fix_size (initial_drag_item_y) - initial_drag_item_y;
+                initial_width = Utils.AffineTransform.fix_size (initial_width);
+                initial_height = Utils.AffineTransform.fix_size (initial_height);
+            }
```
Alecaddd, line 450: This is what causing the shift in the position of the item and the wrong alignment.

Alecaddd (changes requested):

src/Lib/Managers/SelectedBoundManager.vala
Alecaddd: I think this should be wrapped inside a condition like:
`if (item.rotation != null && item.rotation.rotation != 0) {`
So we avoid running it on item creation, or for those items that don't support rotation, like artboards.

mbfraga: rotation != 0 is not the correct condition, since it applies to 90, 180, 270 or 360

Alecaddd: So, could we do this? 
`if (item.rotation != null && GLib.Math.fmod (item.rotation.rotation, 90) == 0)`

PR #520, opened 2021-03-12
Implement multi select

Alecaddd:
Implement items multiselect via Canvas clicks and allow moving the selected array of items.

## Steps to Test
- Create multiple items.
- Select them by holding down the <kdb>SHIFT</kbd> key.
- Move them around.
- Click on a selected item by holding down the <kdb>SHIFT</kbd> key to remove it from the selection.
- Click on a selected item to deselect all the others.

## Screenshots 
group-move

Scale and rotation don't work, but I'm probably gonna wait for the PRs from @mbfraga before implementing those.

~The initial dragging position is collected into an `HashMap`. I'm not sure that's the best approach as we keep fetching the coordinates from the map at every movement, and that might get expensive when dealing with many items.
We use the `item.name.id` as HashMap key, but currently that ID is not really unique. I need to change that.~ *Ignore this as a better method was implemented*.

## This PR fixes/implements the following bugs/features:
- Rename the `Transform` component to `Coordinates` to avoid overriding the built-in `transform` Matrix abstract attribute of the Goo.CanvasItem object.
- Reduce the number of attributes in the `Coordinates` component.
- Properly draw the select effect based on the amount of selected items.
- Enable layer multiselection in the layers panel.

### DIFF
--- a/src/FileFormat/JsonObject.vala
+++ b/src/FileFormat/JsonObject.vala
@@ -111,8 +111,8 @@ public class Akira.FileFormat.JsonObject : GLib.Object {
 
         if (item.transform != null) {
             var transform = new Json.Object ();
-            transform.set_double_member ("x", item.transform.x);
-            transform.set_double_member ("y", item.transform.y);
+            transform.set_double_member ("x", item.coordinates.x);
+            transform.set_double_member ("y", item.coordinates.y);
 
             components.set_object_member ("Transform", transform);
         }
--- a/src/Layouts/Partials/Artboard.vala
+++ b/src/Layouts/Partials/Artboard.vala
@@ -495,20 +495,24 @@ public class Akira.Layouts.Partials.Artboard : Gtk.ListBoxRow {
 
             if (active) {
                 button_locked.get_style_context ().add_class ("show");
+                // Disable any pointer events for a locked item.
+                model.pointer_events = Goo.CanvasPointerEvents.NONE;
+
+                // Let the UI know that this item was locked.
+                window.event_bus.item_locked (model);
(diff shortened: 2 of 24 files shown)


Alecaddd:
I think I can improve this by storing the initial drag coordinates of an item when is first added to the list of selected items, and then update those coordinates after a button release.
In doing so we can remove that storing from the loop and avoid repeatedly calling that condition.

mbfraga:
I think for the item move, it should be handled somewhat differently. I am attaching a git diff of what I think could be done instead. Feel free to do whatever you deem necessary. I wrote it quite quickly, but I think I still am doing generally the right things. Perhaps we need to do a check to see if selected_items is empty...and maybe do a different reference point for the move. Otherwise it should be mostly ok.

Github will not take a .diff...so adding it as .txt

[multi_sel_snapping.txt](https://github.com/akiraux/Akira/files/6133631/multi_sel_snapping.txt)

Alecaddd:
I think this might be a good point to stop and request a full review.
There are a lot of things that needs to be done, and a million different improvements to remove duplicated code all over the place in the Layers panel, but better do those things slowly in follow up PRs.

Things to do for the future:
- The transform panel needs to account for multiple items and their coordinates.
- The fills and borders panel needs to group identical colors into single widgets.
- The layers panel has too much repeated code and a lot of methods should be delegated to components.
- Rotation and scale for multiple items don't work.

Alecaddd:
Converting this to a draft as some performance improvements to other aspects of the UI are necessary before landing this.

Alecaddd:
closing this as many changes happened in the code base and a lot of thins in the PR already landed.
Will recreate a smaller and more targeted PR to implement multi select.

PR #527, opened 2021-03-23
Slight rework to nob manager

mbfraga:
The purpose is twofold:

1. Simplify the laying out of nobs so that it's easier to follow the
   logic

2. Make nobs be placed precisely at the right positions relative to a
   selection, and be able to generate nobs from any arbitrary selection
   of items--regardless of whether or not it is part of the selection.

## Steps to Test
Nobs should be placed just as before, and on rotation they should track the right rotation of the object.

### DIFF
--- a/src/Lib/Managers/NobManager.vala
+++ b/src/Lib/Managers/NobManager.vala
@@ -54,15 +54,21 @@ public class Akira.Lib.Managers.NobManager : Object {
     private Goo.CanvasRect? select_effect;
     private Goo.CanvasItemSimple[] nobs = new Goo.CanvasItemSimple[9];
     private Goo.CanvasPolyline? rotation_line;
-    private Goo.CanvasBounds select_bb;
-    private double top;
-    private double left;
-    private double width;
-    private double height;
-    private double nob_size;
+
+    // values in canvas coordinates
+    private double top_left_x;
+    private double top_left_y;
+    private double width_offset_x;
+    private double width_offset_y;
+    private double height_offset_x;
+    private double height_offset_y;
+    private double bb_width;
+    private double bb_height;
+    Cairo.Matrix bb_matrix;
 
     // Tracks if an artboard is part of the current selection.
     private bool is_artboard;
+    private bool nobs_constructed = false;
 
     // If the effect needs to be created or it's only a value update.
     private bool create { get; set; default = true; }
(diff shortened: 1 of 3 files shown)


src/Lib/Managers/NobManager.vala
```diff
@@ -74,348 +80,433 @@ public class Akira.Lib.Managers.NobManager : Object {
...
-        foreach (var item in selected_items) {
-            bb_left = double.min (bb_left, item.coordinates.x1);
-            bb_top = double.min (bb_top, item.coordinates.y1);
-            bb_right = double.max (bb_right, item.coordinates.x2);
-            bb_bottom = double.max (bb_bottom, item.coordinates.y2);
+        if (items.length () == 1) {
+            var item = items.first ().data;
+            item.get_transform (out matrix);
+
+            Cairo.Matrix nob_matrix = matrix;
```
Alecaddd, line 154: Since we're doing an early return, we don't need to declare the `nob_matrix` and we can use the existing. `matrix` variable.

src/Lib/Managers/NobManager.vala
```diff
@@ -74,348 +80,433 @@ public class Akira.Lib.Managers.NobManager : Object {
...
+        double width_offset_x,
+        double width_offset_y,
+        double height_offset_x,
+        double height_offset_y,
+        ref double pos_x,
+        ref double pos_y
+    ) {
+        switch (nob_name) {
+        case Nob.NONE:
+            break;
```
Alecaddd, line 217: Can we put this attached to the default at the bottom?
```
case Nob.NONE:
case default:
  break;
```
Does that work?

src/Lib/Managers/NobManager.vala
```diff
@@ -74,348 +80,433 @@ public class Akira.Lib.Managers.NobManager : Object {
...
+        case Nob.LEFT_CENTER:
+            pos_x = top_left_x + height_offset_x / 2.0;
+            pos_y = top_left_y + height_offset_y / 2.0;
+            break;
+        default:
+            break;
         }
+    }
+
```
Alecaddd, line 254: nit: remove extra empty line.

Alecaddd (changes requested):
Looks good, great improvements.
I left some requests for changes that are mostly cosmetic.

src/Lib/Selection/Nob.vala
```diff
@@ -26,40 +26,48 @@ public class Akira.Lib.Selection.Nob : Goo.CanvasRect {
...
     public void set_rectangle () {
         x = 0;
         y = 0;
 
-        update_size ();
-
-        radius = handle_id == 8 ? nob_size : 0;
-
-        set ("radius-x", radius);
-        set ("radius-y", radius);
```
Alecaddd, line 62: would be possible to keep this?
The rotation nob should be a circle

PR #531, opened 2021-03-24
A very hacky undo redo implementation

### LINKED ISSUE #263: Implement History Manager
Let's kick-off the conversation on how to implement an History Manager to handle `undo` and `redo` actions.

** Memory management **
The first approach is going the standard route of implementing a memory based history manager, and giving users the ability to define the memory usage limit.

** Version control **
The other approach we could consider is to save the various actions in a local history file if the user hasn't saved the current project, or save it directly inside the file in case it exists.
This will give us the flexibility to have a potential infinite history without taxing on memory management.
If we decide to go this route, we will need to implement a manual "clear file history" for the user in order to better control the file size. Even adding an option like "Clear file history on shutdown" might be useful.

Thoughts?

mbfraga:
Implement a very hacky undo redo using existing serialization methods

Should serve as a proof of concept...and even if it gets implemented like this (with major cleanup and fixes), it would serve as a way to start adding the right events in the right place. And also see any reworks that may need to happen.

Super hacks--if this is the direction initial undo/redo takes, then serialization / deserialization needs to be refactored correctly.

- Fixes #263 - sorta, kinda, not really--but it's a working prototype that may be good to merge (after time spent to do it the right way) if only to test things

### DIFF
--- a/src/Lib/Managers/UndoManager.vala
+++ b/src/Lib/Managers/UndoManager.vala
@@ -1,0 +1,107 @@
+
+
+
+
+
+
+
+public class Akira.Lib.Managers.UndoManager : Object {
+
+    private GLib.Queue<string> undos;
+    private GLib.Queue<string> redos;
+
+    public UndoManager () {
+        undos = new GLib.Queue<string>();
+        redos = new GLib.Queue<string>();
+    }
+
+
+    public void add_undo(Akira.Lib.Canvas canvas) {
+        redos.clear ();
+        inner_add_undo(Akira.FileFormat.JsonContent.serialize_canvas(canvas));
+        debug("%d %d", (int)undos.get_length(), (int)redos.get_length());
+    }
+
+    public void apply_undo(Akira.Lib.Canvas canvas) {
+        if (undos.get_length () == 0)  {
+            debug("undo empty");
+            return;
+        }
+
+        var old_undo = undos.pop_head();
+        inner_add_redo(old_undo);
+
+        try {
+            var parser = new Json.Parser ();
+            parser.load_from_data (old_undo);
+            var obj = parser.get_root ().get_object ();
+
+            clear_canvas(canvas);
+            Akira.FileFormat.JsonLoader.inner_load_content(canvas, obj);
+
+        } catch (Error e) {
+            debug("failed to read undo");
+            return;
+        }
+
+        debug("%d %d", (int)undos.get_length(), (int)redos.get_length());
+    }
+
+    private void inner_add_undo(string undo_to_add) {
+      undos.push_head(undo_to_add);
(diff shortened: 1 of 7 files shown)


src/Lib/Managers/UndoManager.vala
```diff
@@ -0,0 +1,107 @@
+
+
+
+
+
+
+
+public class Akira.Lib.Managers.UndoManager : Object {
```
Alecaddd, line 8: I'd call this HistoryManager since it will handle both undos and redos.

src/Lib/Managers/UndoManager.vala
```diff
@@ -0,0 +1,107 @@
...
+        foreach (var item in item_manager.artboards) {
+            to_delete.push_head(item);
+        }
+
+        foreach (var item in item_manager.images) {
+            to_delete.push_head(item);
+        }
+
+        while (to_delete.get_length () != 0) {
+            canvas.window.event_bus.request_delete_item (to_delete.pop_head());
```
Alecaddd, line 103: We don't need to collect the images as the request_delete_item will remove those automatically.
Just artboards and free items.
Here we could also make this method async and use multi threading to prevent UI freeze when clearing a very busy canvas.

src/Lib/Managers/UndoManager.vala
```diff
@@ -0,0 +1,107 @@
...
+        var old_undo = undos.pop_head();
+        inner_add_redo(old_undo);
+
+        try {
+            var parser = new Json.Parser ();
+            parser.load_from_data (old_undo);
+            var obj = parser.get_root ().get_object ();
+
+            clear_canvas(canvas);
+            Akira.FileFormat.JsonLoader.inner_load_content(canvas, obj);
```
Alecaddd, line 40: I'm not sure about this, it seems a bit extreme.
With this if we move an item and we want to revert to its previous position, we're clearing the canvas and reloading everything. That seems expensive.

mbfraga: If clearing and reloading is expensive, we have bigger issues than undo/redo. There is no reason the view should be expensive to regenerate.

I'm open to new ideas, but am cautious about an alternative that doesn't require cleaning the slate. They tend to be rather brittle and crash or artifact prone approaches.

mbfraga: To be clear, the current implementation of clearing and reloading is not efficient, but that's a different story.

Alecaddd:
@mbfraga wanna close this as it's gonna be invalid with the components refactor you're doing?

Alecaddd:
@mbfraga can we close this since it doesn't fit with the component refactor you're doing?

PR #534, opened 2021-03-30
Add Size State Manager

Alecaddd:
This PR adds a `SizeManager` in order to properly handle sizing for multiple selected items when interacting with the Transform Panel. It also updates the `CoordinatesManager` in order to follow the same approach, as well as removing unnecessary methods, and reducing the calls to update attributes when not necessary.

## Steps to Test
Create a bunch of items and try to move them around and resize them exclusively via the Transform Panel fields.
**DISCLAIMER**: The Rotation field is not properly hooked yet, so rotating via the Transform Panel doesn't properly update the X & Y coordinates like it happens when rotating from the canvas.

- Implement remaining managers for rotation, flipping, and opacity.

- Limit the number of calls to update attributes by only triggering the update of each single field when edited.
- Remove attributes from some events to simplify the code.
- Translate and Resize via the Transform Panel is done via absolute values instead of relative, in preparation of multi selection.

### DIFF
--- a/src/Lib/Components/Coordinates.vala
+++ b/src/Lib/Components/Coordinates.vala
@@ -23,6 +23,7 @@
  * Coordinates component to keep track of the item's initial coordinates.
  */
 public class Akira.Lib.Components.Coordinates : Component {
+    // Returns the bounds.x1 value updated if the item is inside an artboard.
     public double x {
         get {
             // If the item is an artboard we need to get the bounds of the background since
@@ -43,6 +44,7 @@ public class Akira.Lib.Components.Coordinates : Component {
         }
     }
 
+    // Returns the global bounds.x1 value.
     public double x1 {
         get {
             // If the item is an artboard we need to get the bounds of the background since
@@ -56,6 +58,7 @@ public class Akira.Lib.Components.Coordinates : Component {
         }
     }
 
+    // Returns the global bounds.x2 value.
     public double x2 {
         get {
             // If the item is an artboard we need to get the bounds of the background since
@@ -69,6 +72,7 @@ public class Akira.Lib.Components.Coordinates : Component {
         }
     }
 
+    // Returns the bounds.y1 value updated if the item is inside an artboard.
     public double y {
         get {
             // If the item is an artboard we need to get the bounds of the background since
@@ -89,6 +93,7 @@ public class Akira.Lib.Components.Coordinates : Component {
         }
     }
 
+    // Returns the global bounds.y1 value.
     public double y1 {
         get {
             // If the item is an artboard we need to get the bounds of the background since
@@ -102,6 +107,7 @@ public class Akira.Lib.Components.Coordinates : Component {
         }
     }
 
+    // Returns the global bounds.y2 value.
     public double y2 {
         get {
             // If the item is an artboard we need to get the bounds of the background since
--- a/src/Lib/Components/Coordinates.vala
+++ b/src/Lib/Components/Coordinates.vala
@@ -23,7 +23,7 @@
  * Coordinates component to keep track of the item's initial coordinates.
  */
 public class Akira.Lib.Components.Coordinates : Component {
-    // Returns the bounds.x1 value updated if the item is inside an artboard.
+    // LEFT item position, converted to artboard's coordinates if the item is inside an artboard.
     public double x {
         get {
             // If the item is an artboard we need to get the bounds of the background since
@@ -44,7 +44,7 @@ public class Akira.Lib.Components.Coordinates : Component {
         }
     }
 
-    // Returns the global bounds.x1 value.
+    // Global LEFT item position relative to the Canvas, ignoring potential parent's artboard.
     public double x1 {
         get {
             // If the item is an artboard we need to get the bounds of the background since
@@ -58,7 +58,8 @@ public class Akira.Lib.Components.Coordinates : Component {
         }
     }
 
-    // Returns the global bounds.x2 value.
+    // Global RIGHT (LEFT + WIDTH) item position relative to the Canvas,
+    // ignoring potential parent's artboard.
     public double x2 {
         get {
             // If the item is an artboard we need to get the bounds of the background since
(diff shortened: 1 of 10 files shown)


src/Lib/Managers/NobManager.vala
```diff
@@ -66,7 +66,13 @@ public class Akira.Lib.Managers.NobManager : Object {
     private double height_offset_y;
     private double bb_width;
     private double bb_height;
-    Cairo.Matrix bb_matrix;
+    private Cairo.Matrix bb_matrix;
+
+    // Values for the Transform Panel fields.
+    public double selected_x;
+    public double selected_y;
+    public double selected_width;
```
mbfraga, line 74: what is the difference between bbwidth / selected_width?

src/StateManagers/CoordinatesManager.vala
```diff
@@ -76,26 +81,24 @@ public class Akira.StateManagers.CoordinatesManager : Object {
...
-        double item_y = item.coordinates.y;
+    private void on_init_state_coords () {
+        do_update = false;
 
-        // Interrupt if no value has changed.
-        if (item_x == x && item_y == y) {
-            return;
-        }
+        // Get the item X & Y coordinates.
+        x = canvas.nob_manager.selected_x;
```
mbfraga, line 93: why not just request the position here? There is no real performance gain here.

src/StateManagers/CoordinatesManager.vala
```diff
@@ -105,63 +108,68 @@ public class Akira.StateManagers.CoordinatesManager : Object {
...
-     * therefore we set the d_update to false to prevent updating  the selected
-     * items Cairo transform.
+     * Reset the coordinates to get the newly updated coordinates from the selected items.
+     * This method is called when items are moved from the canvas, so we only need to update
+     * the X and Y values for the Transform Panel without triggering the update_items_*().
      */
-    private void on_reset_state_coords (Lib.Items.CanvasItem item) {
-        do_update = false;
+    private void on_reset_state_coords () {
+        on_init_state_coords ();
```
mbfraga, line 116: yes! let's remove all cases where an event takes items...it's asking for trouble!

src/Window.vala
```diff
@@ -33,6 +33,7 @@ public class Akira.Window : Gtk.ApplicationWindow {
     public Akira.Utils.Dialogs dialogs;
 
     public Akira.StateManagers.CoordinatesManager coords_manager;
+    public Akira.StateManagers.SizeManager size_manager;
```
mbfraga, line 36: what managers go in the window and which go in the canvas is really unclear--and was very confusing to me.

src/StateManagers/SizeManager.vala
```diff
@@ -0,0 +1,152 @@
...
+        get {
+            return _width != null ? _width : 0;
+        }
+        set {
+            if (value == _width) {
+                return;
+            }
+
+            _width = Utils.AffineTransform.fix_size (value);
+            update_items_width ();
```
mbfraga, line 48: I don't really love this get set vala mechanic, I think it's flaky. But thats' just a personal opinion and doesn't matter much. What does matter is to make sure these things never have side effects or events attached to them. I should be able to change values without causing events, etc. Not saying it's teh case here, but want to make sure.

mbfraga (changes requested):
Some comments and uncertainties. Generally looks ok. Tried it and most things seemed to work well.

src/Lib/Managers/NobManager.vala
Alecaddd: For now there's no difference at all, but I created that because it's specific to the Transform Panel and I don't know if the bbwidth will remain as it is.
I can definitely drop it if you think it's redundant and you don't think the bbwidth and bbheight will drastically change their purpose (who knows really, tho).
I can write some comments to be sure those variables are referencing the transform panel.

src/StateManagers/CoordinatesManager.vala
Alecaddd: Do you mean by calling the static method?
Wouldn't that be a bit of an overkill since I only need 2 variables?
I'm definitely not against it.

src/Window.vala
Alecaddd: That depends if you want the methods of those managers to affect the whole application and all its opened instances (windows) or if the methods should affect only the currently focused window.

Alecaddd: That depends on what the methods of those managers do.
The Size manager and coordinates manager affect and interact with the Transform Panel, which is a UI element outside the Canvas.
The Export Manager, or Nob manager, don't interact with any part of the UI other than the items in the Canvas, therefore they're bounded to the Canvas.
I'm more than happy to implement a better and more consistent approach.

src/StateManagers/SizeManager.vala
Alecaddd: That's a good point.
Do you have any suggestion for this situation?
Basically the `width` attribute is binded to the Transform Panel Width field, so every time the user interact with that field, the setter is called and triggers the update of the selected items.
I guess we could bind that field to the width of the select_effect, but that would mean then trickling down the updated width to the items.

Alecaddd:
@mbfraga Updated the State Managers to use the existing methods in the selected bounds manager to move and scale.
With that we also nicely inherit the snapping feature.

**Little problem**
When the item is rotate and updating the width or height field to 1, the resizing doesn't work properly.
Any idea why?

mbfraga (commented):
> @mbfraga Updated the State Managers to use the existing methods in the selected bounds manager to move and scale.
> With that we also nicely inherit the snapping feature.
> 
> **Little problem**
> When the item is rotate and updating the width or height field to 1, the resizing doesn't work properly.
> Any idea why?
Not 100% sure. It may have to do with how the size is being clamped to at least 1px. I can test the code.

src/Lib/Components/Coordinates.vala
```diff
@@ -43,6 +44,7 @@ public class Akira.Lib.Components.Coordinates : Component {
         }
     }
 
+    // Returns the global bounds.x1 value.
```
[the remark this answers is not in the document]
Alecaddd: That comment is not there anymore, I think you didn't refresh the page :D

mbfraga (approved):

PR #536, opened 2021-04-03
Use the new elementary OS 6 stylesheet

Alecaddd:
elementary OS 6 just updated their stylesheet in preparation of their beta release and removed the previously named `elementary` Gtk Theme.

This is a temporary fix as they will soon release an update SDK we can embed in the application.
The new default theme is now called `io.elementary.stylesheet.blueberry`

Pinging @bilelmoussaoui on this to see if we need to do anything on our flatpak file.

### DIFF
--- a/data/css/headerbar.css
+++ b/data/css/headerbar.css
@@ -13,3 +13,14 @@
 .export-titlebar {
     padding: 0;
 }
+
+/* Temporary elementary OS 6 stylesheet fixes */
+button.flat:disabled {
+    background-color: transparent;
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Until the elementary SDK's ready and published somewhere so we can use it, I think we should include the theme and probably a newer granite version (it's required i think?) into our flatpak manifest. I can have a look at that today

bilelmoussaoui:
see #538 #537 which should hopefully solve your issue. Let me know how that goes! :D

PR #553, opened 2021-04-12
WIP: Duplicate canvas items

Abdallah-Moh:
This Allows the user to duplicate Canvas Items and properties with `Ctrl + D`

## Screenshots 
AkiraDuplicateTest (1)

When duplicating an artboard it's label does not show unless the user move the artboard
When duplicating an artboard it's children does not get duplicated
### Issue Screenshot
AkiraDuplicateTest

Allow the user to duplicate a CanvasItem with `Ctrl + D`

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -266,27 +266,27 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                         Items.CanvasItem new_item = null;
                         if (selected_item is Items.CanvasArtboard) {
                             window.event_bus.insert_item ("artboard");
-                            new_item = window.items_manager.insert_item (selected_item.coordinates.x, selected_item.coordinates.y);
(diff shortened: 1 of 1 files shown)


Abdallah-Moh:
# I am thinking of converting this to a method and calling it to copy artboard childs

Alecaddd:
Unfortunately this is not a good approach.

As you can see, with this approach you're forced to manually loop through all the items and specifically identify and target the shape type (text, image, etc).
This is a first clear red flag of how not scalable this code is an how it will become more and more unmanageable with time.

This should be done with existing method, in a smarter and more automated way.
You should serialize all the selected items with `JsonItemSerializer`, and then load them into the canvas with the `JsonDeserializer`.
Basically using the same method we're using to save and load files.

This code also shouldn't be in the Canvas file.
That file should remain as small as possible and these type of methods should be delegated to dedicated classes to keep the code lean and easy to read.

Abdallah-Moh:
So what should I do

PR #566, opened 2021-04-25
Mode manager

### LINKED ISSUE #498: Unable to click on a Selection Nob when zoomed in
## Expected Behavior
Selection Nob should work at every zoom.

## Current Behavior
If you zoom in passed 1200%, the Selection Nobs on the left or right edges are not selectable anymore.
It seems that the corner nobs always work.

It's also a bit hard to reproduce and it doesn't happen every time.
Sometimes it happens if you select the item first at regular zoom, then you zoom in, resize it from the corners, and then try to select the Nob at the edge.
Tricky.

mbfraga:
This PR separates logic for mouse interaction modes into a Manager and abstract Modes. It builds stronger invariants and cleans up code, as well as solving ambiguous behaviors.

Some invariants:
1. Only one mode can exist at any time (this is guaranteed by RAII), except for panning that works as a masking mode.
2. Event absorption allows better flow with modes.
3. Cursors are now tied to the mode directly, which benefits from RAII.

## Steps to Test
Test all interaction modes and how they work together:
1. Exporting a selected area
2. Inserting an item and resizing it
3. Transforming an existing item with the mouse.
4. Panning via middle click or space held -- try while other modes are active.

1. Modes use canvas method and in the future this could be narrowed down a bit
2. Modes take the mode_manager to be able to stop themselves, this should probably be fixed by adding a nice way to trigger these events. This requires an architectural decision--no wrong answer.
4. As always, a lot of things could be cleaned up with more valaistic approaches. I kept the code naive and open-ended on purpose. On C++ I would have build a few more invariants into modes, but not sure the best approaches in vala. Not a huge deal.

This PR improves code in terms of readability, modularity, and also opens the path for cool new interactions and features. e.g., it would be nice to add a dead zone for transformations to not perform them by mistake when selecting items.

This PR does improve some behaviors that were a bit broken before.

This PR fixes #498

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -210,14 +210,14 @@ public class Akira.Lib.Canvas : Goo.Canvas {
 
         }
 
-        if (mode_manager.key_press_event(event)) {
+        if (mode_manager.key_press_event (event)) {
             return true;
         }
 
         switch (uppercase_keyval) {
             case Gdk.Key.space:
                 mode_manager.start_panning_mode ();
-                if (mode_manager.key_press_event(event)) {
+                if (mode_manager.key_press_event (event)) {
                     return true;
                 }
                 break;
@@ -254,7 +254,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
                 break;
         }
 
-        if (mode_manager.key_release_event(event)) {
+        if (mode_manager.key_release_event (event)) {
             return true;
         }
 
@@ -276,7 +276,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         }
 
         if (event.button == Gdk.BUTTON_MIDDLE) {
-            mode_manager.start_panning_mode();
+            mode_manager.start_panning_mode ();
             if (mode_manager.button_press_event (event)) {
                 return true;
             }
@@ -309,13 +309,13 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     }
 
     public void start_export_area_selection () {
-        var newMode = new Akira.Lib.Modes.ExportMode(this, mode_manager);
-        mode_manager.register_mode(newMode);
+        var new_mode = new Akira.Lib.Modes.ExportMode (this, mode_manager);
+        mode_manager.register_mode (new_mode);
     }
 
     public void on_insert_item () {
-        var newMode = new Akira.Lib.Modes.ItemInsertMode(this, mode_manager);
-        mode_manager.register_mode(newMode);
+        var new_mode = new Akira.Lib.Modes.ItemInsertMode (this, mode_manager);
+        mode_manager.register_mode (new_mode);
     }
 
     /**
@@ -392,8 +392,8 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         selected_bound_manager.set_initial_coordinates (event.x, event.y);
 
         if (selected_bound_manager.selected_items.length () > 0) {
-            var newMode = new Akira.Lib.Modes.TransformMode(this, mode_manager);
-            mode_manager.register_mode(newMode);
+            var new_mode = new Akira.Lib.Modes.TransformMode (this, mode_manager);
+            mode_manager.register_mode (new_mode);
 
             if (mode_manager.button_press_event (event)) {
                 return true;
@@ -496,7 +496,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         }
     }
 
-    /**
+    /*
      * Show or hide the pixel grid based on its state.
      */
     private void on_toggle_pixel_grid () {
@@ -510,9 +510,12 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         is_grid_visible = false;
     }
 
-    public void update_pixel_grid_if_visible() {
+    /*
+     * Updates pixel grid if visible, useful to guarantee z-order in paint composition.
+     */
+    public void update_pixel_grid_if_visible () {
         if (is_grid_visible) {
-            update_pixel_grid();
+            update_pixel_grid ();
         }
     }
 
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -126,8 +126,6 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         Gtk.drag_finish (drag_context, true, false, time);
 
         update_canvas ();
-        // Reset the edit mode.
-        edit_mode = EditMode.MODE_SELECTION;
     }
 
     private void create_pixel_grid () {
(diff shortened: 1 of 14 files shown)


src/Lib/Managers/HoverManager.vala
```diff
@@ -103,7 +86,7 @@ public class Akira.Lib.Managers.HoverManager : Object {
             canvas.window.event_bus.hover_over_item (item);
         }
 
-        set_cursor_for_nob (Managers.NobManager.Nob.NONE);
+        return true;
```
Alecaddd, line 89: It would be helpful to get the meaning of the return (true or false) in the comment of this method, just to know what they mean and what they affect.

src/Lib/Managers/ModeManager.vala
```diff
@@ -0,0 +1,181 @@
...
+ * in order to allow these modes to affect items in the canvas as well as allowing them to absorb said
+ * events.
+ *
+ * Only one InteractionMode is guaranteed to exist at a time, and this manager will correctly alert the
+ * beginning and end of a registered mode.
+ *
+ * The exception to the above rule is pan_mode, which can be running on top of another mode in certain cases and
+ * masks events appropriately.
+ *
+ * See ItneractionMode.vala for more details on how to create modes.
```
Alecaddd, line 33: nit: little typo in the InteractionMode.

src/Lib/Managers/ModeManager.vala
```diff
@@ -0,0 +1,181 @@
+/**
+ * Copyright (c) 2021 Alecaddd (http://alecaddd.com)
```
Alecaddd, line 2: nit: https :P

src/Lib/Managers/NobManager.vala
```diff
@@ -103,21 +104,18 @@ public class Akira.Lib.Managers.NobManager : Object {
...
+            if (ui_nob_s != null && ui_nob_s.is_visible ()) {
+                var ui_nob = ui_nob_s as Akira.Lib.Selection.Nob;
+                if (ui_nob.hit_test (x, y, scale)) {
+                    return ui_nob.handle_id;
+                }
             }
         }
 
-        return (Nob) grabbed_id;
+        return Nob.NONE;
```
Alecaddd, line 118: I think we could update the attribute to be an array of Selection.Nob instead of Goo.CanvasItemSimple.
`private Lib.Selection.Nob[] nobs = new Lib.Selection.Nob[9];`
Not sure if that affects performance, but in that way we can avoid type casting these nobs every time and simplify the code.

```
foreach (var nob in nobs) {
            if (nob == null || !nob.is_visible ()) {
                continue;
            }

            if (nob.hit_test (x, y, scale)) {
                return nob.handle_id;
            }
}
```

src/Lib/Managers/NobManager.vala
```diff
@@ -509,6 +549,7 @@ public class Akira.Lib.Managers.NobManager : Object {
         for (int i = 0; i < 9; i++) {
             var nob = new Selection.Nob (root, (Managers.NobManager.Nob) i);
             nob.set ("visibility", Goo.CanvasItemVisibility.HIDDEN);
+            nob.pointer_events = Goo.CanvasPointerEvents.NONE;
```
Alecaddd, line 552: We should add a comment here to explain why we're preventing pointer events on the nobs and the way we're handling nobs selection.

src/Lib/Modes/InteractionMode.vala
```diff
@@ -0,0 +1,98 @@
...
+ * 1. Add a new ModeType
+ * 2. Create a new class that inherits from Object and InteractionMode
+ * 3. Implement all abstract methods, and return the correct mode_type
+ * 4. Create the new mode with the right trigger in Canvas, and ModeManager will automatically handle it
+ *    based on the abstract methods below. Sometimes the same event that creates the mode should be passed
+ *    to the mode_manager after the creation of the mode in order to guarantee correct behavior.
+ * 5. For now, modes should take mode_manager on construction in order to be able to stop themselves. In
+ *    the future this should be built into the api in a more abstract way.
+ */
+public interface Akira.Lib.Modes.InteractionMode : Object {
```
Alecaddd, line 41: Do you see a future when this interface might get out of hand?
Originally all the canvas items were based on a common base interface with all attributes and it quickly turned into a  mess to manage, and I switched to components.
This feels pretty contained so I'm not concerned, just a note.

src/Lib/Modes/ItemInsertMode.vala
```diff
@@ -0,0 +1,101 @@
...
+        Object (
+            canvas: canvas,
+            mode_manager : mode_manager
+        );
+    }
+
+    public void mode_begin () {}
+    public void mode_end () {}
+    public InteractionMode.ModeType mode_type () { return InteractionMode.ModeType.ITEM_INSERT; }
+
```
Alecaddd, line 45: nit: remove blank line

src/Lib/Modes/InteractionMode.vala
```diff
@@ -0,0 +1,98 @@
...
+    /*
+     * Mode type that is used for introspection.
+     */
+    public enum ModeType {
+        UNDEFINED = 0,
```
Alecaddd, line 46: Shouldn't this be `SELECTION` as default?
Also, why this mode has a value of 0 but the others don't? Do we need a value 0 if we're using enum?

Alecaddd (approved):
This is great!
I love the approach and the modularity that this brings.
There are some nits fixes and some questions I left.
After those are handled this can be merged.

src/Lib/Canvas.vala
```diff
@@ -481,6 +350,67 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         grab_focus (get_root_item ());
     }
 
+    private bool press_event_on_selection (Gdk.EventButton event) {
+
```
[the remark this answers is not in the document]
mbfraga: done

src/Lib/Managers/HoverManager.vala
mbfraga: Made it void. In the future we can see if it needs to return whether it was absorbed.

src/Lib/Managers/ModeManager.vala
mbfraga: done

src/Lib/Managers/ModeManager.vala
mbfraga: Lol, fixed. Pretty sure I copy/pasted this.

src/Lib/Managers/NobManager.vala
mbfraga: Added a comment

src/Lib/Modes/InteractionMode.vala
mbfraga: It's a valid concern. I don't see these growing too much as interfaces--but I am confident that for now an interface is the easiest approach. If they get out of hand, I don't foresee major refactors to the general approach. I could be wrong though--time will tell?

src/Lib/Modes/ItemInsertMode.vala
mbfraga: done

src/Lib/Modes/InteractionMode.vala
mbfraga: I think SELECTION as default is misleading. Selection is not technically an interaction mode. Potentially it could be, but I don't think we will get much from it. I would leave that for some other time. Removed UNDEFINED, since it's technically unused.

src/Lib/Managers/NobManager.vala
mbfraga: changed it.

PR #576, opened 2021-05-03
Improve build time and responsiveness with language server

Prince781:
Instead of essentially building the entire project twice, we can reduce build time by only building src/Main.vala and then statically linking it with akira-library. Not only does this reduce build time, it also makes [VLS](https://github.com/Prince781/vala-language-server) more responsive (although still not responsive enough—there's much to improve on the VLS side to get it to a satisfactory level).


Alecaddd (approved):
Thank you so much for your contribution, this is great!

PR #577, opened 2021-05-03
Fix left and right not being absorbed

mbfraga:
The left and right key were not being absorbed after moving the time, so focus change was being triggered. I returned true to absorb the event.

## Steps to Test
Create an item, and press left twice. The item should move twice and there should be no focus change.


Alecaddd (approved):
Thanks for the quick fix.

PR #580, opened 2021-05-05
Implement Global Colors

Abdallah-Moh:
This will allow the user to save colors globally to be used again in another files

## Screenshots 
image

### DIFF
--- a/src/Layouts/Partials/FillItem.vala
+++ b/src/Layouts/Partials/FillItem.vala
@@ -217,7 +217,104 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
 
         color_picker = new Gtk.Grid ();
         color_picker.get_style_context ().add_class ("color-picker");
+        color_picker.row_spacing = 12;
+
+        // Document Grid - The Grid that will have all of the document saved colors
+        var document_grid = new Gtk.Grid ();
+        document_grid.margin_start = document_grid.margin_end = 40;
+        document_grid.hexpand = true;
+        var document_lbl = new Gtk.Label ("Document Colors");
+        document_lbl.halign = Gtk.Align.START;
+        document_grid.attach (document_lbl, 0, 0, 1, 1);
+        document_grid.row_spacing = 12;
+        var document_colors_grid = new Gtk.Grid ();
+        document_colors_grid.column_spacing = 5;
+        document_colors_grid.row_spacing = 5;
+
+        int document_column_counter = 0;
+        int document_row_counter = 0;
+
+        if (document_column_counter % 11 == 0) {
+            document_column_counter = 0;
+            document_row_counter++;
+        }
+        var document_add_btn = new Gtk.Image.from_icon_name ("list-add-symbolic", Gtk.IconSize.LARGE_TOOLBAR);
+        var document_btn_event_box = new Gtk.EventBox ();
+        document_btn_event_box.add (document_add_btn);
+
+        document_colors_grid.attach (document_btn_event_box, document_column_counter, document_row_counter, 1, 1);
+        document_grid.attach (document_colors_grid, 0, 1, 1, 1);
+
+        // Global Grid - The Grid that will have all of the document saved colors
+        var global_grid = new Gtk.Grid ();
+        global_grid.margin_start = global_grid.margin_end = 40;
+        global_grid.hexpand = true;
+        var global_lbl = new Gtk.Label ("Global Colors");
+        global_lbl.halign = Gtk.Align.START;
+        global_grid.attach (global_lbl, 0, 0, 1, 1);
+        global_grid.row_spacing = 12;
+        var global_colors_grid = new Gtk.Grid ();
+        global_colors_grid.column_spacing = 5;
+        global_colors_grid.row_spacing = 5;
+
+        int global_column_counter = 0;
+        int global_row_counter = 0;
+        foreach (string global_color in settings.global_colors) {
+            var color_item = new Akira.Partials.RoundedColorButton (global_color);
+            color_item.clicked.connect (()=>{
+                var rgba_color = Gdk.RGBA ();
+                rgba_color.parse (color_item.background_color);
+                color_chooser_widget.set_rgba (rgba_color);
+            });
+            // Break to the next row if the umber of colors in current row = 10
+            if (global_column_counter == 11) {
+                global_column_counter = 0;
+                global_row_counter++;
+            }
+            global_colors_grid.attach (color_item, global_column_counter, global_row_counter, 1, 1);
+            global_column_counter++;
+        }
+        if (global_column_counter == 11) {
+            global_column_counter = 0;
+            global_row_counter++;
+        }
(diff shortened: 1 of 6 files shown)


Abdallah-Moh:
After Finishing this the Global Coloring I will start working on the Document then removing colors

src/Layouts/Partials/FillItem.vala
```diff
@@ -217,7 +217,104 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
 
         color_picker = new Gtk.Grid ();
         color_picker.get_style_context ().add_class ("color-picker");
+        color_picker.row_spacing = 12;
+
+        // Document Grid - The Grid that will have all of the document saved colors
+        var document_grid = new Gtk.Grid ();
```
Alecaddd, line 223: This shouldn't be a grid.
It's better to use a `Gtk.FlowBox` since we need these colors to be floating in different rows based on the space available.
This will also allow us to reorder them with drag&drop more easily.
https://valadoc.org/gtk4/Gtk.FlowBox.html

src/Partials/RoundedColorButton.vala
```diff
@@ -0,0 +1,53 @@
+/*
+* Copyright (c) 2021 Alecaddd (http://alecaddd.com)
```
Alecaddd, line 2: nit: https

src/Partials/RoundedColorButton.vala
```diff
@@ -0,0 +1,53 @@
...
+*/
+
+public class Akira.Partials.RoundedColorButton : Gtk.Button {
+    public string background_color;
+
+    public RoundedColorButton (string bg_color) {
+        background_color = bg_color;
+        can_focus = false;
+        width_request = 25;
+        height_request = 25;
```
Alecaddd, line 29: You can put this inline: `width_request = height_request = 24;`
Let's use 24 so we can better round em up.

Alecaddd (changes requested):
Good stuff, but I think using a flowbox is a more solid and scalable approach.
I like how contained and targeted the PR is, good job.

src/Layouts/Partials/FillItem.vala
Abdallah-Moh: I think you meant the `document_colors_grid` to be a FlowBox since this is what's containing the colors

src/Layouts/Partials/FillItem.vala
```diff
@@ -217,7 +217,73 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
...
+        global_colors_grid.row_spacing = 5;
+
+        foreach (string global_color in settings.global_colors) {
+            var color_item = new Akira.Partials.RoundedColorButton (global_color);
+            color_item.clicked.connect (()=>{
+                var rgba_color = Gdk.RGBA ();
+                rgba_color.parse (color_item.background_color);
+                color_chooser_widget.set_rgba (rgba_color);
+            });
+            global_colors_grid.insert (color_item, -1);
```
Alecaddd, line 258: As we discussed on Matrix, this should be added into the last position before the add_button.
A sort function needs to be defined in order to allow it.

src/Layouts/Partials/FillItem.vala
```diff
@@ -217,7 +217,73 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
...
+        }
+
+        var add_btn = new Akira.Partials.ColorAddButton ();
+        add_btn.button_press_event.connect (() => {
+            string[] array = settings.global_colors;
+            array += color_chooser_widget.rgba.to_string ();
+            settings.set_strv ("global-colors", array);
```
Alecaddd, line 265: You already defined a setter method, so here you can directly do `settings.global_colors = array`.
Also, let's rename `array` with something more eloquent, like `new_colors` to make the code more readable.

src/Layouts/Partials/FillItem.vala
```diff
@@ -217,7 +217,73 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
...
+            // Remove Add Button to place it as the last in the end
+            global_colors_grid.remove (add_btn);
```
Alecaddd, line 267: Let's not remove and re-add the button but properly use the sorting capability of the FlowBox element.

src/Layouts/Partials/FillItem.vala
```diff
@@ -217,7 +217,73 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
...
+            // Create Item
+            var color_item = new Akira.Partials.RoundedColorButton (color_chooser_widget.rgba.to_string ());
```
Alecaddd, line 269: We're calling the `color_chooser_widget.rgba.to_string ()` method a couple of times.
Better store this string into a variable at the beginning so we don't need to fetch it again.

src/Layouts/Partials/FillItem.vala
```diff
@@ -217,7 +217,73 @@ public class Akira.Layouts.Partials.FillItem : Gtk.Grid {
...
+            color_item.clicked.connect (()=>{
+                var rgba_color = Gdk.RGBA ();
+                rgba_color.parse (color_item.background_color);
+                color_chooser_widget.set_rgba (rgba_color);
+            });
+            global_colors_grid.insert (color_item, -1);
+        }
+
+        var add_btn = new Akira.Partials.ColorAddButton ();
+        add_btn.button_press_event.connect (() => {
```
Alecaddd, line 262: Move this method to its own instead of having it inline.

src/Partials/ColorAddButton.vala
```diff
@@ -0,0 +1,30 @@
...
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+
+* You should have received a copy of the GNU General Public License
+* along with Akira.  If not, see <https://www.gnu.org/licenses/>.
+*
+* Authored by: Abdallah "Abdallah-Moh" Mohammad <abdullah_mam1@icloud.com>
+*/
+
+public class Akira.Partials.ColorAddButton : Gtk.EventBox {
```
Alecaddd, line 22: This should be a button, not an event box.

Alecaddd (changes requested):

Alecaddd (changes requested):
You didn't fix my original comment regarding adding the `add_btn`.
This doesn't work as you can't re-add the same widget. Aren't you getting these errors in the terminal?

```
(Akira:22645): Gtk-WARNING **: 22:33:00.325: Attempting to add a widget with type AkiraPartialsColorAddButton to a container of type GtkFlowBoxChild, but the widget is already inside a container of type GtkFlowBoxChild, please remove the widget from its existing container first.
```

I said, you shouldn't simply add the new color in the last position but you should create a sorting function to properly handle the position of the colors.

I can take care of it if you want.

Abdallah-Moh:
@Alecaddd Sure

Alecaddd:
Closing this in favor of  #599

PR #587, opened 2021-05-17
Component refactor

mbfraga:
Large refactor meant to separate view / model.

An architecture for a model was built up.

Some details:
1. Goo.Canvas items are composed to the model items.
2. Components are meant to be largely immutable. When values are changed, the component should be replaced. This guarantess that the history of an item is kept.
3. I'll think of more...

## Steps to Test
When disabled, make sure everything in Akira works correctly. Then enable the new components.

To enable: set Window.vala 's `use_new_component` to true and rebuild.

WIP, a lot of things to do.


mbfraga:
No longer relevant.

PR #588, opened 2021-05-19
Convert a few private properties to private fields

Prince781:
These are invalid property names anyway (and valac 0.54 will complain
about them), and they don't need getters and setters.


Alecaddd (approved):
Thanks for the fixes.

PR #589, opened 2021-05-20
Fix Theme not loading on non-elementary distros if opened from files manager

Abdallah-Moh:
This problem only occurs with non-elementary distros if you open a `.akira` file from the files manager akira will load without the elementary theme and icons

## Steps to Test
Use any distro that is not Elementary OS open a `.akira` file from the files manager

This fixes theme not loading on non elementary distros when opening a file

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -41,7 +41,7 @@ public class Akira.Application : Gtk.Application {
 
         weak Gtk.IconTheme default_theme = Gtk.IconTheme.get_default ();
         default_theme.add_resource_path ("/com/github/akiraux/akira");
-            
+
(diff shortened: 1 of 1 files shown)


Alecaddd (changes requested):
Unfortunately this is not a good solution.
By doing this you're setting the theme and loading the icons every time a file is opened, even after the application was already initialized.
With this, if the user opens 6 files, the same themeing operation is triggered for each opened file.

We should investigate why the `activate()` method is not triggered when the application is launched trough a double click on an `.akira` file.

Alecaddd:
Closing in favor of #590

PR #590, opened 2021-05-21
Fix application not properly activating if opened via saved file

### LINKED ISSUE #551: Something wrong with css theme
Some icons are white, text on rigth layer panel too, the color picker only has an outline of the selected color:  
image  
"Use Dark Theme" changes only color of preferences headerbar:  
image
Changing the GTK theme or icon theme has no effect.
Desktop: GNOME 40

Alecaddd:
If Akira is opened via a saved `.akira` file, the `activate ()` method doesn't run properly, preventing proper setup and theme loading.
This patch make sure to run that method on the proper occasion, and avoid it if the application is already active.

## Steps to Test
1. Create a file, save it, and close Akira.
2. Double click on the saved file.
3. Confirm the theme and settings are working properly.
4. This can be verified only on a non-elementary system.

- Fixes #551, probably...hopefully.
- Closing #589 in favor of this.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -36,6 +36,11 @@ public class Akira.Application : Gtk.Application {
     }
 
     public override void open (File[] files, string hint) {
+        // The application was requested to open some files. Be sure to activate
+        // it in case it wasn't already running.
+        activate ();
+
+        // Loop through all selected files.
         foreach (var file in files) {
             if (is_file_opened (file)) {
                 // Present active window with currently opened file.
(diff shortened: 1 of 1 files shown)


src/Application.vala
```diff
@@ -107,6 +112,11 @@ public class Akira.Application : Gtk.Application {
     }
 
     protected override void activate () {
+        // Interrupt if we have at least one existing window.
+        if (windows.length () > 0) {
+            return;
+        }
+
```
Abdallah-Moh, line 119: This will not fix the double windows since it is being called before the windows are even created and this is not the only problem the user may have an akira window already opened so the windows will be more than 0 and that will not let the user open a new window 
so if you mean to do this then the code that is responsible of not allowing same file not opening and the code which is responsible to load the file in an empty window is non reachable

Abdallah-Moh (changes requested):

src/Application.vala
Alecaddd: good point

Alecaddd: It should be fixed now.

Abdallah-Moh (approved):
Every thing works as expected

PR #591, opened 2021-05-23
Fix terminal warnings and crosshair cursor after Image File is dragged into the canvas

Abdallah-Moh:
After the modes manager has been implemented some code stayed outdated
This will deregister the insert mode after the canvas image has been created

## Steps to Test
Drag an Image File into the canvas

## Screenshots
image

No known issues

Fix terminal warnings and crosshair cursor after Image File is dragged into the canvas

- Fixes Fix terminal warnings after Image File is dragged into the canvas
- Crosshair cursor after Image File is dragged into the canvas

### DIFF
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -117,7 +117,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         }
 
         Gtk.drag_finish (drag_context, true, false, time);
-        mode_manager.deregister_mode (Modes.InteractionMode.ModeType.ITEM_INSERT);
+        mode_manager.deregister_active_mode ();
 
         update_canvas ();
     }
--- a/src/Lib/Canvas.vala
+++ b/src/Lib/Canvas.vala
@@ -98,6 +98,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
     ) {
         // Loop through the list of the dragged files.
         int index = 0;
+        window.event_bus.insert_item ("image");
         foreach (string link in data.get_uris ()) {
             var file_link = link.replace ("file://", "").replace ("file:/", "");
             file_link = Uri.unescape_string (file_link);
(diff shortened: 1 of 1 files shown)


src/Lib/Canvas.vala
```diff
@@ -117,6 +117,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
         }
 
         Gtk.drag_finish (drag_context, true, false, time);
+        mode_manager.deregister_mode (Modes.InteractionMode.ModeType.ITEM_INSERT);
```
Alecaddd, line 120: No need to specify the mode here as we can directly call `mode_manager.deregister_active_mode ();`

Alecaddd (changes requested):

Alecaddd:
Since you're improving this area, there a bunch of other strange issues happening that you could investigate.

First, we could move the `window.event_bus.insert_item ("image");` outside the `foreach` loop since we don't need to re-set it for each image.

A couple of other problems appear when multiple images are dropped together.
1. Some images are not fully sized as they end up with 1px by 1px size.
2. Randomly, this error appears: `Gtk-WARNING **: 21:47:38.979: drawing failure for widget 'AkiraLibCanvas': invalid matrix (not invertible)` alongside others and the whole UI breaks.

I suspect there's a race issue with the generation of the pixbuf of each image.

Abdallah-Moh:
Can you please open an Issue or a discussion with more info and ping me

Abdallah-Moh:
I am currently investigating for the problem of the image with short height or short width
I did some investigation which lead me to know that when the height is shorter than the width and the width was resized to a number that made calculating the height a decimal which is less than 1 which in int gets converted to a 0 which breaks the whole app so I got an Idea before setting the pixpuf we will check if the calculated pixpuf height is less than a 1 then we set it to 1 and vice versa for the width

src/Lib/Canvas.vala
```diff
@@ -108,7 +109,7 @@ public class Akira.Lib.Canvas : Goo.Canvas {
             // Create the image manager.
             var manager = new Lib.Managers.ImageManager (image, index);
             // Let the app know that we're adding image items.
-            window.event_bus.insert_item ("image");
+
```
Alecaddd, line 112: You moved the method but you forgot the comment.
I'll fix this, but try to be more aware of the code around and the structure you're touching.

PR #593, opened 2021-05-25
Fix Image with short height or width

### LINKED ISSUE #567: Errors when trying to add an image with long width and short height
## Expected Behavior
Images should be normally added no matter of their size.

## Current Behavior
If an image with a long width and short height is added via the image tool, multiple errors are generated and the UI glitches.

## Possible Solution
This is probably caused by the default sizing added to the image on creation and the pixbuf sizing happening to maintain the original image ratio.

1. Download this image: 
Screenshot from 2021-04-07 20-19-01
2. Add it through the image tool.

Abdallah-Moh:
## Summary / How this PR fixes the
Errors and glitch errors when inserting Image with short height or width
this will fix #567 

## Steps to Test
Insert image from image tool

No issues for now


src/Lib/Items/CanvasImage.vala
```diff
@@ -131,10 +157,14 @@ public class Akira.Lib.Items.CanvasImage : Goo.CanvasImage, Akira.Lib.Items.Canv
         manager.get_pixbuf.begin (w, h, (obj, res) => {
             try {
                 var _pixbuf = manager.get_pixbuf.end (res);
+                // TODO: run is_height_small and is_width_small
```
Abdallah-Moh, line 160: The user may resize the image to set the width to 2 for example and the height breaks

src/Lib/Items/CanvasImage.vala
```diff
@@ -92,6 +92,8 @@ public class Akira.Lib.Items.CanvasImage : Goo.CanvasImage, Akira.Lib.Items.Canv
                 // the Size component properly calculate the correct size ratio.
                 size.locked = true;
 
+                if (is_height_small ((int)size.width, (int)size.height) || is_width_small ((int)size.width, (int)size.height)) return;
+
```
Alecaddd, line 96: Wrong condition formatting.
Please use the standard formatting:
```
if () {

}
```

Also, write a comment above this condition to explain why we're doing an early return.

src/Lib/Items/CanvasImage.vala
```diff
@@ -103,6 +105,30 @@ public class Akira.Lib.Items.CanvasImage : Goo.CanvasImage, Akira.Lib.Items.Canv
         });
     }
 
+    private bool is_width_small (int width, int height) {
```
Alecaddd, line 108: The `size` component is available anywhere in this file, no need to pass any attribute for these methods.

src/Lib/Items/CanvasImage.vala
```diff
@@ -103,6 +105,30 @@ public class Akira.Lib.Items.CanvasImage : Goo.CanvasImage, Akira.Lib.Items.Canv
...
+        // Callculate the width if height was 1
```
Alecaddd, line 109: Typo for "Calculate".
Also, this sentence is not grammatically correct and it doesn't make sense.
What are you trying to say here?

src/Lib/Items/CanvasImage.vala
```diff
@@ -103,6 +105,30 @@ public class Akira.Lib.Items.CanvasImage : Goo.CanvasImage, Akira.Lib.Items.Canv
...
+        var adjusted_width = width / height;
+
+        // If width is less than 1 then it is 0 we will set it to 1
+        if (adjusted_width < 1) {
+            size.width = 2;
+            return true;
+        }
```
Alecaddd, line 116: There's not need to store a variable if it's only used for one condition.
You can shorten this with:
```
if (width / height < 1) {
  size.width = 2;
  return true;
}
```

src/Lib/Items/CanvasImage.vala
```diff
@@ -131,10 +157,14 @@ public class Akira.Lib.Items.CanvasImage : Goo.CanvasImage, Akira.Lib.Items.Canv
...
+                var pixpuf_width = _pixbuf.get_width ();
+                var pixpuf_height = _pixbuf.get_height ();
+
                 pixbuf = _pixbuf;
                 if (update) {
-                    width = _pixbuf.get_width ();
-                    height = _pixbuf.get_height ();
+                    width = pixpuf_width;
+                    height = pixpuf_height;
                 }
```
Alecaddd, line 168: mhhh...what's the reason for this?

Alecaddd (changes requested):
Thanks for the contribution but I don't think this is the right approach.
You're deferring the update of the `width` and `height` to a boolean method, but that's a wrong architecture.
Boolean methods should only be used for conditions, and then the consequence of those conditions should be handled in the main thread.
This is a problem because if you take a look at the `Size` component, you will see that whenever you update the `width` or the `height` another series of calculations are triggered and the `detect_image_size_change` event is called.

src/Lib/Items/CanvasImage.vala
Abdallah-Moh: the attributes is for the future because in my next PR I will add this method when the Image size is changed
since the user can resize the Image width to 2 and break the app again

src/Lib/Items/CanvasImage.vala
Abdallah-Moh: TODO: to call `is_height_small` and `is_width_small` with the `pixpuf_width, pixpuf_height` parameters that's why the methods take parameters

src/Lib/Items/CanvasImage.vala
Abdallah-Moh: I did not add it in this PR for some issues it had but I will add it in my next PR

src/Lib/Items/CanvasImage.vala
Alecaddd: No need to do this.
It's generally bad practice to update the code "in preparation for something else".
You're also affecting performance here as you're calling pixbuf methods which are heavy before are even needed, if any.

src/Lib/Items/CanvasImage.vala
Alecaddd: In that case you should rework this and do these calculations in the `Size` component, check if the item is an image and run these checks.

src/Lib/Items/CanvasImage.vala
```diff
@@ -103,6 +111,26 @@ public class Akira.Lib.Items.CanvasImage : Goo.CanvasImage, Akira.Lib.Items.Canv
...
+        if (width / height < 1) {
+            size.width = 2;
+            return true;
+        }
+        return false;
+    }
+
+    private bool is_height_small (int width, int height) {
+        // Calculate the height of the image if width was 2 without changing image size
+        // If height is less than 1 then it is 0 we will set it to 1
```
Alecaddd, line 126: I'm sorry but these comments are very confusing.
English is not my first language either, but these sentences don't make much sense.

Alecaddd (changes requested):

README.md
```diff
@@ -78,7 +78,7 @@ ninja && sudo ninja install
 ## 🤔 Questions
 
 If you want to ask any questions about the project, join us on our public Matrix channel at 
-[#akiraux:matrix.org](https://matrix.to/#/!kpfeTRbpocQrOFCFnJ:matrix.org) or on IRC at #akiraux on Freenode.
+[#akiraux:matrix.org](https://matrix.to/#/#akiraux:matrix.org) or on IRC at #akiraux on LiberaChat.
```
LucileDT, line 81: It may be nice to make an announcement on social medias about this change @Alecaddd

Abdallah-Moh: @LucileDT What change

Abdallah-Moh: We haven't made any change we fixed the link after we switched to another channel

LucileDT: Yes and I never saw any post explaining AkiraUX moved to LiberaChat, this correction just made me think it would be a good idea :slightly_smiling_face:

Issue #21, opened 2018-01-06
Main Menu gear Popover


Alecaddd:
Main Menu gear to the far right of the Toolbar.
List the full list of currently available tools and commands with shortcuts and tooltip description.

Alecaddd:
Updated main menu to the left with folder icon

PR #597, opened 2021-05-26
Fix bugs in Akira.Utils.Color

AshishS-1123:

There were two bugs in the implementation of Akira.Utils.Color.

1. The app crashes when you manually delete the '#' character from the ColorField text box.

  To fix this, another condition has been added to the ```Akira.utils.Color.is_valid_hex``` method to ensure that such an entry is caught and flagged as a non-valid hex.

2. If you enter 2 characters one after the other in an empty ColorField text box, the characters are misplaced.
  For example, if you type ```12```, the text comes out to be ```#21```.

  The reason this happens is that after entering ```1```, the code inside the ```insert_text``` signal handler(in ```Akira.Partials.ColorField```) adds a '#', but the cursor position is not updated. It stays where it was before- after the first character which now is '#'.

  To fix this bug, the ```position``` referential parameter has been used in the ```insert_text``` signal handler to move the cursor to the end of the text.

This way, the user doesn't have to type a '#' every time they want to add a new color.

## Steps to Test
 
Repeat the steps in point 1 and 2. The app should not crash and text displayed should be the same you entered.

### DIFF
--- a/src/Partials/ColorField.vala
+++ b/src/Partials/ColorField.vala
@@ -37,7 +37,7 @@ public class Akira.Partials.ColorField : Gtk.Entry {
         focus_in_event.connect (handle_focus_in);
         focus_out_event.connect (handle_focus_out);
 
-        insert_text.connect ((_new_text, new_text_length) => {
+        insert_text.connect ((_new_text, new_text_length, ref position) => {
             string new_text = _new_text.strip ();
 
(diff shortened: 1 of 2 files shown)


Alecaddd:
Thanks for the contribution, and welcome to the project :smile: 
I'll fix a couple of small linting issues before merging.

PR #616, opened 2021-06-27
Fix Image with short height and long width and vice versa

### LINKED ISSUE #567: Errors when trying to add an image with long width and short height

Abdallah-Moh:
This will fix ui crash when inserting image with short width or height

## Steps to Test
Insert image from menu

- [x] Fix image with short width
- [x] Fix image with short height

- Fixes #567

### DIFF
--- a/src/Lib/Components/Size.vala
+++ b/src/Lib/Components/Size.vala
@@ -41,7 +41,13 @@ public class Akira.Lib.Components.Size : Component {
 
             if (locked && !auto_resize) {
                 auto_resize = true;
-                height = Utils.AffineTransform.fix_size (value / ratio);
+                if (item is Items.CanvasImage && Utils.AffineTransform.fix_size (value / ratio) < 1) {
+                    height = 1;
+                    print("SHit");
+                }
+                else {
+                    height = Utils.AffineTransform.fix_size (value / ratio);
+                }
                 auto_resize = false;
             }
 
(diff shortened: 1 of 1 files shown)


src/Lib/Components/Size.vala
```diff
@@ -41,7 +41,12 @@ public class Akira.Lib.Components.Size : Component {
 
             if (locked && !auto_resize) {
                 auto_resize = true;
-                height = Utils.AffineTransform.fix_size (value / ratio);
+                if (item is Items.CanvasImage && Utils.AffineTransform.fix_size (value * ratio) < 1) {
+                    height = 1;
+                }
+                else {
+                    height = Utils.AffineTransform.fix_size (value * ratio);
+                }
```
Alecaddd, line 49: Good solution but we can clean up the code a bit to avoid repetition and unnecessary double calculations.
```suggestion
var locked_height = Utils.AffineTransform.fix_size (value / ratio);
height = item is Items.CanvasImage && locked_height < 1 ? 1 : locked_height;
```

src/Lib/Components/Size.vala
```diff
@@ -70,7 +75,12 @@ public class Akira.Lib.Components.Size : Component {
 
             if (locked && !auto_resize) {
                 auto_resize = true;
-                width = Utils.AffineTransform.fix_size (value * ratio);
+                if (item is Items.CanvasImage && Utils.AffineTransform.fix_size (value * ratio) < 1) {
+                    width = 2;
+                }
+                else {
+                    width = Utils.AffineTransform.fix_size (value * ratio);
+                }
```
Alecaddd, line 83: Same here:
```suggestion
var locked_width = Utils.AffineTransform.fix_size (value * ratio);
width = item is Items.CanvasImage && locked_width < 1 ? 1 : locked_width;
```

Alecaddd (changes requested):

Alecaddd (approved):

PR #623, opened 2021-07-04
Allow reading and parsing gpl files

Abdallah-Moh:

This pr will implement functions/methods that can read and convert a gpl file to an ArrayList of Gdk.RGBA colors
this pr is a part of a bigger implementation of importing gpl files in akira

## Steps to Test
Use functions/methods implemented and give it a path for a .gpl file
and it will return a class containing an ArrayList of Gdk.RGBA colors called colors


mbfraga:
Cool! Only thing is, perhaps we should look at https://gitlab.gnome.org/GNOME/gimp/-/blob/gimp-2-10/app/core/gimppalette-load.c and follow it so that we get some documentation on the file format. e.g., I don't think the colors necessarily start at line 3, but they start after '#' or something like that

Abdallah-Moh:
Waiting for `main` to be finished so that this pr does not face conflicts

PR #636, opened 2021-07-31
Implement solid transformation code

mbfraga:
This transformation code robustly works with local transformations
and multi-select.

Main missing feature for now is pixel-snappping.

## Steps to Test
1. Scale single objects
2. Scale single rotated objects
3. Scale multiple objects (including a rotated one)
4. After 3, the rotated one should be skewed. Scale that object independently. 

1. Snapping to the pixel grid would be a solid next step.


Alecaddd (approved):
This looks good, thank you so much.
There a couple of issues I noticed, but we can deal with those in a follow up patch:

1. The selection nobs don't rotate alongside the selection box, but I think they should.
2. When rotating again a rotated item, there's an initial jump on the first click (we had this problem in the past and fixed a couple of times)

PR #638, opened 2021-07-31
Hide selection box when interacting with a single item

Alecaddd:
Hide the selection box when transforming/translating a single item.
This leaves only the clicked Nob visible, which improves visual edge snapping and alignment.

The selection box should remain visible when multiple items are selected since we can't fully visually rely on the natural bounding box of a simple item.


mbfraga (approved):

Alecaddd:
Nevermind, I changed my mind on this.
Hiding the selection box looks pretty weird in some cases.

PR #648, opened 2021-08-16
Implement serialization / deserialization for Lib2

mbfraga:
Still a WIP, a lot of checks aren't there, so if something
isn't right it could crash. For the most part it should
be ok. It can be improved with time.

Structure is also not ideal, but not sure what would be the
best way in the future. I mostly want to get the ball rolling,
and I kept serialization and deserialization aspects in
files together to make things easier to follow.

Things could potentially be better with delegates and whatnot,
but meh.

## Steps to Test
Save and load normally with Ctrl-s and Ctrl-o. Only works within Lib2, it will not open master files.


src/FileFormat/JsonItemSerializer.vala
```diff
@@ -28,180 +28,40 @@ public class Akira.FileFormat.JsonItemSerializer {
...
-     */
-    private static void serialize_type_specifics (Lib.Items.CanvasItem item, ref Json.Object object) {
-        // Set a string of the type so we're not tied to the namespace and location.
-        if (item is Lib.Items.CanvasArtboard) {
-            object.set_string_member ("type", "artboard");
+        // serialize type
+        {
+            builder.set_member_name ("type");
+            builder.add_string_value (node.instance.type.name_id);
         }
```
Alecaddd, line 38: why this is inside brackets?

Alecaddd (approved):

PR #649, opened 2021-08-17
Enable Layers UI

Alecaddd:
Re-enable the Layers UI to slowly get to feature parity with `master`.
Many many things to do, but this is just a start to test things out and consider different approaches.

A lot of things are broken.

- [x] Show the Layers Panel
- [x] Create a new layer every time an item is created
- [x] Select layer when item is selected
- [x] Delete the layer when the item is deleted


src/Layouts/Sidebars/Partials/LayerElement.vala
```diff
@@ -0,0 +1,86 @@
...
+public class Akira.Layouts.Sidebars.Partials.LayerElement : Gtk.ListBoxRow {
+    public unowned Lib2.ViewCanvas view_canvas { get; construct; }
+    public unowned Lib2.Items.ModelInstance node { get; construct; }
+
+    private Gtk.Label label;
+
+    public LayerElement (Lib2.Items.ModelInstance instance, Lib2.ViewCanvas canvas) {
+        Object (
+            view_canvas: canvas,
+            node: instance
```
mbfraga, line 35: call it instance instead of node? :P  Maybe cloning the instance may be better. But not sure if we need to keep the instance or simply the id and any other data that is need (name?)

mbfraga (approved):
small comments. looks like a good start!

src/Layouts/Sidebars/Partials/LayerElement.vala
Alecaddd: Layers will need to trigger the hover effect, as well as selecting the item, deleting it, changing the `name`, and updating the `lock` and `visibility` attributes.
What do you think would be better to dd to handle those things?

Alecaddd:
Very rough first UI PR just to test things out.
Many changes and improvements are necessary, with most likely some smarter ways to handle things, but it's a start.

mbfraga (approved):

PR #652, opened 2021-08-19
No goocanvas

mbfraga:
This PR implements a custom Canvas that does not rely on goocanvas (although it is heavily insipired by it and takes code directly from it--so we still need attribution).

The goal is to have more control over drawables and performance.

Some important changes:
* Added ViewLayer concept and rewrote managers and previous CanvasItems. This actuallly reduced code by a lot.
    - Grid (not 1:1 with master)
    - Hover (now displays the path rather than the bounding box)
    - Nobs and Selection
    - Snaps
 * Added a BaseCanvas to replace Goo.Canvas. In the future we should probably restructure and rename it.
 * Since don't use GooCanvas, the view is a direct reflection of the model...so all synchronization code can go away! This speeds up removal and sorting SIGNIFICANTLY. For remove, we go from >O(n^2) to <O(n) roughly. It scales very well! Re-stacking will be similar. And the code complexity will go down by a lot.
 * The model now takes the canvas to be able to properly request redraws. In the future it would be nice to decouple these, but for now the usage is minimal and optional.

* There are drawing artifacts (that also existed in goocanvas. These happen when you move items around--once the full canvas redraws they go away. So they are fairly harmless for now. Still something that would be nice to get rid of.
artifacts


.gitignore
```diff
@@ -2,6 +2,7 @@
 .vscode
 src/config.vala
 build/
+build*/
```
Alecaddd, line 5: mh, why? we need the `build-aux/` folder.

Alecaddd (approved):
Awesome work!
This is a massive milestone!

albfan:
Congrats on this @mbfraga

PR #660, opened 2021-08-28
Basic Paths Feature

AshishS-1123:
- Created action to insert a path when user presses ```p``` or from header bar.
- Created a **PathEditMode** to for creating a new path. In future PRs, this can be reused to edit paths of other CanvasItems when user double clicks on them.
- Created a **ViewLayerPath** so that users can easily visualise the points they are inserting in the path

## Steps to Test

- Press ```p``` or click on **Vector** in the insert menu
- Click randomly on the canvas. The points where you click will be highlighted in blue. This is the work of ViewLayerPath. All these points will be connected using lines.
- Press ```Esc``` to exit **PathEditMode**. The blue dots from **ViewLayerPath** should now be replaced by nobs and selection box.

## Screenshots 
animation

The following features will be implemented step by step in small PRs to make developing and reviewing easier.

- Add curves using the ```Ctrl```-```Alt``` method.
- Double clicking on an item should enter user in PathEditMode
- Scaling paths doesn't work either.


src/ViewLayers/ViewLayerPath.vala
```diff
@@ -0,0 +1,111 @@
...
+    public const double UI_NOB_SIZE = 4;
+    public const double UI_LINE_WIDTH = 1.01;
+
+    private Gee.ArrayList<Geometry.Point?>? points = null;
+    private Geometry.Rectangle extents;
+    private Geometry.Point first_point;
+
+    public void set_reference_point (Geometry.Point reference_point) {
+        first_point = reference_point;
+        canvas.request_redraw (extents);
```
mbfraga, line 32: it's better to create an "update" override that redraws the extents before and after an update. We can keep it to only update the new extent, but still creating that update method would be better.

mbfraga (changes requested):
Some relatively small issues that should not be hard to fix. The main one is that components are copy-on-write and need to be replaced whenever any change happens to them. Once in an item, a component is NEVER modified.

In general nice work!

src/Lib/Modes/ItemInsertMode.vala
```diff
@@ -93,6 +111,17 @@ public class Akira.Lib.Modes.ItemInsertMode : AbstractInteractionMode {
             view_canvas.selection_manager.reset_selection ();
             view_canvas.selection_manager.add_to_selection (instance.id);
 
+            // if a path is being inserted, then start the PathEditMode
+            if (item_insert_type == "path") {
+                path_edit_mode = new Akira.Lib.Modes.PathEditMode (view_canvas, true, instance);
```
Alecaddd, line 116: nit: no need for the `Akira.` namespace, you can start directly with `Lib.` since we're already withing the `Akira` namespace.

src/Lib/Modes/PathEditMode.vala
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * Authored by: Ashish Shevale <shevaleashish@gmail.com>
+*/
+
+public class Akira.Lib.Modes.PathEditMode : AbstractInteractionMode {
+
+    public weak Lib.ViewCanvas view_canvas { get; construct; }
+    public Lib.Items.ModelInstance instance { get; construct; }
+    public bool is_insert { get; construct; }
+    // keep track of previous point for calculating relative positions of points
```
Alecaddd, line 27: Little styling nit.
We're trying to write inline comments with this format:
`// Capital letter to start and then end with a period.`

src/Lib/Modes/PathEditMode.vala
```diff
@@ -0,0 +1,127 @@
...
+    private Geometry.Point first_point;
+    private ViewLayers.ViewLayerPath path_layer;
+
+
+    public PathEditMode (Lib.ViewCanvas canvas, bool is_insert, Lib.Items.ModelInstance instance) {
+        Object (
+            view_canvas: canvas,
+            is_insert: is_insert,
```
Alecaddd, line 35: What's this attribute? It doesn't seem to be used anywhere yet.

src/Services/ActionManager.vala
```diff
@@ -152,6 +154,7 @@ public class Akira.Services.ActionManager : Object {
         typing_accelerators.set (ACTION_ELLIPSE_TOOL, "e");
         typing_accelerators.set (ACTION_TEXT_TOOL, "t");
         typing_accelerators.set (ACTION_IMAGE_TOOL, "i");
+        typing_accelerators.set (ACTION_PATH_TOOL, "p");
```
Alecaddd, line 157: Let's use `v` as accelerator since we're calling it the vector tool.
`p` will be used for the pencil too.

src/Lib/Modes/ItemInsertMode.vala
```diff
@@ -32,14 +32,24 @@ public class Akira.Lib.Modes.ItemInsertMode : AbstractInteractionMode {
...
 
     private Lib.Modes.TransformMode transform_mode;
+    private Lib.Modes.PathEditMode path_edit_mode;
 
     public ItemInsertMode (Lib.ViewCanvas canvas, string item_type) {
         Object (view_canvas: canvas);
         item_insert_type = item_type;
+
+        // if PathEditMode is active, it must deregisterd with user presses escape
+        view_canvas.window.event_bus.request_escape.connect (() => {
```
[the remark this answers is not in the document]
AshishS-1123: I have handled ```Esc``` from ModeManager. In ModeManager, the request_escape signal causes the currently active mode the end.

src/Lib/Items/ModelTypePath.vala
```diff
@@ -85,9 +86,14 @@ public class Akira.Lib.Items.ModelTypePath : ModelType {
...
-                instance.drawable.center_y = -instance.compiled_geometry.source_height / 2.0;
-                instance.drawable.transform = instance.compiled_geometry.transformation_matrix;
+                // The points property is only available to DrawablePath, so first typecast it
+                // modify it, then assign to instance.
+                Drawables.DrawablePath drawable = instance.drawable as Drawables.DrawablePath;
+                drawable.center_x = -instance.compiled_geometry.source_width / 2.0;
+                drawable.center_y = -instance.compiled_geometry.source_height / 2.0;
+                drawable.transform = instance.compiled_geometry.transformation_matrix;
+                drawable.points = instance.components.path.data;
+                instance.drawable = drawable;
```
mbfraga, line 96: You don't need to do this (the last line). The drawable is a pointer, so you are modifying the drawable directly.

src/Drawables/DrawablePath.vala
```diff
@@ -37,11 +37,14 @@ public class Akira.Drawables.DrawablePath : Drawable {
             return;
         }
 
+        Geometry.Point[] translated_points = recalculate_points ();
```
mbfraga, line 40: it's probably better to do this on construction, it doesn't seem like there is a reason to do it every time the path is created. If you want to keep the original points, you can just keep both the original and the recalculated points.

src/Lib/Managers/ModeManager.vala
```diff
@@ -53,6 +53,11 @@ public class Akira.Lib.Managers.ModeManager : Object {
 
     public ModeManager (Akira.Lib.ViewCanvas canvas) {
         Object (view_canvas: canvas);
+
+        // any active mode must be ended when user presses Esc.
+        view_canvas.window.event_bus.request_escape.connect (() => {
```
mbfraga, line 58: For now we can keep this, but this is not the correct implementation of this. We are already filtering key events, so there should be no reason to listen to the event. I may be wrong though, since the event_bus event may trigger outside of explicit esc calls.

src/Lib/Modes/ItemInsertMode.vala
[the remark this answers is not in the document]
mbfraga: We can keep that for now. But not sure it makes sense to filter events AND listen to that signal. I am not even sure why that signal exists.

src/Lib/Modes/PathEditMode.vala
```diff
@@ -0,0 +1,160 @@
...
+            new_path_points[i] = old_path_points[i - 1];
+        }
+
+        instance.components.path = new Lib.Components.Path.from_points (new_path_points);
+    }
+
+    /*
+     * Recalculates the extents and updates the ViewLayerPath
+     */
+    private void update_view () {
```
mbfraga, line 135: This should be done by overloading the existing "update ()" method in the corresponding drawable. (see ViewLayersHover.update () )

mbfraga (changes requested):

src/Lib/Modes/PathEditMode.vala
AshishS-1123: @mbfraga I tried to do this similar to **ViewLayerHover** by passing the drawable to **ViewLayerPath** and getting bounds and points from it.
But that leads to a bug. Placing points in a zig zag manner outside the current bounds causes the points to shift from their original position. It will take some time to figure why that's happening. 
So for now, I will manually calculate the bounds, pass it to a ViewLayerPath to update its variables and then call update from there.

mbfraga: ok, I can take a look at this later. no problem.

mbfraga (approved):
thanks!

PR #663, opened 2021-09-03
Path Actions and Curves

AshishS-1123:

The following features have or will be added to Path tool in this PR.

- [x] Double click a **ModelTypePath** to enter PathEditMode.
- [x] When creating paths, press ```Backspace``` to delete the last inserted point.
- [ ] Create Curves.
- [ ] Rotation and Scaling of Path should draw points in PathEditMode correctly.


Alecaddd:
This looks very good as initial step towards a great path tool.
Thank you so much.

PR #668, opened 2021-10-31
Modify Points in Vector Tool and Other Features

AshishS-1123:

1. Fixed glitch in live path when drawing curves.
2. In edit mode, user can select a point and move it. (Edit mode is the path edit mode when you double click on a Path)
The selected point turns red to indicate that it was selected.
3. User can add more points to selection by holding the ```Shift``` key. 
Note that, selecting the first point of curve (to which both the tangents are connected) will automatically select the tangents.
4. If one of the tangents is selected, then the other one will be selected too. Moving one of those tangents will move the other one in a reflected manner. Note that, if the tangents and first point of curve don't lie on a line, only 1 tangent will be selected.
5. If ```Alt``` is clicked when selecting a tangent, then only one tangent is selected.
6. To close a path, you can either place the last point of path on first or press ```C```.

#### Moving Points in Path
1. Create a path with lines and enter edit mode by double clicking.
2. Click on a point and with the button still held down, move the point around.

#### Selecting Multiple Points
1. Create a path with lines and enter edit mode by double clicking.
2. Click on a bunch of points with ```Shift``` key held down.
3. Then release the ```SHIFT``` key, click on one of the selected point and drag it. All the selected points should move too.

#### Selecting Curves
1. Create a path with curves and enter edit mode by double clicking.
2. Click on the **CURVE_BEGIN** point and move it around. The selected point along with the tangents should move.
3. Same can be done for **CURVE_END** point.

#### Selecting Tangents
1. Create a path with curves and enter edit mode by double clicking.
2. Click on one of the tangents. The other should get selected too. 
3. Moving this tangent will move the other one in a reflected manner. Moving the selected tangent moves the other one such that it seems the second tangent is a mirror image of first w.r.t the **CURVE_BEGIN** point.
5. Hold down the ```Alt``` key and click on a tangent. This will select only the clicked tangent.
6. Once you move one of the tangent such that it is not in a straight line with the other tangent and the curve point, the tangents will no longer move in a reflected manner.

#### Selecting Multiple Curves
This is the same as selected multiple lines, except, selecting a **CURVE_BEGIN** point will also select the tangent points.

##### Method 1
Create a path using lines and curves, but instead of pressing ```Esc``` to exit the create mode, press ```C``` to join first and last point.

##### Method 2
Create a path using lines and curves, but instead of pressing ```Esc``` to exit the create mode, click on the first point. This will close the path and exit the mode.

##### Method 3
Create a path using lines and curves, then exit create mode by pressing ```Esc```. Double click on the path to enter edit mode.
Now drag and place the last point of curve on the first point. This will close the path and exit the mode.


mbfraga (changes requested):

src/Lib/Modes/PathEditMode.vala
```diff
@@ -145,12 +142,117 @@ public class Akira.Lib.Modes.PathEditMode : AbstractInteractionMode {
     }
 
     public override bool button_press_event (Gdk.EventButton event) {
-        if (!is_edit_path) {
+        // If button is clicked in edit mode, check if some point was clicked.
+        if (is_edit_path) {
+            int[] index = new int[3];
+            index[0] = index[1] = index[2] = -1;
+
+            var sel_type = edit_model.hit_test (event.x, event.y, ref index);
```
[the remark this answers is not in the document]
AshishS-1123: @mbfraga Don't actually need to do this. Hit test on zoom works properly.
This is because the event is scaled and then propagated from the **ViewCanvas** itself.

src/Lib/Modes/PathEditMode.vala
```diff
@@ -39,7 +55,8 @@ public class Akira.Lib.Modes.PathEditMode : AbstractInteractionMode {
     private int live_idx;
 
     // This flag tells if we are adding a path or editing an existing one.
-    private bool is_edit_path = true;
+    // Basically decides is we are in "edit" mode or "create" mode.
+    private bool is_edit_path = false;
```
mbfraga, line 59: Probably should be an enum, since there will probably be more submodes for path editing

mbfraga: create mode should probably just be called append or add

src/Lib/Modes/PathEditMode.vala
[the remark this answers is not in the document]
mbfraga: Zooming in past a certain level completely breaks editing--so I think something is still broken. Even so...we really need to get rid of scaling the event since that's bad behavior. i'll probably do that sooner rather than later.

src/Lib/Modes/PathEditMode.vala
```diff
@@ -145,12 +142,117 @@ public class Akira.Lib.Modes.PathEditMode : AbstractInteractionMode {
...
+                            edit_model.set_selected_points (index[i], true);
+                        }
+                    }
+
+                    edit_model.reference_point = index[0];
+                }
+            }
+
+            // Tangents behave in a different way. So we are placing the logic for that seperately.
+            if (sel_type == PointType.TANGENT_FIRST || sel_type == PointType.TANGENT_SECOND) {
```
mbfraga, line 195: This function is too long and does too much. separate.

src/Lib/Modes/PathEditMode.vala
```diff
@@ -192,7 +314,11 @@ public class Akira.Lib.Modes.PathEditMode : AbstractInteractionMode {
     public override bool motion_notify_event (Gdk.EventMotion event) {
         Geometry.Point point = Geometry.Point (event.x, event.y);
 
-        if (!is_edit_path) {
+        if (is_edit_path) {
+            if (!edit_model.selected_pts.is_empty && is_click) {
+                // If user selected a point and clicked and dragged it, change its position.
+                edit_model.modify_point_value (point);
```
mbfraga, line 320: we are not snapping to pixel grid. And after adding the second point, the whole thing moves.

src/Models/PathEditModel.vala
```diff
@@ -117,17 +128,171 @@ public class Akira.Models.PathEditModel : Object {
...
+        if (delta_x <= 2 && delta_y <= 2) {
+            make_path_closed ();
+            return true;
+        }
+
+        return false;
+    }
+
+    private bool compare_points (Geometry.Point a, Geometry.Point b) {
+        double thresh = 4 / view_canvas.scale;
```
mbfraga, line 284: This should be passed through, not gotten from here.

mbfraga, line 284: The scale should be passed through, not gotten from here.

src/Models/PathEditModel.vala
```diff
@@ -117,17 +128,171 @@ public class Akira.Models.PathEditModel : Object {
...
+        double delta_x = Math.ceil ((a.x + first_point.x - b.x).abs ());
+        double delta_y = Math.ceil ((a.y + first_point.y - b.y).abs ());
```
mbfraga, line 286: Should just use distance from geometry math or similar. If you use magic numbers (4), make them a private static int and name it so that people know what it is.

src/ViewLayers/ViewLayerPath.vala
```diff
@@ -119,14 +119,28 @@ public class Akira.ViewLayers.ViewLayerPath : ViewLayer {
                     context.fill ();
                 }
 
-                context.move_to (points[point_idx + 1].x + reference_point.x, points[point_idx + 1].y + reference_point.y);
```
mbfraga, line 122: Something is really buggy about rendering the edit mode. I think this needs to be stabilized. Control circles disappear randomly. In some cases lines don't show up (this may be an issue with the patheditmode invariants).

mbfraga (changes requested):

src/ViewLayers/ViewLayerPath.vala
AshishS-1123: Nope. Not buggy. Just forgot to write some lines in the view layer.

src/Lib/Modes/PathEditMode.vala
AshishS-1123: @mbfraga Can you elaborate as to what the problem is? I haven't worked with pixel grid yet. So I don't understand what exactly you want me to do here.

mbfraga:
Editing rotated vectors is completely broken.

src/Lib/Modes/PathEditMode.vala
mbfraga: Akira works mostly by abiding to the pixel grid. The core idea is that anything you add will generally be snapped to the pixel grid--which makes it good for UI design (figma). So when you add points to a path, it should always be choosing the best pixel position, and not the arbitrary location that was clicked. 

For better understanding--go to figma and zoom at the maximum level and add a path and see the positions where you are actually allowed to add points to. Now do the same in Akira to see the difference (enable the grid view in Akira--maybe it's the default?)

src/ViewLayers/ViewLayerPath.vala
mbfraga: Screenshot from 2021-12-18 13-59-04

See artifacts in here. It seems the more important issues are resolved now, thanks! but these artifacts should not be hard to resolve. I'm assuming the circle radius is not being applied to the extents correctly--just a guess.

AshishS-1123:
@mbfraga Any updates on the review for this PR? Getting a little anxious here!

src/Lib/Modes/PathEditMode.vala
```diff
@@ -50,10 +72,28 @@ public class Akira.Lib.Modes.PathEditMode : AbstractInteractionMode {
...
+        // If we are editing the path, we need to set the first point in edit_model
+        if (mode == Submode.EDIT) {
+            double center_x = instance.components.center.x;
+            double center_y = instance.components.center.y;
+            double width = instance.components.size.width;
+            double height = instance.components.size.height;
+
+            var first_point = Geometry.Point (center_x - width / 2.0, center_y - height / 2.0);
+
+            double rotation = instance.components.transform.rotation;
```
mbfraga, line 90: I mentioned this on element--but you can't do this. You have to apply the instance's transformation, not just the rotation. It is _never_ right to just apply rotation, You can test this by looking at transformed paths and you will see that editing is completely broken.

src/Lib/Modes/PathEditMode.vala
mbfraga: This is still not happening.

src/Models/PathEditModel.vala
```diff
@@ -48,19 +57,27 @@ public class Akira.Models.PathEditModel : Object {
...
 
-    public void add_live_points_to_path (Geometry.Point[] points, Lib.Modes.PathEditMode.Type live_command, int length) {
+    public void add_live_points_to_path (Geometry.Point[] live_pts, Lib.Modes.PathEditMode.Type live_command, int length) {
         commands.resize (commands.length + 1);
         commands[commands.length - 1] = live_command;
 
         for (int i = 0; i < length; ++i) {
-            var new_pt = Geometry.Point (points[i].x - first_point.x, points[i].y - first_point.y);
+            var new_pt = rotate_point_around_item_origin (live_pts[i], -instance.components.transform.rotation);
+            var orig_first_pt = rotate_point_around_item_origin (first_point, -instance.components.transform.rotation);
```
mbfraga, line 72: Again--only rotating is wrong.

mbfraga (changes requested):

PR #669, opened 2021-11-13
Guidelines Feature in Akira

AshishS-1123:
This PR will allow the user to draw guidelines.
Guidelines are simple horizontal or vertical lines that go from edge to edge on the canvas. They are mostly used in Image To HTML conversion for responsive design.
After this PR is done, you will be able to
- [x] Draw guidelines using ```Q``` and ```W``` keys.
- [ ] Move guidelines around by clicking and dragging.
- [ ] Measure distances between the selected guideline and its nearest neighbor of same direction.
- [ ] Measure the distance between the selected **CanvasItem** and the hovered guideline.

*Will add more items to this list as I go on.*

### DIFF
--- a/src/Lib/Managers/GuideManager.vala
+++ b/src/Lib/Managers/GuideManager.vala
@@ -1,0 +1,109 @@
+/**
+ * Copyright (c) 2019-2021 Alecaddd (https://alecaddd.com)
+ *
+ * This file is part of Akira.
+ *
+ * Akira is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+
+ * Akira is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+
+ * You should have received a copy of the GNU General Public License
+ * along with Akira. If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Authored by: Ashish Shevale <shevaleashish@gmail.com>
+ */
+
+ public class Akira.Lib.Managers.GuideManager : Object {
+    public enum Direction {
+        NONE,
+        HORIZONTAL,
+        VERTICAL
+    }
+
+    public unowned Lib.ViewCanvas view_canvas { get; construct; }
+    private GuideData guide_data;
+
+    public GuideManager (Lib.ViewCanvas view_canvas) {
+        Object (
+            view_canvas: view_canvas
+        );
+
+        guide_data = new GuideData ();
+    }
+
+     public bool key_press_event (Gdk.EventKey event) {
+        uint uppercase_keyval = Gdk.keyval_to_upper (event.keyval);
+
+        if (uppercase_keyval == Gdk.Key.Q) {
+            print("press q\n");
+            int x_pos = 0, y_pos = 0;
+            get_current_pointer_position (out x_pos, out y_pos);
+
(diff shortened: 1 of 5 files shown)


Alecaddd (changes requested):
This looks very good, thank you so much.
Upon using it, there's are some bugs that should be fixed, and some basic features that should be changed.
Here's the list of things I discovered, in no particular order.

**Bugs**
- Create an Artboard, move it around, delete it. It seems that some guides are created around the edges of the artboard and they remain visible.
- Guides are not deleted when an artboard is delete, but they should.
- When an artboard with guides is moved, the guides don't move with it, but they should as guides should be anchored to the top/left corner of an artboard.
- Guides disappear and appear inside and outside the artboard when the artboard is moved around.

**Features**
- A guide should be deleted only when the user releases the mouse click after moving it to the edge. If I move a guide to the edge, but keep my mouse clicked, I expect to be able to move it back into the artboard since I didn't let go.
- Guides should be usable also in the entire canvas, not just limited to artboards. Canvas guides and artboard guides shouldn't interfere with each other.

Let me know what you think.
Great work!

AshishS-1123:
> Guides should be usable also in the entire canvas, not just limited to artboards. Canvas guides and artboard guides shouldn't interfere with each other.

Haven't figured out how to do this yet. Will work on it in the next PR.
I have done all the other things you asked for, and it should be working properly now. 

Let me know if there are any other changes.

src/Utils/SortedArray.vala
```diff
@@ -0,0 +1,200 @@
...
+
+ /*
+  * This data structure will be used for storing the positions of guidelines.
+  * Storing all elements in a sorted array will make it easier to check if a guideline
+  * exists at the given position, to get the nearest neighbourest neightbours.
+  * Since this data structure does not allow storing duplicate values,
+  * placing a guideline over another will automatically delete it.
+  *
+  */
+ public class Akira.Utils.SortedArray : Object {
```
mbfraga, line 30: I have several issues with this.
1. This is not really a sorted array as such. It's a set with bounds.
2. It's very specifically programmed for guidelines, which makes the generic looking name misleading
3. It's not very efficient, even for binary search (this could be easily fixed)
4. It probably makes more sense to use a Gee.TreeSet in a wrapper

mbfraga (changes requested):
The architecture here seems not ideal, but I think it's fixable. Mainly the issue I have is storing some guidelinemodels in the instance and other in the manager. I think the ModelInstance should have the minimal amount of data needed to generate guidelines, and then the guidelinemanager would generate those based on instances. 

I think that if we want this, we should start with guidelines ONLY in the canvas for the time being and try to get that right.

src/Utils/SortedArray.vala
AshishS-1123: I wrapped a TreeSet in the SortedArray. Also renamed SortedArray to SortedGuides. 

> Mainly the issue I have is storing some guidelinemodels in the instance and other in the manager

I don't understand what this means. The GuidelineManager only stores a reference to the models from the instance. This way, based on the events, the guide data for the artboard instance is directly changed.

Alecaddd (changes requested):
file:///home/alecaddd/Pictures/Screenshots/Screenshot%20from%202021-12-12%2021-36-24.png

There's an issue with the artboard.
It seems that guides are generated immediately around the edges of an artboard when the selection changes.
You can see these guides above the nobs.

AshishS-1123:
These are the changes made in the latest commit.
1. Place the guide layer below the nobs layer.
2. Instead of each artboard storing guide data individually, the guide manager itself stores a map of artboard id's and their data.
3. When the artboard is being modified (scaled or moved), the guides are hidden.

*P.S. The guides don't work very well if the artboard is rotated. I don't think artboard are allowed to be rotated .So I haven't handled that case.*

PR #672, opened 2021-11-19
Layers list improvements

Alecaddd:
Now that we have a layers listbox with some basic selection functionalities, it's time to improve the performance and add missing features.

- [x] Improve the layers creation performance and selection performance.
- [x] Double click to edit the layer name.
- [x] Add hover effect in both direction (item to layer | layer to item).

### DIFF
--- a/src/Layouts/LayersList/LayerItemModel.vala
+++ b/src/Layouts/LayersList/LayerItemModel.vala
@@ -41,6 +41,12 @@ public class Akira.Layouts.LayersList.LayerItemModel : GLib.Object {
         owned get {
             return node.components.name.name;
         }
+        set {
+            if (value == name) {
+                return;
+            }
+            node.components.name = new Lib.Components.Name (value, id.to_string ());
+        }
     }
 
     /*
--- a/src/Layouts/LayersList/LayerItemModel.vala
+++ b/src/Layouts/LayersList/LayerItemModel.vala
@@ -39,7 +39,7 @@ public class Akira.Layouts.LayersList.LayerItemModel : GLib.Object {
      */
     public string name {
         owned get {
-            return node.id.to_string (); // Temporarily use the ID.
+            return node.components.name.name;
         }
     }
 
--- a/src/Layouts/LayersList/LayerListBox.vala
+++ b/src/Layouts/LayersList/LayerListBox.vala
@@ -54,7 +54,7 @@ public class Akira.Layouts.LayersList.LayerListBox : VirtualizingListBox {
                 row = new LayerListItem ();
             }
 
-            row.assign ((LayerItemModel)item);
+            row.assign ((LayerItemModel) item);
             row.show_all ();
 
             return row;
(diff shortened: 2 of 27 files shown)


Alecaddd:
I'll do the removal of child layers in a follow up PR since this is big enough, and also because I'm thinking maybe it's better to use the `ModelNode` as layer model instead of the `ModelInstance`.

src/Layouts/LayersList/LayerItemModel.vala
```diff
@@ -39,7 +39,13 @@ public class Akira.Layouts.LayersList.LayerItemModel : GLib.Object {
...
     public string name {
         owned get {
-            return node.id.to_string (); // Temporarily use the ID.
+            return node.components.name.name;
+        }
+        set {
+            if (value == name) {
+                return;
+            }
+            node.components.name = new Lib.Components.Name (value, id.to_string ());
```
Alecaddd, line 48: @mbfraga is this correct?
I feel like it doesn't follow the approach other components are replaced as I'm changing it directly inside the Layer Model, which is its own thing separated from the Item Model, but I thought it was fine since the name can't be really edited from within the canvas without a layer UI.

src/Layouts/LayersList/LayerListBox.vala
```diff
@@ -52,9 +53,10 @@ public class Akira.Layouts.LayersList.LayerListBox : VirtualizingListBox {
                 row = old_widget as LayerListItem;
             } else {
                 row = new LayerListItem ();
+                row.row_updated.connect (on_row_updated);
```
Alecaddd, line 56: I'm not super happy about this as I was trying to avoid having a signal per layer as much as possible, that's why the hovered and edited signals are coming from the listbox and triggered on click.
I'm not really sure how to avoid this and listen for when the user presses Enter and edits the layer's name.

mbfraga: Not sure I understand why this is necessary. There should probably be one editor per layerlistbox that gets listened to, and an active row. For now if this works, it's not a big deal tbh.

mbfraga (commented):
Added some comments. I think those things could be improved, but they are not current deal breakers as far as I can tell.

src/Layouts/LayersList/LayerItemModel.vala
Alecaddd: Can you point me to where in the current code we do something like that?

src/Layouts/LayersList/LayerListBox.vala
Alecaddd: Yeah, I was having issues in finding a proper approach to having an editor listening to the active row and handling the editing of the name, same way I do for the hover and select signal, which are handled by the listbox itself and not the single row.
I'll try to rework this.

src/Layouts/LayersList/LayerItemModel.vala
Alecaddd: This? https://github.com/akiraux/Akira/blob/main/src/Utils/AffineTransform.vala#L80-L82

PR #676, opened 2021-12-04
Groups and Artboards layers UI

Alecaddd:
Show child layers of groups and artboards properly indented in the Layers Panel.

- [x] Delete child layers when a parent is deleted.
- [x] Group child layers underneath their parents.

### DIFF
--- a/src/Lib/Items/Model.vala
+++ b/src/Lib/Items/Model.vala
@@ -356,7 +356,7 @@ public class Akira.Lib.Items.Model : Object {
         );
         var new_node = new ModelNode (candidate, (int) pos);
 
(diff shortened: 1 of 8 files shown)


src/Lib/Items/Model.vala
```diff
@@ -381,6 +379,8 @@ public class Akira.Lib.Items.Model : Object {
 
         new_node.parent = parent_node;
 
+        add_to_maps (new_node, true);
+
```
Alecaddd, line 383: @mbfraga I had to move the adding of the node to the canvas after the parent was assigned in order to create layers in the proper position.
Is this okay? It doesn't seem to affect performance or anything else/

PR #679, opened 2021-12-18
Initial Implementation for Pencil Tool (Freehand Drawing)

AshishS-1123:

The Pencil Tool will allow users to draw freehand shapes and squiggles.

When drawing such a canvas item, all the users pointer events would be captured and stored. When rendering these points, we join them using a polyline. While this makes the task of creating such squiggles easy, storing large amounts of points becomes a problem.

To reduce the number of points, we use **Schneider's Algorithm**. This algorithm will compress the set of points into a set of bezier curves that are very close to what the user originally drew.

Using this algorithm will do two things-
1. Reduce the number of points.
2. Make the drawing more aesthetically pleasing. (because bezier curves are very smooth)..

##### Reference
1. The Schneider's Algorithm can be found in the book Graphics Gems I. I found a pdf [here](https://www.pdfdrive.com/graphics-gems-1-e38625848.html)
2. The actual code was a little hard to develop. So I referred [this](https://github.com/soswow/fit-curve/blob/master/src/fit-curve.js) repo. The code here is in JavaScript which I translated to Vala as best as I could. I also ran this implementation and my own one on different inputs and checked their outputs. They matched more or less.

#### Summary of major changes I made

- Created **Lib.Modes.FreehandMode** to get user events for creating freehand drawings.
- Created **Models.FreehandModel** to store and process the raw points from user events.
- Used *Schneider's Algorithm** to optimize the number of points needed to represent the drawing. Approximates the raw points into a set of bezier curves.

## Steps to Test

- Press ```P``` or use the **Insert** menu to start the pencil tool.
- Hold down your mouse button and get crazy.
- As soon as you are done drawing, release the mouse button. The curve should have formed.
- Double clicking on this item will show the control points of bezier curves formed.

## Screenshots 

https://user-images.githubusercontent.com/64722310/146642939-5334d9b2-70ac-474d-a943-b20e1cfd6a87.mp4

### DIFF
--- a/src/Drawables/DrawablePath.vala
+++ b/src/Drawables/DrawablePath.vala
@@ -73,6 +73,23 @@ public class Akira.Drawables.DrawablePath : Drawable {
                 cr.curve_to (x1, y1, x3, y3, x4, y4);
 
                 point_idx += 4;
+            } else if (commands[i] == Lib.Modes.PathEditMode.Type.BEZIER) {
+                var x1 = points[point_idx].x;
+                var y1 = points[point_idx].y;
+
+                var x2 = points[point_idx + 1].x;
+                var y2 = points[point_idx + 1].y;
+
+                var x3 = points[point_idx + 2].x;
+                var y3 = points[point_idx + 2].y;
+
+                var x4 = points[point_idx + 3].x;
+                var y4 = points[point_idx + 3].y;
+
+                cr.move_to (x1, y1);
+                cr.curve_to (x2, y2, x3, y3, x4, y4);
+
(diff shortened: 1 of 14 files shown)


mbfraga (changes requested):
I have not yet reviewed the FreehandMode. There seems to be some general bugginess with some specific traces. Other times it works really nicely. Not sure what exactly triggers the bad behavior. See the screenshot below, where the blue represents roughly the trace I did with my mouse--and the black the result.
Screenshot from 2021-12-18 14-09-41
.

PR #682, opened 2021-12-20
Compile after remove

mbfraga:
When removing items from artboards a critical error occured because the artboard had been marked as dirty but had not been recompiled.d

## Steps to Test
Create an artboard, give it a few items. Delete those items and then select the artboard.
result: critical error
expected: no error


Alecaddd (approved):
Sweet, thanks.

PR #683, opened 2021-12-27
Add alignment panel, logic & anchor alignment

giacomoalbe:

This PR adds the Alignment Panel to the left sidebar, which I decided to name `PropertiesSidebar` in order to follow the naming convention for the other sidebar. 

AlignmentPanel buttons are enabled only when the selection is a multi selection. 
By default, the alignment type is auto, but it is possible to select an item in the selection by simply clicking on it to make the alignment relative to that item (called "anchor"). 

## Steps to Test
 

* Draw some items on the canvas
* Select two or more of them by SHIFT clicking 
* Use the buttons in the AlignmentPanel on the Properties sidebar to align the items
* Click (without holding SHIFT) on one of the selected items in order to align the items relative to it

## Screenshots 

https://user-images.githubusercontent.com/11556031/147503903-29723385-7049-47de-9e5d-290522772833.mp4

Is `PropertiesSidebar` a good name for the previously named `LeftSidebar`?


src/ViewLayers/ViewLayerAnchor.vala
```diff
@@ -0,0 +1,64 @@
...
+ * GNU General Public License for more details.
+
+ * You should have received a copy of the GNU General Public License
+ * along with Akira. If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Authored by: Giacomo "giacomoalbe" Alberini <giacomoalbe@gmail.com>
+ */
+
+public class Akira.ViewLayers.ViewLayerAnchor : ViewLayer {
+    public const double UI_LINE_WIDTH = 4.0;
```
mbfraga, line 23: I actually think this should be part of the ViewLayerNob. Perhaps "ViewLayerNob" is the wrong name, but I think it makes more sense associated with that mechanism.

src/Lib/Managers/SelectionManager.vala
```diff
@@ -51,14 +51,21 @@ public class Akira.Lib.Managers.SelectionManager : Object {
     }
 
     public Lib.Items.NodeSelection selection;
+    public int? alignment_anchor_id;
```
mbfraga, line 54: Maybe we should give this a more generic name and keep it open for potentially other features? Perhaps calling it subselection?

src/ViewLayers/ViewLayerAnchor.vala
```diff
@@ -0,0 +1,64 @@
...
+        if (canvas == null || drawable == null) {
+            return;
+        }
+        var color = Gdk.RGBA () { red = 0.25, green = 0.79, blue = 0.98, alpha = 1.0 };
+        drawable.paint_hover (context, color, UI_LINE_WIDTH, target_bounds, scale);
+        last_drawn_bb = drawable.bounds;
+    }
+
+    public override void update () {
+        if (canvas == null) {
```
mbfraga, line 51: If this gets moved to the ViewLayerNobs, then it won't be an issue; but right now this update wouldn't work generically because it wouldn't account for resizes caused by the item moving. Right now it works because the ViewlayerNob requests the update itself. That's another reason to keeping it part of the ViewLayerNob.

mbfraga (changes requested):

Alecaddd:
> Is `PropertiesSidebar` a good name for the previously named `LeftSidebar`?

I think `OptionsSidebar` sounds better.

src/ViewLayers/ViewLayerAnchor.vala
giacomoalbe: Ok got it! I just need to add another "case", in order to draw the anchor selection border around the anchor item. Would it be better to just call the file ViewLayerSelection? It does draw everything related to selection, imho it's a better fit than just referencing the Nob alone

src/ViewLayers/ViewLayerAnchor.vala
giacomoalbe: Ok!

src/Lib/Managers/SelectionManager.vala
giacomoalbe: To me it's ok, this was only intended for the alignment anchor thing, but I can change it to `sub_selection_id`.

giacomoalbe:
Ok, i solved the issues and renamed the "properties" sidebar to "options": 

The only thing I'm not completely ok with is the fact that whenever a redraw of the selected items occurs the selection bound and the nobs seem to "flicker", and since I added the sub_selection to the `nob_layer` this happens to the sub_selection rect too. 

Nevertheless, I think the PR now could be merged, let me know what you think!

mbfraga (approved):

src/Layouts/Partials/AlignItemsPanel.vala
```diff
@@ -19,48 +19,47 @@
  * Authored by: Giacomo "giacomoalbe" Alberini <giacomoalbe@gmail.com>
  */
 public class Akira.Layouts.Partials.AlignItemsPanel : Gtk.Grid {
-    public weak Akira.Window window { get; construct; }
+    public unowned Lib.ViewCanvas view_canvas { get; construct; }
 
     public int current_button_column { get; set; default = 0; }
```
Alecaddd, line 24: We can make this `private` since it's only used inside this class.

src/Layouts/Partials/AlignItemsPanel.vala
```diff
@@ -19,48 +19,47 @@
...
-        AlignBoxItem ("btn", "alig-h-l", "align-horizontal-left", _("Align Left"), {"<Ctrl><Shift>3"}),
-        AlignBoxItem ("btn", "alig-h-c", "align-horizontal-center", _("Align Center"), {"<Ctrl><Shift>4"}),
-        AlignBoxItem ("btn", "alig-h-r", "align-horizontal-right", _("Align Right"), {"<Ctrl><Shift>5"}),
-        AlignBoxItem ("sep"),
-        AlignBoxItem ("btn", "alig-v-t", "align-vertical-top", _("Align Top"), {"<Ctrl><Shift>6"}),
-        AlignBoxItem ("btn", "alig-v-c", "align-vertical-center", _("Align Middle"), {"<Ctrl><Shift>7"}),
-        AlignBoxItem ("btn", "alig-v-b", "align-vertical-bottom", _("Align Bottom"), {"<Ctrl><Shift>8"})
+        AlignBoxItem ("btn", Utils.ItemAlignment.AlignmentDirection.TOP, "align-vertical-top", _("Align Top"), {"<Ctrl><Shift>6"}),
+        AlignBoxItem ("btn", Utils.ItemAlignment.AlignmentDirection.VCENTER, "align-vertical-center", _("Align Middle"), {"<Ctrl><Shift>7"}),
+        AlignBoxItem ("btn", Utils.ItemAlignment.AlignmentDirection.BOTTOM, "align-vertical-bottom", _("Align Bottom"), {"<Ctrl><Shift>8"})
```
Alecaddd, line 57: These accelerators have changed since the SHIFT modifier was creating problems for different keyboard layouts.
Now we use `Alt+1...6`.
https://github.com/akiraux/Akira/blob/main/src/Services/ActionManager.vala#L174-L179

src/Layouts/Partials/AlignItemsPanel.vala
```diff
@@ -19,48 +19,47 @@
...
     };
 
-    public AlignItemsPanel (Akira.Window window) {
+    public AlignItemsPanel (Lib.ViewCanvas view_canvas) {
         Object (
-            window: window,
+            view_canvas: view_canvas,
             orientation: Gtk.Orientation.VERTICAL
```
Alecaddd, line 63: Nit: the orientation attribute is not necessary anymore when extending a Gtk.Grid

src/Layouts/Sidebars/OptionsSidebar.vala
```diff
@@ -0,0 +1,84 @@
...
+        } set {
+            visible = value;
+            no_show_all = !value;
+        }
+    }
+
+    public OptionsSidebar (Lib.ViewCanvas view_canvas) {
+        Object (
+            view_canvas: view_canvas,
+            orientation: Gtk.Orientation.HORIZONTAL
```
Alecaddd, line 40: nit: remove this as well

src/Layouts/Sidebars/OptionsSidebar.vala
```diff
@@ -0,0 +1,84 @@
...
+        show_all ();
+    }
+
+    private void toggle () {
+        toggled = !toggled;
+    }
+
+    private void on_widget_redraw_request () {
+        queue_resize ();
+    }
```
Alecaddd, line 83: Let's remove this as it comes from the old architecture and bad coding approach I introduced. We're gonna use a new listview which will not require a forced redraw on the whole widget.

src/Lib/ViewCanvas.vala
```diff
@@ -338,6 +339,11 @@ public class Akira.Lib.ViewCanvas : ViewLayers.BaseCanvas {
 
                     selection_manager.add_to_selection (target.id);
                     selection_manager.selection_modified_external ();
+                } else {
+                    if (selection_manager.selection.count () > 1) {
+                        // Don't trigger sub selection when only one item's selected
+                        nob_manager.toggle_sub_selection (target.id);
+                    }
```
Alecaddd, line 346: I think the sub selection should get triggered only if the user holds the <kbd>ALT</kbd> key.
Clicking on an item that is part of a multi selection should deselect the other items and only select the clicked one. (this is something not implemented yet but we should in a future PR).

src/ViewLayers/ViewLayerNobs.vala
```diff
@@ -35,6 +42,17 @@ public class Akira.ViewLayers.ViewLayerNobs : ViewLayer {
         update ();
     }
 
+    public void add_sub_selection (Drawables.Drawable? new_sub_selection_drawable) {
+        if (new_sub_selection_drawable == sub_selection_drowable) {
+            return;
+        }
+
+        old_sub_selection_drawable = sub_selection_drowable;
```
Alecaddd, line 50: little typo: `drAwable` instead of `drOwable`, a drow is a dark elf :D

src/Utils/ItemAlignment.vala
```diff
@@ -144,6 +148,15 @@ public class Akira.Utils.ItemAlignment : Object {
...
 
+                break;
+            case AlignmentType.ANCHOR:
+                unowned var bb = anchor.instance.bounding_box;
+
+                align_to.left = bb.left;
+                align_to.right = bb.right;
+                align_to.top = bb.top;
+                align_to.bottom = bb.bottom;
+                found_bound = true;
```
Alecaddd, line 159: Little English nit: every time we use the word `bound` it should actually be `bounds` as it represents the boundaries of an object, so it should be plural.

Alecaddd (changes requested):
I'd like to propose a little file restructuring just for clarity.
We should have:
- `src/Layout/Alignment`
- `src/Layout/Alignment/AlignmentPanel.vala`
- `src/Layout/Alignment/AlignmentButton.vala`

This is to keep all related files in the same organized location. Files in the `Widgets/` folder should be more abstract classes that are extended by other widgets in multiple locations, kinda like I did for the virtualizing listbox which will be used for layers, fills, borders, etc.

This looks very good, thank you so much for your hard work.
Just a few nits and fixes to address.

src/Layouts/Partials/AlignItemsPanel.vala
giacomoalbe: This is not possible due to the fact that the assignment is done using the Object function which does require the construct keyword in the property assignment and properties must be public. I'll leave it public for now, tell me if there is another method to deal with those cases.

src/Lib/ViewCanvas.vala
giacomoalbe: Coming from Illustrator, I tend to disagree with this behaviour. In Inkscape (which is what I use exclusively today) always relying on the UI select to choose which item to use as anchor is very frustrating. In Illustrator all I had to do was to click on the already selected item to make it become the anchor.

I think this approach is winning and way easier than having to press the ALT key. 

Also, I'm not very used with deselecting all the items is I click on an already selected item. Moreover, I think this could be annoying, imagine I was about to select 30 items one at a time and suddenly, by mistake, I would click on an already selected item with the not so desirable result that my entire selection has gone. 

With the anchor approach, I would have simply "elected" that item to become the anchor. If this is not the case, I could simply click it again and remove the anchoring. If I want to deselect it, I simply need to hold the SHIFT key. 

I think this approach is better. 
Let me know what you think, I can change the code, of course, but this particular feature was something I wanted so badly and I was not able to find in any other software but Illustrator (which is, you know, a good example :)) that I want to point out my thoughts about it.

Alecaddd: Thanks for the detailed explanation.
This is mostly a UX decision, and my proposal was coming from my past experience with Sketch and Figma.

I see your points and I think they're very valid, so I'm okay with following your suggested approach.

Alecaddd (approved):

PR #686, opened 2021-12-29
Add a basic text item

mbfraga:
Implement a simple implementation of a text item. It is not editable yet, but has enough in there to start building a TextEditMode on top of it.

### DIFF
--- a/src/Drawables/DrawableText.vala
+++ b/src/Drawables/DrawableText.vala
@@ -80,8 +80,8 @@ public class Akira.Drawables.DrawableText : Drawable {
 
         var layout = create_pango_layout (context, label, w, scale, null);
         context.set_matrix (global_transform);
-        context.rectangle(x, y, w, h);
-        context.clip();
+        context.rectangle (x, y, w, h);
+        context.clip ();
         context.move_to (x, y);
         Pango.cairo_show_layout (context, layout);
     }
--- a/src/Drawables/DrawableText.vala
+++ b/src/Drawables/DrawableText.vala
@@ -1,0 +1,123 @@
+/**
+ * Copyright (c) 2019-2021 Alecaddd (https://alecaddd.com)
+ *
+ * This file is part of Akira.
+ *
+ * Akira is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+
+ * Akira is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+
+ * You should have received a copy of the GNU General Public License
+ * along with Akira. If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Authored by: Martin "mbfraga" Fraga <mbfraga@gmail.com>
+ *
+ * Parts of this code are taken from GooCanvas
+ */
+
+/*
+ * Drawable for rects.
+ */
+public class Akira.Drawables.DrawableText : Drawable {
+
+    public DrawableText (double tl_x, double tl_y, double width, double height, string text) {
+       this.center_x = tl_x + width / 2.0;
+       this.center_y = tl_y + height / 2.0;
+       this.width = width;
+       this.height = height;
+       this.label = text;
+    }
+
+    public override void simple_create_path (Cairo.Context context) {
+        var w = width;
+        var h = height;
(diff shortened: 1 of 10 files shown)


src/Drawables/DrawableText.vala
```diff
@@ -0,0 +1,132 @@
+/**
+ * Copyright (c) 2019-2021 Alecaddd (https://alecaddd.com)
```
bilelmoussaoui, line 2: You probably want to put your information here

src/Drawables/DrawableText.vala
```diff
@@ -0,0 +1,132 @@
...
+       this.center_x = tl_x + width / 2.0;
+       this.center_y = tl_y + height / 2.0;
+       this.width = width;
+       this.height = height;
+       this.label = text;
+    }
+
+    public override void simple_create_path (Cairo.Context context) {
+        var w = width;
+        var h = height;
```
bilelmoussaoui, line 39: are those variables needed?

src/Drawables/DrawableText.vala
```diff
@@ -0,0 +1,132 @@
...
+
+    public override void paint (Cairo.Context context, Geometry.Rectangle target_bounds, double scale) {
+        // Simple bounds check
+        if (bounds.left > target_bounds.right || bounds.right < target_bounds.left
+            || bounds.top > target_bounds.bottom || bounds.bottom < target_bounds.top) {
+            return;
+        }
+
+        var w = width;
+        var h = height;
```
bilelmoussaoui, line 55: similarly here

src/Drawables/DrawableText.vala
```diff
@@ -0,0 +1,132 @@
...
+
+        if (layout_width > 0) {
+            layout.set_width ( (int)(layout_width * Pango.SCALE));
+        }
+
+        layout.set_text (text, -1);
+
+        // Load font description
+        var desc = new Pango.FontDescription ();
+        desc.set_family ("Open Sans");
```
bilelmoussaoui, line 116: This probably shouldn't be hardcoded, what happens if Open Sans isn't installed? You can probably get around this issue by getting this information from the parent widget's pango context.

src/Lib/Components/Text.vala
```diff
@@ -0,0 +1,42 @@
+/**
+ * Copyright (c) 2019-2021 Alecaddd (https://alecaddd.com)
```
bilelmoussaoui, line 2: Should have your information

src/Lib/Items/ModelTypeText.vala
```diff
@@ -0,0 +1,93 @@
+/**
+ * Copyright (c) 2019-2021 Alecaddd (https://alecaddd.com)
```
bilelmoussaoui, line 2: Same

src/Lib/Modes/ItemInsertMode.vala
```diff
@@ -203,24 +203,11 @@ public class Akira.Lib.Modes.ItemInsertMode : AbstractInteractionMode {
 
             case "text":
                 // TODO: Temporarily draw a path shape. We'll deal with text later.
```
bilelmoussaoui, line 205: this should be dropped or not?

src/Drawables/DrawableText.vala
Alecaddd: Nah, the copyright of every file is correct, then each author can write their name blow it in the Authored section.
Only change here is that this should only have 2021 since this file was created now.

src/Drawables/DrawableText.vala
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        Cairo.Matrix global_transform = context.get_matrix ();
+        var tr = transform;
+        context.set_matrix (tr);
+
+        var layout = create_pango_layout (context, label, w, scale, null);
+        context.set_matrix (global_transform);
+        context.rectangle(x, y, w, h);
+        context.clip();
```
Alecaddd, line 93: nit: missing space before parenthesis.

Alecaddd (changes requested):
Looking like a great start.
Could you please add this snippet to the `LayerItemModel.vala` in the `icon` getter?
```
else if (type is Lib.Items.ModelTypeText) {
    return "shape-text-symbolic";
}
```
With this we get the proper icon for the text layer.
Great work!

src/Drawables/DrawableText.vala
mbfraga: I assume that if open sans isn't install, the closest one will be used by pango. This item is meant to be a starting place for a lot of work that needs to go into text items (editing, etc). Ideally a font will be passed down as a component once the proper implementation is done.

Alecaddd (approved):
Great start!

PR #687, opened 2021-12-31
Generic alert change

mbfraga:
Simplifies process of marking items as changed and nullifying compiled components.

The system still need to be implemented in a better way. But there are still some open questions about groupings that make it a bit hard to stabilize until we cross those bridges.


Alecaddd (approved):
Looks great, thanks!

PR #688, opened 2021-12-31
Add fill panel for single and multiple items

giacomoalbe:

This PR will restore the FillPanel, for single and multiple items. 

In the beginning I though I restored also the BorderPanel, but in the end I think this PR is wide enough for it to be tested separately, and the smaller the change the better :)

## Steps to Test
 

* Create one or more fillable items (rect or ellipses, for example)
* Select one of them
* Change the color by editing the ColorRow
* Select multiple items, now the ColorRow should show a question mark on the color. Choose a color to change it for every selected item

## Screenshots 

Single item color change

https://user-images.githubusercontent.com/11556031/147827072-41ab1772-5c7a-438a-bf88-b20cd04fe0fc.mp4

Multiple items with same fill/color

https://user-images.githubusercontent.com/11556031/147827078-4e0c51db-d32e-4729-9b4f-6d6eeb25a94d.mp4

Multiple items with different colors

https://user-images.githubusercontent.com/11556031/147827081-a1597743-afbb-48bd-978e-f814ee61e8b5.mp4

This update consider just the case of items with one fill. The add multiple fill option is not working, and neither the delete or hide fill. Only one fill at a time is working atm. 

I think that, for now, this should be enough, we can add it later, tell me what you think


tmux-client-173868.log
```diff
@@ -0,0 +1,90 @@
+1640792588.737343 client started (173868): version 3.0a, socket /tmp//tmux-1000/default, protocol 8
```
mbfraga, line 1: lol

mbfraga (changes requested):

tmux-client-173868.log
Alecaddd, line 1: What's this? :P

src/Lib/Components/Fill.vala
```diff
@@ -19,14 +19,16 @@
  * Authored by: Martin "mbfraga" Fraga <mbfraga@gmail.com>
  */
 
-public class Akira.Lib.Components.Fill {
+public class Akira.Lib.Components.Fill : GLib.Object {
```
Alecaddd, line 22: This should extend `Component, Copyable<Name>` like the other components.

Alecaddd (changes requested):
Thanks for starting to work on this, but unfortunately the approach is not correct since we shouldn't use the same structure and widgets we're using on `master`.

We should follow what we're doing with the layers UI, by leveraging the VirtualizingListBox, alongside creating a `Layout/Fills` folder which contains the list item, list model, list box, and so on.

Also, we should absolutely keep the multiple fills feature.

tmux-client-173868.log
giacomoalbe: I wanted to let you aware of my tech stack :) I'll get rid of it

Alecaddd:
Gently closing this in favor of #709

PR #690, opened 2021-12-31
Multi selection performance

Alecaddd:
- Improve the multi selection performance by preventing scroll+selection of the entire listbox if not necessary.
- Improve the hover effect performance on layers by preventing triggering the update of the hover effect in the canvas if we're hovering over the same layer.


giacomoalbe (approved):
I tested the code locally and the speed improvement of selecting many items is more than visible. 

I think that for now is pretty acceptable, in the end I'd say it's comparable with other software when it comes to big numbers of selected items. 

Good work!

PR #694, opened 2022-01-07
Block UI and selection improvements

Alecaddd:
Some much needed improvements to the layers panel.

## Steps to Test
- Create items with and without the UI by toggling the presentation mode with <kbd>CTRL</kbd>+<kbd>.</kbd>
- Toggle the UI before and after creating various item.
- Confirm the layers are restored correctly.

- Block the creation of layers when the app is in Presentation Mode.
- Fix selection performance issues by setting the selected state to the layers model without requiring the generation/reveal of row widgets.


Alecaddd:
Pulling a YOLO and merging this as it fixes a lot of annoying performance issues and allows us to better discern if the UI is dramatically affecting the performance of the canvas.

PR #699, opened 2022-01-24
Inefficient implementation of undo/redo

mbfraga:
Implement basic undo/redo. code is somewhat inefficient...but honestly fairly ok considering other limitations currently in the software. I don't guarantee this is bug-free ;)

## Steps to Test
Undo with ctrl-z, redo with ctrl-shift-z

Todo:
* probably should limit number of undos/redos
* There is probably a way to reduce special code to make sure
things are copied and not just reference-shared...vala...

### DIFF
--- a/src/Lib/Items/Model.vala
+++ b/src/Lib/Items/Model.vala
@@ -56,12 +56,7 @@ public class Akira.Lib.Items.Model : Object {
 
     private unowned ViewLayers.BaseCanvas? canvas = null;
 
-    public Model.live_model (ViewLayers.BaseCanvas? canvas) {
-        is_live = true;
-        this.canvas = canvas;
-    }
-
-    construct {
+    public Model () {
         item_map = new Gee.HashMap<int, ModelInstance> ();
         group_map = new Gee.HashMap<int, ModelInstance> ();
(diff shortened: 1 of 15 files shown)


src/Lib/Items/Model.vala
```diff
@@ -56,12 +56,7 @@ public class Akira.Lib.Items.Model : Object {
 
     private unowned ViewLayers.BaseCanvas? canvas = null;
 
-    public Model.live_model (ViewLayers.BaseCanvas? canvas) {
-        is_live = true;
-        this.canvas = canvas;
-    }
-
-    construct {
+    public Model () {
```
Alecaddd, line 59: I'm wondering if there's a particular reason why you converted this from a construct creation to a class creation. The `construct` always runs before any class with argument is called, so there's not need to call `this()`.
Did you stumble upon any issue with it?

Alecaddd (approved):
Great initial implementation!
The only issue is that the app crashes if I create a new item after an `undo` action. This is due to the layers item not updating properly at every history interaction and freaking out when trying to update the list.
I'll deal with it in a follow up PR, I'm okay with merging this.

PR #700, opened 2022-01-26
Reimplement the boilerplate UI of the transform panel

Alecaddd:
Re-implement the transform panel without connecting anything to the items.
Just the UI work.

### Screenshot
image


Alecaddd:
YOLO merge for this since it's only UI boilerplate and doesn't interact with the model.

PR #707, opened 2022-02-04
Implement 705: Undo should not get stacks in no-ops

### LINKED ISSUE #705: Operations should only add undo stacks if necessary
## Expected Behavior
Some operations will currently add undo stacks even though they may lead to no-ops. For example: if you try to shift a top-most item, you will get an undo stack even though it's a no-op.

## Man culprits
* Change z-order of items
* Align items via Alt+{1,2,3,4,5,6}

## Proposal to fix
When running a potential operation do the following steps:
1. Pre-process an operation over all candidates and find the affected node/instance and any data needed for the operation.
2. If there are no valid ops, return out.
3. Otherwise add an undo stack and execute the operations.

1. Add a few times
2. Click on the last item added (topmost)
3. Press Ctrl+Up several times (each will add an undo stack)
4. Undo multiple times until you see a change

mbfraga:
Don't trigger unnecessary undos on some ops.

Shifting items (z-order) don't trigger undo stack adds in no-ops

Added undo to aligning items, and added code to make sure they
also don't add undo stacks in no-ops.

## Steps to Test
Shift top-most item multiple times and make sure undos work correctly.
Run align-items multiple times and check that undo works correctly

- Fixes #705

### DIFF
--- a/src/Lib/Items/Model.vala
+++ b/src/Lib/Items/Model.vala
@@ -257,7 +257,25 @@ public class Akira.Lib.Items.Model : Object {
         return inner_splice_new_item (parent_node, pos, candidate);
     }
 
-    public int move_items (int parent_id, uint pos, uint newpos, int length, bool restack) {
+    /*
+     * Move items within a parent to change their z-order.
+     * 
+     * Set restack to false if several move operations will be executed, which could make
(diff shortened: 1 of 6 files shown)


Alecaddd (approved):
This looks great!

PR #708, opened 2022-02-04
Refactor Path Segment Representation

AshishS-1123:

- [x] Create a PathSegment to represent different types of segments in path. This way, points and commands will be stored together.
- [x] Refactor existing code to use this PathSegment.
- [x] Cubic and Quadratic Beziers.

#### Test new path representation
With new representation, all the old features should still be working.

#### Quadratic Curves
A **click-drag-click** action creates a CUBIC bezier.
A **click-drag-click-drag . . .** action creates QUADRATIC bezier.

#### Interconversion
Create a path using lines and curves and enter edit mode by double clicking on the canvas item.
Double click on any point. Curves should turn into lines and vice versa.

*Note: Sometimes double click does not seem to have any effect. I think this is because of some issue of path in ItemManager.node_at_canvas_position but haven't pinpointed this issue.*

## Screenshots 

Screenshot below shows 3 paths. The top left is made from lines only, the top right from CUBIC beziers.
The bottom-most is created using QUADRATIC beziers.

refactor


src/Utils/PathSegment.vala
```diff
@@ -0,0 +1,302 @@
...
+ * 1. Symmetric Bezier Curve (both control points are equi-distant from curve_begin)
+ * 2. Asymmetric Bezier Curve (control points can be at different distances from curve_begin)
+ * 3. Quadratic Curve (single control point)
+ * 4. Line (pretty obvious)
+ */
+public struct Akira.Utils.PathSegment {
+    public Lib.Modes.PathEditMode.Type type;
+
+    // This point is used for drawing lines.
+    // Store point in segment_begin to save space, but use the alias to make code readable.
```
mbfraga, line 40: I don't understand how you are saving space, you are still copying those values and keeping them in memory--so you have duplicated storage.

mbfraga: I also don't think this is part of the fundamental path structure--it should be an API on top.

src/Utils/PathSegment.vala
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    // These points are used for drawing curves.
+    public Geometry.Point curve_begin;
+    public Geometry.Point tangent_1;
+    public Geometry.Point tangent_2;
+    public Geometry.Point curve_end;
+
+    public PathSegment () {
+        type = Lib.Modes.PathEditMode.Type.NONE;
```
mbfraga, line 69: Why do you initialize these values, it seems very inefficient to me. You don't need the values to make sense until you use them.

src/Lib/Components/Path.vala
```diff
@@ -87,22 +83,21 @@ public class Akira.Lib.Components.Path : Component, Copyable<Path> {
         double min_y = double.MAX;
```
mbfraga, line 83: It seems to me calculate_extents should not be part of the Path component, but part of the PathSegment

src/Models/PathEditModel.vala
```diff
@@ -68,103 +70,164 @@ public class Akira.Models.PathEditModel : Object {
         update_view ();
     }
 
-    public void add_live_points_to_path (Geometry.Point[] live_pts, Lib.Modes.PathEditMode.Type live_command, int length) {
-        commands.resize (commands.length + 1);
-        commands[commands.length - 1] = live_command;
-
+    public void add_live_points_to_path (Utils.PathSegment live_pts) {
```
mbfraga, line 73: I haven't read all this in detail. But my feeling is that this class has way more code than it should for what it does. I'm not sure if it's code duplication or something in the PathSegment struct that requires extra code.

src/Utils/PathSegment.vala
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        if (ptype == Lib.Modes.PathEditMode.PointType.TANGENT_SECOND) {
+            return tangent_2;
+        }
+
+        if (ptype == Lib.Modes.PathEditMode.PointType.CURVE_END) {
+            return curve_end;
+        }
+
```
mbfraga, line 197: You should probably assert that this can't happen here.

mbfraga (changes requested):

src/Utils/PathSegment.vala
AshishS-1123: @mbfraga Actually it does save space. 

If you take a look at the generated .c file in *build/src/libakira-library-1.0.a.p/Utils/PathSegment.c*, **PathSegment** is just a struct with a type and 4 points. 

In Vala, any getter and setter variables you create become methods. You can find the method **akira_utils_path_segment_set_line_end** in the .c file.

PR #709, opened 2022-02-06
Initial implementation of the fills color panel

Alecaddd:
Very simple initial PR to restore the fills color controls in the UI by using the same virtualizing listbox used in the layers panel.

- [x] Convert the `Components.Fill` into a simple `struct`.
- [x] Create the repeatable UI of the color row.
- [x] Defer the creation of heavy widgets only when needed.


Alecaddd:
This is ready for a review since it re-implements a lot of things.
The major thing missing is the color chooser when clicking on the color button. I'll do that later since it needs to be converted to a utility widget that can be reused.

Alecaddd:
Converted to a draft due to some substantial architectural changes to this whole area.

src/Models/ColorModel.vala
```diff
@@ -21,72 +21,74 @@
...
-    private unowned Lib.Components.Fill? fill;
-    private unowned Lib.Components.Border? border;
+    public signal void value_changed ();
 
-    public Type type;
-    public enum Type {
-        FILL,
-        BORDER
-    }
+    public class SignalBlocker {
```
mbfraga, line 29: Why does this signal blocker exist?

src/Lib/Components/Fills.vala
```diff
@@ -58,20 +106,50 @@ public class Akira.Lib.Components.Fills : Component, Copyable<Fills> {
...
         }
-        return tmp;
+        return false;
     }
 
-    public void prep_fills (uint number_to_prep) {
-        data = new Fill.FillData[number_to_prep];
-        for (var i = 0; i < number_to_prep; ++i) {
-            data[i]._id = i;
+    public void append (Fill fill) {
```
mbfraga, line 129: don't add public mutators to components. This breaks the model contract and thus copy/paste/undo/redo

mbfraga (changes requested):

src/Models/ColorModel.vala
Alecaddd: It's used by the FillitemModel when the node is updated for the first time, which blocks the `on_value_changed()` in the subclass.
Should the signal blocker live only in the subclass?

src/Lib/Components/Fills.vala
Alecaddd: What would you recommend to do here to handle multiple single `Fill` for the same `Fills` component?

PR #711, opened 2022-02-22
Fix various typos

luzpaz:
Found via `codespell -q 3 -L filetest,uint -S ./po`

### DIFF
--- a/src/Drawables/Drawable.vala
+++ b/src/Drawables/Drawable.vala
@@ -21,7 +21,7 @@
 
 /*
  * Overrides methods in Goo.CanvasItemSimple to handle a few things differently:
- * 1. Have uniform strokes in transfomed objects.
+ * 1. Have uniform strokes in transformed objects.
  * 2. Add a clipping quad to make clipping more efficient.
  */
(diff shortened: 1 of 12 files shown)


src/Drawables/Drawable.vala
```diff
@@ -262,17 +262,17 @@ public class Akira.Drawables.Drawable {
...
         context.move_to (-lenght, 0);
-        context.line_to (lenght, 0);
+        context.line_to (length, 0);
         context.stroke ();
 
         // Create the vertical stroke.
         context.new_path ();
         context.move_to (0, -lenght);
-        context.line_to (0, lenght);
+        context.line_to (0, length);
```
luzpaz, line 275: please review closely

Alecaddd: Thanks for the typos.
You need to replace all instances of `lenght` with `length`, even those in `context.move_to (-lenght, 0);`, otherwise the build is busted.

luzpaz: Done. Not sure how I missed those.

PR #714, opened 2022-02-28
Simple ListBox for Borders and Fills panels

Alecaddd:
I'm implementing a new `VirtualizingSimpleListBox` which is a simplified version of the listbox used in the layers panel.
This simplified version is not scrollable and is way simpler. There's still a lot that can be cleaned up in there, but this is a start to have a simple container with a factory method to reuse widgets.

These are all the things I did in this patch, in not particular order:
- Fills and Borders panels now use the simple listbox.
- If multiple items are selected, and more than 4 fills or borders are getting generated, just break out of the loop. In a later patch I will handle the "Show all" link, and also not recreating the widget if the color is the same, but rather having 1 widget to handle all identical fills/borders of currently selected items.
- Added the borders panel. For now it's identical to the fills panel, more work is needed.


Alecaddd:
Merging this because YOLO.
These widgets will improve drastically with GTK4, but as a starter these are good.

PR #729, opened 2022-07-14
Implement layers groups

Alecaddd:
Initial implementation to create a group from a selection of nodes.
Hit <kbd>CTRL</kbd>+<kbd>G</kbd> to create a group from a selection.

### ToDo
This is a first rough implementation of the feature, a lot more works is needed.
The most obvious issues are:
- An initial draft of a `transfer()` method is part of this PR, it's unfinished and unused, but the scope is to later use that to handle the group creation.
- The nodes names should be maintained when a transfer happens, instead of manually updating the name to match the old ones after the fact.
- Empty Groups shouldn't exist. When a node is delete from a group, we should check if that group is currently empty and delete it automatically.
- The UI of the layers group needs to be done.
- Hover effect for both canvas and layers panel needs to be implemented.
- Nested groups are not a thing at all.
- Many more little things I'm probably missing.

### Screenshot
group

### DIFF
--- a/src/Layouts/LayersList/LayerListBox.vala
+++ b/src/Layouts/LayersList/LayerListBox.vala
@@ -323,7 +323,7 @@ public class Akira.Layouts.LayersList.LayerListBox : VirtualizingListBox {
         var node1 = im.node_from_id (layer1.id);
         var node2 = im.node_from_id (layer2.id);
 
-        bool node1_is_group = node1.instance.is_group;
+        var node1_is_group = node1.instance.is_group;
         var node2_is_group = node2.instance.is_group;
 
         if (node1_is_group != node2_is_group) {
--- a/src/Layouts/LayersList/LayerListBox.vala
+++ b/src/Layouts/LayersList/LayerListBox.vala
@@ -323,6 +323,7 @@ public class Akira.Layouts.LayersList.LayerListBox : VirtualizingListBox {
         var node1 = im.node_from_id (layer1.id);
         var node2 = im.node_from_id (layer2.id);
 
+        // Why does this crash?
         var node1_is_group = node1.instance.is_group;
         var node2_is_group = node2.instance.is_group;
 
--- a/src/Layouts/LayersList/LayerListBox.vala
+++ b/src/Layouts/LayersList/LayerListBox.vala
@@ -323,8 +323,7 @@ public class Akira.Layouts.LayersList.LayerListBox : VirtualizingListBox {
         var node1 = im.node_from_id (layer1.id);
         var node2 = im.node_from_id (layer2.id);
 
-        // Why does this crash?
-        var node1_is_group = node1.instance.is_group;
+        bool node1_is_group = node1.instance.is_group;
         var node2_is_group = node2.instance.is_group;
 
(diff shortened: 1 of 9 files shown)


src/Layouts/LayersList/LayerListBox.vala
```diff
@@ -323,6 +323,7 @@ public class Akira.Layouts.LayersList.LayerListBox : VirtualizingListBox {
         var node1 = im.node_from_id (layer1.id);
         var node2 = im.node_from_id (layer2.id);
 
+        // Why does this crash?
         var node1_is_group = node1.instance.is_group;
         var node2_is_group = node2.instance.is_group;
```
Alecaddd, line 328: @mbfraga This creates a `SEGFAULT` and I can't figure out why

src/Lib/Managers/ItemsManager.vala
```diff
@@ -564,6 +563,56 @@ public class Akira.Lib.Managers.ItemsManager : Object, Items.ModelListener {
         }
     }
 
+    public void create_group_from_selection () {
```
Alecaddd, line 566: @mbfraga Initial rough implementation of what I think I figured it should happen.
TLDR:
- Collect all selected items in a tree map sorted by pos key.
- Create the group and add it to the canvas.
- Loop through the sorted tree and clone instance, delete node, add instance to the group.

Does this make sense or am I doing something incredibly silly?

src/Layouts/LayersList/LayerListBox.vala
Alecaddd: Uh, actually it seems that `node1.instance` doesn't exist.
I'm definitely doing something wrong.

Alecaddd: All right, I think I figured what I was doing wrong, which was removing the original node before the cloned instance had the chance to be used to create the new node.

src/Lib/Managers/ItemsManager.vala
Alecaddd: I think it would make sense to use a Util here to ensure the new node is an exact copy of the old node, maintaining its name and all its properties.
Unless, this is overkill and the only thing needed here is changing the parent ID (not sure if that would work), like a proper "move this item into this container", which will be the same method used to move items between artboards.

Alecaddd: This method will most likely leverage the `Model::transfer()` method I'm slowly adding.

src/Layouts/LayersList/LayerListBox.vala
```diff
@@ -323,7 +323,7 @@ public class Akira.Layouts.LayersList.LayerListBox : VirtualizingListBox {
         var node1 = im.node_from_id (layer1.id);
         var node2 = im.node_from_id (layer2.id);
 
-        var node1_is_group = node1.instance.is_group;
+        bool node1_is_group = node1.instance.is_group;
         var node2_is_group = node2.instance.is_group;
 
         if (node1_is_group != node2_is_group) {
```
Alecaddd, line 329: This doesn't properly handle nested groups.

PR #731, opened 2022-07-26
Drop standalone usage of Gdk.ModifierType

### LINKED ISSUE #730: Drop Gdk.ModifierType
We're currently using the `Gdk.ModifierType` to detect if the `CTRL` or `SHIFT` key are pressed.
We randomly use these inside various methods and it's not efficient.

Those modifiers they're not always accurate as they interfere with the OS window management shortcuts and they don't "clear" themselves nicely, resulting in inconsistent behaviors.

We have more efficient methods in the `ViewCanvas` called `ctrl_is_pressed()` and `shift_is_pressed()`, which use the more reliable modifier mask at the time of the interaction (click or keypress) and don't linger around.

We should replace the usage of `Gdk.ModifierType.SHIFT_MASK` and `Gdk.ModifierType.CONTROL_MASK` with those methods.

Alecaddd:
Drop standalone usage of Gdk.ModifierType in favor of boolean getters from the main canvas based on modifier intent.

- Fixes #730

### DIFF
--- a/src/Lib/Modes/PathEditMode.vala
+++ b/src/Lib/Modes/PathEditMode.vala
@@ -312,7 +312,6 @@ public class Akira.Lib.Modes.PathEditMode : AbstractInteractionMode {
 
         bool is_selected = edit_model.hit_test (event.x, event.y, ref sel_point, ref underlying_pt);
 
-        bool is_shift = (event.state == Gdk.ModifierType.SHIFT_MASK);
         bool is_alt = (event.state == Gdk.ModifierType.MOD1_MASK);
 
         if (!is_selected) {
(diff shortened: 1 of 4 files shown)


src/Lib/Modes/PathEditMode.vala
```diff
@@ -312,7 +312,6 @@ public class Akira.Lib.Modes.PathEditMode : AbstractInteractionMode {
 
         bool is_selected = edit_model.hit_test (event.x, event.y, ref sel_point, ref underlying_pt);
 
-        bool is_shift = (event.state == Gdk.ModifierType.SHIFT_MASK);
         bool is_alt = (event.state == Gdk.ModifierType.MOD1_MASK);
```
Alecaddd, line 315: @AshishS-1123 I can't figure out what do these shift and alt modifiers do while in editing mode.
Could you give me an example?

Alecaddd:
YOLO merging this since it's just code clean up

PR #734, opened 2022-07-27
Implement group hover and selection

Alecaddd:
Implement hover effect for groups, and the ability to select groups or child nodes by holding <kbd>CTRL</kbd>

### ToDo
- [x] Ignore artboards and allow selecting/hover child items
- [x] <kbd>CTRL</kbd> + Click on an item inside a selected group should select the item
- [x] If an item is already selected, don't go through the whole recursive parent detection again
- [x] Groups inside an artboard are selectable via click on an empty area. They should not
- [x] Double clicking an item inside a group should "open" the group, and select the double clicked item

### DIFF
--- a/data/css/artboard.css
+++ b/data/css/artboard.css
@@ -1,25 +1,13 @@
-.artboard {
+row.artboard {
     transition: box-shadow 160ms ease, background-color 160ms ease;
     background-color: shade (@bg_color, 1);
 }
 
-.artboard entry {
-    background-color: @base_color;
-}
-
-.artboard label {
+row.artboard label {
     font-weight: bold;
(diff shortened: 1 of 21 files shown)


Alecaddd:
This needs some refinement as it's not possible to select items inside artboards.
I will fix the recursive parent detection to account for those.

Alecaddd:
@mbfraga the `Utils.Array.insert_array_at_iarray()` method fails the assertion if the selected items to add to a group are not one after another, aka different indexes. Any clue?

Alecaddd:
Ready for a full review!

Alecaddd:
Doing a yolo merge!

PR #736, opened 2022-09-29
Fix unit tests for model

mbfraga:
Fix unit tests for model


tests/utils/'
```diff
@@ -0,0 +1,108 @@
+/*
```
Alecaddd, line 1: Typo when writing the file name?

Alecaddd (changes requested):

tests/utils/'
Alecaddd: It seems a duplicate of the test-array file

Alecaddd:
This is good to go, thank you so much for the addition.
Just waiting to hear about that extra file with a back tick as name, just in case, but I suspect is just an old accidental duplicate.

Alecaddd (approved):

Issue #732, opened 2022-07-26
Reimplement Save and Load file


Alecaddd:
Now that we switched to the `main` branch, in order to create a new Flatpak and elementaryOS release, we need to make sure the saving and loading of files works as expected.
Right now it just crashes, but it shouldn't be too difficult to make it work again as all the methods are in place.

mbfraga:
Reasons for assertions atm:
1. "rect" should be changed to "rectangle" in JsonDeserializer.vala
2. serialization/deserialization for path was never implemented.

PR #748, opened 2024-02-08
Fixes deserializing rectangle on file load

mradcliffe:

- Resolves loading a saved akira file with "rectangle" type.

I found akira and started playing around with it, then ran into this bug. I figured out the file format was a zipped json file pretty quickly, opened it up, saw the inconsistency, and the found the issue. I don't really feel comfortable writing in vala, but I felt I would be comfortable enough to create a pull request for this small fix.

## Steps to Test

1. Open Akira
2. Press the Insert button
3. Choose Shapes
4. Choose Rectangle
5. Place the Rectangle
6. Save the file
7. Close Akira
8. Open Akira
9. Open the file

Result: the application does not bail with 

    Akira:ERROR:../src/FileFormat/JsonDeserializer.vala:113:akira_file_format_json_deserializer_deserialize_node: assertion failed: (false)

- The rest of #732 is out-of-scope of what I can contribute.

### DIFF
--- a/src/FileFormat/JsonDeserializer.vala
+++ b/src/FileFormat/JsonDeserializer.vala
@@ -95,7 +95,7 @@ public class Akira.FileFormat.JsonDeserializer {
         var components = Lib.Components.Components.deserialize (components_json);
 
         Lib.Items.ModelInstance? new_inst = null;
-        if (type == "rect") {
+        if (type == "rectangle") {
             new_inst = new Lib.Items.ModelInstance (-1, new Lib.Items.ModelTypeRect ());
         } else if (type == "ellipse") {
             new_inst = new Lib.Items.ModelInstance (-1, new Lib.Items.ModelTypeEllipse ());


albfan:
To really avoid the problem, can you do another commit to change all those Magic words into constants?

mradcliffe:
> can you do another commit to change all those Magic words into constants?

I reviewed to see if I could handle this. Is like the best place for this

- enum constants within Akira.Lib.Items.Model?
- constant in each Akira.Lib.Items.ModelTypeThing?
- enum constants in one of the Managers?

Regardless it seems like model name_id method should return that constant as well.

Alecaddd (approved):
Thank you for your contribution, and apologies for the delay in the review.
I'm going through a much needed Gtk4 upgrade so some code will change, but this fix is small and good enough to merge.
Converting those string literal comparison to enums would be a nice thing to do, but it can be done later.
Thank you again for your contribution!

PR #761, opened 2024-09-21
Basic Rust layout

Alecaddd:
Implement a basic 3 pane layout in Rust


src/layout/left_sidebar.rs
```diff
@@ -0,0 +1,54 @@
...
+
+impl Default for LeftSidebar {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl LeftSidebar {
+    pub fn new() -> Self {
+        glib::Object::builder()
```
bilelmoussaoui, line 50: This could be moved to the Default impl, you can drop the new method then

src/layout/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+ * Akira is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * Akira. If not, see <https://www.gnu.org/ licenses/>.
+ */
+pub mod left_sidebar;
+
+pub mod sidebars {}
```
bilelmoussaoui, line 19: Seems unused

src/window.rs
```diff
@@ -132,6 +114,60 @@ impl AppWindow {
...
+
+        let gtk_settings = gtk::Settings::default().expect("Unable to get GtkSettings object");
+        mode_switch
+            .bind_property("active", &gtk_settings, "gtk-application-prefer-dark-theme")
+            .bidirectional()
+            .build();
+
+        let header_bar = gtk::HeaderBar::builder().show_title_buttons(true).build();
+
+        header_bar.style_context().add_class("default-decoration");
```
bilelmoussaoui, line 135: Use add_css_class, this method is deprecated

bilelmoussaoui (approved):

src/layout/mod.rs
Alecaddd: Yes, thank you, I was experimenting with Rust modules and I need to clean this up

src/window.rs
Alecaddd: Nice!

bilelmoussaoui (approved):

PR #776, opened 2025-12-23
Meson clean up

Alecaddd:
- Use `gnome` scripts and remove the custom Python post install
- Add meson minimum version
- Remove a couple of deprecated utils
- Replace deprecated meson `source_root()` and `build_root()` methods 
- Use the recommended `.Devel` RDNN format
- Updated the github actions for linting and building
- Small indentation clean up


com.github.akiraux.akira.yml
```diff
@@ -1,19 +1,19 @@
...
-  - '--share=ipc'
-  - '--socket=fallback-x11'
-  - '--socket=wayland'
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
 
   # needed for perfers-color-scheme
-  - '--system-talk-name=org.freedesktop.Accounts'
+  - --system-talk-name=org.freedesktop.Accounts
```
danirabbit: You can actually drop this since we use portals these days

